PR #58, opened 2026-05-05
[algo, trainer, rollout] feat: integrate MixGRPO sliding-window SDE training

SamitHuang:
### What does this PR do?

Add trainer-side **MixGRPO** (sliding-window ODE-SDE rollout) support, implementing the algorithm from [arXiv:2507.21802](https://arxiv.org/abs/2507.21802).

- **SDE-window scheduler** (`sde_window_scheduler.py`): a training-time state machine that slides the SDE window across the denoising trajectory. Supports `random` (seeded uniform draw each step) and `progressive` (advance by `sde_window_size` every `iters_per_group` steps) strategies.
- **Algo config** (`DiffusionRolloutAlgoConfig`): new `algo_type` selector (`flow_grpo` / `mix_grpo`) plus MixGRPO-specific knobs (`sample_strategy`, `iters_per_group`, `seed`). Fully backward-compatible: `algo_type=flow_grpo` is the default.
- **Trainer integration**: `RayFlowGRPOTrainer.fit()` queries the scheduler each step and injects `sde_window_size` / `sde_window_range` overrides via `meta_info["algo_overrides"]`. Also injects a deterministic per-step `rollout_seed` derived from `data.seed + global_steps` for reproducible A/B comparisons.
- **Agent loop**: `DiffusionAgentLoopWorker` merges `algo_overrides` into rollout sampling params before dispatch. Derives per-rollout seeds from the base step seed so `rollout.n` group members see different initial noise (avoids advantage collapse).

### Checklist Before Starting

- [ ] Search for similar PRs. Paste at least one query link here: ...
- [ ] Format the PR title as `[{modules}] {type}: {description}` (This will be checked by the CI)
  - `{modules}` include `fsdp`, `vllm_omni`, `rollout`, `trainer`, `ci`, `training_utils`, `recipe`, `ray`, `worker`, `single_controller`, `misc`, `perf`, `model`, `algo`, `env`, `tool`, `ckpt`, `doc`, `data`, `cfg`, `reward`, `diffusion`, `omni`, `tests`, `docker`
  - If this PR involves multiple modules, separate them with `,` like `[diffusion, doc]`
  - `{type}` is in `feat`, `fix`, `refactor`, `chore`, `test`
  - If this PR breaks any API (CLI arguments, config, function signature, etc.), add `[BREAKING]` to the beginning of the title.
  - Example: `[BREAKING][diffusion, fsdp] feat: new rollout scheduler`

### Test

image

[...]

PR #60, opened 2026-05-06
[trainer, rollout] fix: deterministic per-step and per-rollout seeding

SamitHuang:
### What does this PR do?

Based on #58

  - Fix 0-indexed vs 1-indexed mismatch in the trainer's `fit()` loop: pass `self.global_steps - 1` to `get_window()` and `rollout_seed` so the first training step correctly starts
  at `init_timestep` and uses the exact base seed.
  - Derive unique per-rollout seeds in the agent loop so group members (`n > 1`) see different SDE noise instead of identical trajectories.

### Checklist Before Starting

### Test

### API and Usage Example

```python
```

### Design & Code Changes

> [!IMPORTANT]


zhtmike:
won't this change make the rollout always have the same seed at each step?

For example, I expect the initial latent should be different for each rollout step. And we can give a config to enable same `latent`, and it is `off` by default.

verl_omni/trainer/diffusion/ray_diffusion_trainer.py
```diff
@@ -885,6 +885,12 @@ def fit(self):
 
                 # pass global_steps to trace
                 gen_batch.meta_info["global_steps"] = self.global_steps
+
+                # Per-step rollout seed for reproducibility
+                data_seed = self.config.data.get("seed")
```
zhtmike: better to separate data_seed and rollout_seed, we can add a option of rollout_seed in yaml and config class

SamitHuang:
> won't this change make the rollout always have the same seed at each step?
> 
> For example, I expect the initial latent should be different for each rollout step. And we can give a config to enable same `latent`, and it is `off` by default.

  No — the rollout does not have the same seed at each step. Two layers of variation are stacked:

  Per training step (in ray_diffusion_trainer.py):

   verl_omni/trainer/diffusion/ray_diffusion_trainer.py lines 879-882
```
                  # Per-step rollout seed for reproducibility
                  data_seed = self.config.data.get("seed")
                  if data_seed is not None:
                      gen_batch.meta_info["rollout_seed"] = int(data_seed) + self.global_steps - 1
```
  So the base seed advances by 1 every training step:
  • step 1 → data_seed + 0
  • step 2 → data_seed + 1
  • …

  Per rollout within a step (in diffusion_agent_loop.py):

   verl_omni/agent_loop/diffusion_agent_loop.py lines 48-54
```
  _PER_ROLLOUT_SEED_STRIDE = 1_000_003
  _MAX_SEED = (1 << 63) - 1
  def _derive_rollout_seed(base_seed: int, rollout_index: int) -> int:
      """Derive a unique per-rollout seed so group members see different noise."""
      return (int(base_seed) * _PER_ROLLOUT_SEED_STRIDE + int(rollout_index)) % _MAX_SEED
```
  So for prompt i at step s the actual seed handed to the rollout is ((data_seed + s - 1) * 1_000_003 + i) % 2^63-1.

zhtmike:
let's move this PR to v0.2.0 ?

SamitHuang:
> let's move this PR to v0.2.0 ?

It will be helpful to debug training convergence and be relied by https://github.com/verl-project/verl-omni/issues/105. Let me run an e2e training to ensure it won't affect current performance

zhtmike:
> > let's move this PR to v0.2.0 ?
> 
> It will be helpful to debug training convergence and be relied by #105. Let me run an e2e training to ensure it won't affect current performance

OK. Then needs to be care with seed.

I suggest to provide a enough test converage:

1. make sure the initial noise is distinct for each rollout step if the seed is enabled.
2. make sure the initial noise is same across TP ranks.
3. make sure the initial noise is different across DP ranks.
4. leave a option to disable seed for cross checking.

zhtmike:
> > > let's move this PR to v0.2.0 ?
> > 
> > 
> > It will be helpful to debug training convergence and be relied by #105. Let me run an e2e training to ensure it won't affect current performance
> 
> OK. Then needs to be care with seed.
> 
> I suggest to provide a enough test converage:
> 
> 1. make sure the initial noise is distinct for each rollout step if the seed is enabled.
> 2. make sure the initial noise is same across TP ranks.
> 3. make sure the initial noise is different across DP ranks.
> 4. leave a option to disable seed for cross checking.

may added to test on GPU directly, maybe on `tests/agent_loop/test_diffusion_agent_loop.py`

zhtmike:
And split data seed and rollout seed

zhtmike:
@AndyZhou952 please take a look, since it is agent-loop related.

AndyZhou952 (commented):
Suggest cleaning CPU tests and only keep the GPU test as suggested by @zhtmike

MmMaiIIi:
I’m working on the #105 deterministic rollout RFC follow-up and did a small audit around DP seed identity.

One concrete risk I found is around the index used for per-rollout seed derivation after chunk dispatch.

The simplified path is:

```text
trainer:
  repeat prompts by rollout.n
  -> expanded rollout rows

AgentLoopManager:
  split rows into worker chunks

AgentLoopWorker:
  receives one local chunk
  local indices start again from i = 0
```

If per-rollout seeds are derived from the worker-local `i`, then two DP workers can derive the same seed for different global rollout rows.

Example:

```text
16 expanded rows, 2 workers

worker 0: global rows 0..7,  local i = 0..7
worker 1: global rows 8..15, local i = 0..7
```

With local-index-only derivation:

```text
worker 0 seeds = derive(base, 0..7)
worker 1 seeds = derive(base, 0..7)
```

So both workers can get the same derived seed list, even though they are handling different global samples.

I reproduced this with a CUDA test: two simulated workers produced identical derived seeds and identical latent hashes without an offset.

I also drafted GPU-only invariant tests for:

```text
- seed-disabled path;
- same-seed repeatability;
- step-to-step distinctness;
- rollout.n diversity;
- TP same-noise invariant;
- DP collision repro without offset;
- DP distinctness with offset;
- debug/report coverage.
```

All 8 pass locally.

The main invariant I think we should preserve is:

```text
Per-rollout seed identity should be based on the expanded global rollout row,
not the worker-local row index.
```

In other words, local-index-only derivation is unsafe:

```text
seed = derive(base, local_i)
```

A safer identity would be equivalent to:

```text
seed = derive(base, global_expanded_rollout_row_index)
```

This could be implemented through an explicit global row id, prefix offset, or a rollout seed offset passed into the worker chunk.

I haven't patched source behavior yet. I think the ownership question matters first:

Should the global sample offset / DP-aware seed identity be injected in the upstream `AgentLoopManager` dispatch path, or handled locally in the verl-omni rollout path after this PR stabilizes?

zhtmike:
@MmMaiIIi Thank you for your clear explanation!

That is exactly about the seed safety bugs I am worrying about.

I think we should clarify the ownership first before any implementation. Can you help 
1. design the exact seed mechanism under DP 
2. the GPU test converage as your mentioned

MmMaiIIi:
> @MmMaiIIi Thank you for your clear explanation!
> 
> That is exactly about the seed safety bugs I am worrying about.
> 
> I think we should clarify the ownership first before any implementation. Can you help
> 
> 1. design the exact seed mechanism under DP
> 2. the GPU test converage as your mentioned

Thanks! I’m happy to take these two points.

I’ll follow up with a first pass on the DP seed mechanism and the GPU test coverage.

zhtmike:
> > @MmMaiIIi Thank you for your clear explanation!
> > That is exactly about the seed safety bugs I am worrying about.
> > I think we should clarify the ownership first before any implementation. Can you help
> > 
> > 1. design the exact seed mechanism under DP
> > 2. the GPU test converage as your mentioned
> 
> Thanks! I’m happy to take these two points.
> 
> I’ll follow up with a first pass on the DP seed mechanism and the GPU test coverage.

Thanks! pin me once it is ready

knlnguyen1802:
I think it's better to add some test and assertion fix the result, it's a test to ensure the rollout result is determinisic

MmMaiIIi:
> I think it's better to add some test and assertion fix the result, it's a test to ensure the rollout result is determinisic

That makes sense. I’ll make the smoke assertion-backed rather than report-only.

MmMaiIIi:
@zhtmike @knlnguyen1802 I opened a first-pass follow-up PR here: #112.

It focuses on the DP/chunk-dispatch seed identity issue discussed above. The main change is to assign explicit expanded global rollout row ids right after `gen_batch.repeat(...)`, then use those ids in the AgentLoop per-rollout seed path instead of relying on worker-local chunk indices.

This is intended to preserve the invariant:

```text
per-rollout seed identity = expanded global rollout row
not worker-local row index
```

The PR is currently scoped as a small production-side patch. I summarized the GPU / real-model assertion-backed validation evidence in the PR body, but I did not include the heavier GPU validation harness as committed test files because it is environment-specific.

If you prefer, I can add a small portable regression test in this PR for the reusable seed-contract pieces, and keep the dual-A100 real-model checks as external validation evidence.

zhtmike:
> @zhtmike @knlnguyen1802 I opened a first-pass follow-up PR here: #112.
> 
> It focuses on the DP/chunk-dispatch seed identity issue discussed above. The main change is to assign explicit expanded global rollout row ids right after `gen_batch.repeat(...)`, then use those ids in the AgentLoop per-rollout seed path instead of relying on worker-local chunk indices.
> 
> This is intended to preserve the invariant:
> 
> ```
> per-rollout seed identity = expanded global rollout row
> not worker-local row index
> ```
> 
> The PR is currently scoped as a small production-side patch. I summarized the GPU / real-model assertion-backed validation evidence in the PR body, but I did not include the heavier GPU validation harness as committed test files because it is environment-specific.
> 
> If you prefer, I can add a small portable regression test in this PR for the reusable seed-contract pieces, and keep the dual-A100 real-model checks as external validation evidence.

Thanks, please add GPU tests in your PR. Let us discuss on your PR directly.

SamitHuang:
image
image

With the same seed, the two runs show correlated curves  (r=0.92). 
It not yet deterministic due to
  1. SDE window start uses the same generator as latents (should be seeded) — OK in theory
  2. Async agent loop (asyncio.gather over 512 images) — ordering / timing nondeterminism  #112 
  3. CUDA nondeterministic ops in training and vLLM-omni rollout
  4. GRM reward scoring (async vLLM judge) — can reorder or jitter scores slightly. #111 
  5. Actor full_determinism is not enabled. actor_rollout_ref.actor.fsdp_config.full_determinism (default false)

SamitHuang:
> And split data seed and rollout seed

I think one seed control is more ease-of-use. Currently, rollout seed is derived from data.seed at runtime.

zhtmike:
> > And split data seed and rollout seed
> 
> I think one seed control is more ease-of-use. Currently, rollout seed is derived from data.seed at runtime.

No, it is not a good design. A `xxx` seed should point to the `xxx` clearly. It will avoid any ambiguity raised in the code and implementation.

zhtmike:
And if I am understanding correctly, this PR is actually implementing `same latent` mechanism in dancegrpo. It is not intended as found by @MmMaiIIi , please take a look

tests/agent_loop/test_diffusion_agent_loop_seed_on_cpu.py
zhtmike, line 1: this cpu test is useless, please drop

verl_omni/agent_loop/diffusion_agent_loop.py
```diff
@@ -45,6 +46,91 @@ def _config_to_sampling_dict(config: Optional[BaseConfig]) -> dict:
...
+            "Using data.seed for rollout RNG is deprecated; set actor_rollout_ref.rollout.seed instead.",
+            DeprecationWarning,
+            stacklevel=2,
+        )
+        base = data_seed
+    if base is None:
+        return None
+    return int(base) + int(global_steps) - 1
+
+
```
zhtmike: 1. move to verl_omni/agent_loop/utils.py
2. drop global variable
3. shorten the docstring into 2 lines (it is too AI now)

verl_omni/trainer/diffusion/ray_diffusion_trainer.py
```diff
@@ -950,6 +951,16 @@ def fit(self):
 
                 # pass global_steps to trace
                 gen_batch.meta_info["global_steps"] = self.global_steps
+
+                # Per-step rollout seed for reproducibility
+                rollout_seed = _build_rollout_seed(
+                    self.config.actor_rollout_ref.rollout.get("seed"),
+                    self.global_steps,
+                    data_seed=self.config.data.get("seed"),
```
zhtmike: drop `data_seed=self.config.data.get("seed")`

verl_omni/trainer/config/data/legacy_data.yaml
```diff
@@ -61,7 +61,10 @@ return_full_prompt: False
 # Whether to shuffle the data in the dataloader.
 shuffle: True
 
-# Seed to use when shuffling the data
+# Seed for train dataloader shuffling when data.shuffle=True.
+# Rollout RNG is controlled separately by actor_rollout_ref.rollout.seed.
+#  - integer (e.g. 42): reproducible dataloader order across restarts.
+#  - null (default): unseeded shuffle when data.shuffle=True.
 seed: null
```
zhtmike, line 65: revert this

zhtmike (commented):
left comments

verl_omni/trainer/diffusion/ray_diffusion_trainer.py
```diff
@@ -50,6 +50,7 @@
 from verl.utils.tracking import ValidationGenerationsLogger
 from verl.workers.rollout.llm_server import LLMServerManager
 
+from verl_omni.agent_loop.diffusion_agent_loop import _build_rollout_seed
```
zhtmike: `_build_rollout_seed` is just a single line function, make it inline here

examples/flowgrpo_trainer/run_qwen_image_ocr_lora.sh
```diff
@@ -24,6 +24,9 @@ python3 -m verl_omni.trainer.main_diffusion \
     data.val_files=$ocr_test_path \
     data.train_batch_size=32 \
     data.max_prompt_length=256 \
+    # Optional reproducibility overrides (defaults are null / unseeded):
+    # data.seed=42 \
+    # actor_rollout_ref.rollout.seed=42 \
```
zhtmike: the script will not be runnable

examples/flowgrpo_trainer/run_qwen_image_ocr_lora.sh
```diff
@@ -18,6 +18,9 @@ REWARD_TP=4
 ENGINE=vllm_omni
 REWARD_ENGINE=vllm
 
+# Optional reproducibility (yaml defaults are null / unseeded):
+#   data.seed=42
+#   actor_rollout_ref.rollout.seed=42
```
knlnguyen1802, line 23: Better to create a separate script for deterministic reproduce

tests/agent_loop/test_diffusion_agent_loop_seed_on_cpu.py
zhtmike: this CPU test file is useless, please drop

tests/agent_loop/test_diffusion_rollout_seed_gpu.py
zhtmike, line 1: as pointed by @MmMaiIIi , there is a known bug of seed collision for current implementation.

Since we know it is a bug, we should presented here with a test case, mark it as `xfailed` for pytest, fix it in future PR.

verl_omni/agent_loop/diffusion_agent_loop.py
```diff
@@ -174,7 +177,11 @@ async def generate_sequences(self, batch: DataProto) -> DataProto:
         tasks = []
         for i in range(len(batch)):
             kwargs = {k: v[i] for k, v in batch.non_tensor_batch.items()}
-            tasks.append(asyncio.create_task(self._run_agent_loop(sampling_params, **kwargs)))
+            task_sampling_params = sampling_params
+            if per_rollout_seeds is not None:
+                task_sampling_params = dict(sampling_params)
```
zhtmike: `task_sampling_params = dict(sampling_params)` is implicit, as we know it is aimed for copy, why don't use sampling_params.copy(), it is more clear.

verl_omni/agent_loop/utils.py
```diff
@@ -0,0 +1,40 @@
...
+    """Build one seed per post-repeat rollout row (batch_size == len(batch)).
+    Returns None when meta_info rollout_seed is unset."""
+    base = meta_info.get("rollout_seed")
+    if base is None:
+        return None
+    base = int(base)
+    return [_derive_rollout_seed(base, i) for i in range(batch_size)]
+
+
+def _build_rollout_seed(rollout_seed, global_steps: int) -> Optional[int]:
```
zhtmike: `_build_rollout_seed` does not belongs to agent loop, but for `ray_diffusion_trainer.py`. Please either make it inline or put it on the top of `ray_diffusion_trainer.p`

verl_omni/trainer/config/data/legacy_data.yaml
```diff
@@ -61,7 +61,8 @@ return_full_prompt: False
 # Whether to shuffle the data in the dataloader.
 shuffle: True
 
-# Seed to use when shuffling the data
+# Seed for train dataloader shuffling when data.shuffle=True.
+# Rollout RNG is controlled separately by actor_rollout_ref.rollout.seed.
```
zhtmike: it is not reverted.

Apparently users will not think `data.seed` has correlation with `rollout.seed`. Added a extra comment here is "畫蛇添足“。

zhtmike (commented):
commented

verl_omni/workers/rollout/vllm_rollout/vllm_omni_async_server.py
```diff
@@ -92,6 +92,22 @@ def _get_cli_modules(self) -> list:
...
+                value = original_get(key, default)
+                return 0 if value is None else value
+            return original_get(key, default)
+
+        self.config.get = get_with_engine_seed_default
+        try:
+            await super().launch_server(master_address, master_port, dp_rpc_port)
+        finally:
+            self.config.__dict__.pop("get", None)
+
```
zhtmike, line 117: all looks fine, except this not sure @knlnguyen1802 can you take look?

knlnguyen1802: I think this is not necessary, in the upstream verl https://github.com/verl-project/verl/blob/381917433d781b68e87644cb3d621a3bfadd96b9/verl/workers/rollout/vllm_rollout/vllm_async_server.py#L268 the default value for `"seed"` is already `0`

Also if we need to keep this to avoid change in upstream `verl` broke it then at line 109 should set the `get` into `original_get` instead of remove it

zhtmike: nice catch! FYI @SamitHuang

SamitHuang: The 0 is only used when the seed attribute is missing. With Hydra actor_rollout_ref.rollout.seed=null, the field `seed` exists and is None. So need to override `launch_server` method to handle `seed=null` case.

but in long term, it should be fixed in verl side

knlnguyen1802: I see, it's a very weird setup, but could you fix this ?

`at line 109 should set the get into original_get instead of remove it`

examples/flowgrpo_trainer/run_qwen_image_ocr_lora.sh
SamitHuang: will included in another PR

verl_omni/workers/rollout/vllm_rollout/vllm_omni_async_server.py
SamitHuang: done

zhtmike (approved):
LGTM, please raise a PR in verl for this fix

SamitHuang:
> LGTM, please raise a PR in verl for this fix

https://github.com/verl-project/verl/pull/6503

verl_omni/workers/rollout/vllm_rollout/vllm_omni_async_server.py
```diff
@@ -48,6 +48,37 @@
...
+    def __getitem__(self, key: str) -> Any:
+        return self._config[key]
+
+    def __iter__(self):
+        return iter(self._config)
+
+    def __len__(self) -> int:
+        return len(self._config)
+
+
```
zhtmike: again @knlnguyen1802 ..

SamitHuang: it can also be dropped after the verl PR merged. For now, we can replace this class with pop-based version for less diff

zhtmike:
Possibly to add a "# TODO" here to remember to drop them.

SamitHuang:
> Possibly to add a "# TODO" here to remember to drop them.

added

zhtmike:
I think we should increase timeout-minutes to 30 mins for `.github/workflows/gpu_smoke.yml` or drop the vllm resintall after verl pin / image upgrade.

zhtmike:
LGTM

zhtmike (approved):

PR #110, opened 2026-05-24
[rollout] fix: use reload_weights for vLLM v1 worker compatibility in non-LoRA path

qinganrice:
### What does this PR do?

The non-LoRA branch in `vLLMOmniColocateWorkerExtension._update_weights calls self.load_weights(weights)`, but vLLM v1's GPUWorker exposes the method as reload_weights (renamed from v0, see vllm/v1/worker/gpu_worker.py:354). The bug never surfaces during LoRA training (the if `peft_config `and `base_sync_done` branch uses `add_lora`) but breaks any non-LoRA rollout weight sync (most commonly base-model evaluation via `val_only=True` without lora_rank). The class docstring already claims compatibility with both vLLM V0 and V1, so this fix completes that intent rather than changing it.

### Checklist Before Starting

- [x] Search for similar PRs. Paste at least one query link here: https://github.com/verl-project/verl-omni/pulls?q=is%3Apr+is%3Aopen+reload

### Test

```
python3 -m verl_omni.trainer.omni.main_ppo \
    actor_rollout_ref.model.path=Qwen/Qwen3-Omni-30B-A3B-Instruct \
    actor_rollout_ref.rollout.name=vllm_omni \
    actor_rollout_ref.rollout.tensor_model_parallel_size=4 \
    ++trainer.val_only=True \
    ++trainer.resume_mode=disable \
    trainer.val_before_train=True \
    trainer.n_gpus_per_node=4 \
    trainer.nnodes=1 \
    data.val_files=$HOME/data/math/test.parquet \
    data.val_max_samples=-1
```
Result before fix:
```
AttributeError: 'GPUARWorker' object has no attribute 'load_weights'.
Did you mean: 'reload_weights'?
  File "verl_omni/workers/rollout/vllm_rollout/utils.py", line 64, in _update_weights
    self.load_weights(weights)
```

### API and Usage Example

No API changes. Behaviour change is limited to the previously-broken non-LoRA `update_weights_from_ipc path`, which now succeeds instead of raising AttributeError.

### Design & Code Changes

One-line change in `verl_omni/workers/rollout/vllm_rollout/utils.py,` inside `vLLMOmniColocateWorkerExtension._update_weights`:

```
         else:
             logger.info("Loading standard weights (async)")
-            self.load_weights(weights)
+            self.reload_weights(weights)

```

> [!IMPORTANT]

### DIFF
--- a/verl_omni/workers/rollout/vllm_rollout/utils.py
+++ b/verl_omni/workers/rollout/vllm_rollout/utils.py
@@ -82,7 +82,7 @@ class vLLMOmniColocateWorkerExtension(NPUColocateWorkerMixin, CustomPipelineWork
             logger.info(f"vLLM-Omni load weights, loaded_params: {len(weights)}")
         else:
             logger.info("Loading standard weights (async)")
-            self.load_weights(weights)
+            self.reload_weights(weights)
 
     def _get_zmq_handle(self) -> str:
         """Get ZMQ handle for communication.


zhtmike:
so far we don't encounter any weight update problem for full model diffusion training. 

@AndyZhou952 @knlnguyen1802 can you take a look?

AndyZhou952:
> so far we don't encounter any weight update problem for full model diffusion training.
> 
> @AndyZhou952 @knlnguyen1802 can you take a look?

Thanks for the PR!

`reload_weights` does not seem to be available in vllm-omni and Qwen3-Omni has not been officially supported in this repo yet. Not too sure about the use case here. Could you elaborate a bit more on why we need this change? The original impl works fine on my side.

verl_omni/workers/rollout/vllm_rollout/utils.py
```diff
@@ -82,7 +82,7 @@ def _update_weights(self, weights: list[tuple[str, torch.Tensor]], peft_config:
             logger.info(f"vLLM-Omni load weights, loaded_params: {len(weights)}")
         else:
             logger.info("Loading standard weights (async)")
-            self.load_weights(weights)
+            self.reload_weights(weights)
```
knlnguyen1802, line 85: It's true that vllm use reload_weights API for this, but to avoid conflict with vllm-omni diffusion model, I think you need to separate two case

qinganrice: Thanks for your comments. This issue is fixed in PR #113, so this PR can be closed.

AndyZhou952:
Follow #113.

PR #112, opened 2026-05-26
[trainer, rollout] fix: use global row ids for per-rollout seeds

MmMaiIIi:
## Context

Follow-up to #60, which added deterministic per-step and per-rollout seeding. That covered the basic seed schedule and made different `rollout.n > 1` group members derive distinct rollout seeds.

One distributed identity gap remained: after `gen_batch.repeat(...)` expands the batch, AgentLoop DP/chunk dispatch can split expanded rows across workers. Worker-local row indices are not globally unique, so using local chunk positions for per-rollout seed identity can collide across chunks.

## What this PR changes

- Assigns `_rollout_seed_global_idx` immediately after `gen_batch.repeat(...)` in `PolicyGradientRayTrainer.fit()`.
- Reads `_rollout_seed_global_idx` from each dispatched chunk in `DiffusionAgentLoopWorker`.
- Makes `_maybe_per_rollout_seeds(...)` derive seeds from explicit global row ids when provided.
- Keeps local-index seed derivation only as a compatibility fallback for older/direct callers.
- Rejects malformed explicit global-index columns whose length does not match the dispatched batch.

## Why this matters

`rollout.n` means each prompt produces multiple stochastic samples. Those samples should keep distinct seed/noise identity even after dispatch chunking. The identity should come from the expanded global row, not from the row’s position inside a worker-local chunk.

Before:

- worker 0 chunk local rows: `0, 1`
- worker 1 chunk local rows: `0, 1`
- seed identity based on local row can collide

After:

- expanded global rows: `0, 1, 2, 3`
- dispatched chunks carry global ids
- seed identity remains unique across workers

## Validation

Main validation was GPU / real-model backed on a dual NVIDIA A100 80GB machine. The evidence below comes from preserved assertion-backed rollout artifacts for this seed/noise identity path.

- Single-GPU real-model rollout: checked same seed/row repeatability, rollout-member diversity, step diversity, and seed-disabled behavior.
- Dispatch DP=2, TP=1: checked global row ids are present, local-index fallback is not used, derived seeds are unique, and initial latent hashes are unique across expanded rows.
- Dispatch-DP stress with larger `rollout.n`: checked the same uniqueness invariants across 8 expanded rows.
- TP=2: checked TP ranks share the same seed/hash for the same expanded row, so TP rank is not part of seed identity.
- Bounded trainer integration: entered `PolicyGradientRayTrainer.fit()`, executed real `gen_batch.repeat(...)`, produced `_rollout_seed_global_idx` in the trainer path, and observed it downstream before stopping before reward/actor update.
- Negative guard: disabling trainer global-row production caused the expected failure, catching missing global ids, local-index fallback use, duplicated seeds, and duplicated latent hashes.

Environment:

- GPU: 2 x NVIDIA A100-SXM4-80GB
- Driver: `580.126.09`
- CUDA runtime reported by PyTorch: `13.0`
- PyTorch: `2.11.0+cu130`
- Python: `3.10.16`
- Model path: Qwen-Image / FlowGRPO rollout through `vllm_omni` and `AgentLoopManager`

Additional local checks on this production-only branch:

- `git diff --check`
- `python -m py_compile` on modified Python files

## Boundaries

- This PR does not claim full deterministic RL training.
- This PR does not claim final image quality or convergence.
- This PR does not claim all distributed topologies.
- This PR does not claim vLLM internal `data_parallel_size=2` semantics.
- This PR only fixes the rollout seed identity path covered by the listed tests.

## Reviewer notes

- This is intentionally small: trainer-side global row id production plus AgentLoop seed consumption.
- The fallback remains for older/direct callers, but the trainer path now carries explicit expanded-row identity.
- The negative guard shows the exact DP/chunk collision mode is detectable.

### DIFF
--- a/tests/agent_loop/test_diffusion_rollout_seed_identity.py
+++ b/tests/agent_loop/test_diffusion_rollout_seed_identity.py
@@ -1,0 +1,70 @@
+# Copyright 2026 Bytedance Ltd. and/or its affiliates
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import importlib.util
+from pathlib import Path
+
+import numpy as np
+import pytest
+
+
+def _load_seed_utils():
+    # Keep this CPU unit test independent of package-level diffusion stack auto-registration.
+    utils_path = Path(__file__).resolve().parents[2] / "verl_omni" / "agent_loop" / "utils.py"
+    spec = importlib.util.spec_from_file_location("diffusion_seed_utils", utils_path)
+    module = importlib.util.module_from_spec(spec)
+    spec.loader.exec_module(module)
+    return module
+
+
+_seed_utils = _load_seed_utils()
+_derive_rollout_seed = _seed_utils._derive_rollout_seed
+_maybe_per_rollout_seeds = _seed_utils._maybe_per_rollout_seeds
+
+
+def test_per_rollout_seeds_use_explicit_global_row_ids_across_chunks():
+    base_seed = 42
+    first_chunk = _maybe_per_rollout_seeds(
+        {"rollout_seed": base_seed},
+        batch_size=2,
+        global_indices=np.array([0, 1], dtype=np.int64),
+    )
+    second_chunk = _maybe_per_rollout_seeds(
+        {"rollout_seed": base_seed},
+        batch_size=2,
+        global_indices=np.array([2, 3], dtype=np.int64),
+    )
+
+    assert first_chunk == [_derive_rollout_seed(base_seed, 0), _derive_rollout_seed(base_seed, 1)]
+    assert second_chunk == [_derive_rollout_seed(base_seed, 2), _derive_rollout_seed(base_seed, 3)]
+    assert set(first_chunk).isdisjoint(second_chunk)
+
+
+def test_per_rollout_seeds_keep_worker_local_fallback():
+    base_seed = 42
+
+    assert _maybe_per_rollout_seeds({"rollout_seed": base_seed}, batch_size=3) == [
+        _derive_rollout_seed(base_seed, 0),
+        _derive_rollout_seed(base_seed, 1),
+        _derive_rollout_seed(base_seed, 2),
+    ]
+
+
+def test_per_rollout_seeds_disabled_when_rollout_seed_is_unset():
+    assert _maybe_per_rollout_seeds({}, batch_size=2, global_indices=np.array([0, 1], dtype=np.int64)) is None
+
+
+def test_per_rollout_seeds_reject_malformed_global_row_ids():
+    with pytest.raises(ValueError, match="Expected 2 global rollout indices, got 1"):
+        _maybe_per_rollout_seeds({"rollout_seed": 42}, batch_size=2, global_indices=np.array([0], dtype=np.int64))
--- a/verl_omni/agent_loop/diffusion_agent_loop.py
+++ b/verl_omni/agent_loop/diffusion_agent_loop.py
@@ -168,7 +168,10 @@ class DiffusionAgentLoopWorker:
             sampling_params["logprobs"] = False
         else:
             sampling_params["global_steps"] = batch.meta_info["global_steps"]
-            per_rollout_seeds = _maybe_per_rollout_seeds(batch.meta_info, len(batch))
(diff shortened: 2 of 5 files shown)


tests/agent_loop/test_diffusion_rollout_seed_identity_gpu.py
zhtmike, line 1: Please take a look of `tests/agent_loop/test_diffusion_agent_loop.py`.

We use `tiny-random/Qwen-Image` for online CI and it is not heavy. I suggest to follow its practice.

Possibly direct append the test on `tests/agent_loop/test_diffusion_agent_loop.py`.

pyproject.toml
```diff
@@ -65,6 +65,12 @@ ignore = [
     "UP035",
 ]
 
+[tool.pytest.ini_options]
+markers = [
+    "gpu: marks tests that require CUDA devices",
+    "integration: marks tests that run integration or real-model workflows",
+]
+
```
zhtmike: let us drop this

verl_omni/pipelines/qwen_image_flow_grpo/vllm_omni_rollout_adapter.py
```diff
@@ -39,6 +41,70 @@ def _coalesce_not_none(value, default):
...
+        "initial_latent_sha256": hashlib.sha256(latent_bytes).hexdigest(),
+    }
+    encoded = (json.dumps(record, separators=(",", ":"), sort_keys=True) + "\n").encode("utf-8")
+    fd = os.open(trace_path, os.O_CREAT | os.O_APPEND | os.O_WRONLY, 0o644)
+    try:
+        os.write(fd, encoded)
+    finally:
+        os.close(fd)
+
+
```
zhtmike: I don't like the idea of add latent tracing on real pipeline.

Possibly add a simple pipeline under `tests/`, we can trace on that during the test

verl_omni/agent_loop/diffusion_agent_loop.py
```diff
@@ -45,6 +45,63 @@ def _config_to_sampling_dict(config: Optional[BaseConfig]) -> dict:
...
+
+    Otherwise returns ``int(data_seed) + int(global_steps) - 1``, which mirrors
+    the trainer's 1-indexed step convention: step 1 uses base = data_seed,
+    step 2 uses base = data_seed + 1, etc.
+    """
+    if data_seed is None:
+        return None
+    return int(data_seed) + int(global_steps) - 1
+
+
```
zhtmike: move them under `agent_loop/utils.py`

and no global variable, please

verl_omni/trainer/diffusion/ray_diffusion_trainer.py
```diff
@@ -948,11 +949,17 @@ def fit(self):
 
                 gen_batch = self._get_gen_batch(batch)
 
-                # pass global_steps to trace
+                # Pass step metadata to rollout before expansion.
                 gen_batch.meta_info["global_steps"] = self.global_steps
+                rollout_seed = _build_rollout_seed(self.config.data.get("seed"), self.global_steps)
```
zhtmike: please separate data seed and rollout seed

zhtmike (commented):
Thank you for your PR. Left with few comments

SamitHuang:
@MmMaiIIi thanks for the bugfix. Can you rebase to main so that we can merge quickly?

MmMaiIIi:
> @MmMaiIIi thanks for the bugfix. Can you rebase to main so that we can merge quickly?

Sure! Rebased onto latest main and force-pushed.

MmMaiIIi:
@zhtmike Thanks for the review. I agree with the direction.

I will revise the PR to make the change less invasive and closer to the existing test style:

1. Drop the new pytest markers from `pyproject.toml`.
2. Avoid adding latent tracing to the real rollout adapter.
3. Move the seed helper logic into `agent_loop/utils.py` and remove module-level globals.
4. Separate the data seed from the rollout seed instead of deriving rollout seed directly from `config.data.seed`.
5. Replace the standalone heavy GPU/integration-style test with a lighter test following `tests/agent_loop/test_diffusion_agent_loop.py`, using `tiny-random/Qwen-Image`, and append it there if possible.

tests/agent_loop/test_diffusion_rollout_seed_identity.py
zhtmike, line 1: seems we can drop this test script? the above gpu test already has the coverage

zhtmike (commented):
Looks good!

tests/agent_loop/test_diffusion_rollout_seed_identity.py
MmMaiIIi: Yes, we can drop it.

zhtmike: ok let us drop it

MmMaiIIi:
Hi all, sorry that I’ve been quite busy recently, so I may not have time to continue working on this over the next few days.

Please feel free to pick this up, push any improvements, or take it over if anyone has bandwidth. Thanks!

zhtmike:
thanks for your contribution! @SamitHuang can you do a follow up？

zhtmike:
Thank you for your contribution. It is merged in #136

PR #115, opened 2026-05-28
[reward] feat: support deterministic reward inference

### LINKED ISSUE #111: [RFC] Deterministic Reward Inference and VLM Reward Model Support
## Motivation

verl-omni's reward framework supports two reward model pathways:

- **DisRM** (discriminative): text-only, subject to floating-point nondeterminism, **cannot handle visual input**
- **GenRM** (generative): supports VLM Vision API, but **sampling params hardcoded** (`temperature=0.7, top_p=0.8`) in every `/v1/chat/completions` request, with no seed control for reproducibility

Three gaps:

1. **No floating-point determinism** — flash attention split-K, CUDA matmul, NCCL all-reduce, MoE routing can all produce floating-point variation across RM inference runs
2. **No configurable GenRM sampling** — GenRM's hardcoded per-request params bypass RM rollout config; users cannot control sampling behavior or provide seed for reproducibility
3. **No VLM scoring path** — enabling RM without a custom reward function has no built-in visual scoring entry point, requiring manual configuration

## Design

Add `deterministic` and `seed` fields under `reward_model` in YAML config:

- **`deterministic=true`** — full determinism applied to RM actors via `enable_full_determinism(seed)` + `PYTHONHASHSEED` injected via `runtime_env`; `seed` propagated to GenRM HTTP requests (same seed + same input → same output, controlled pseudorandom)
- **`deterministic=false`** — full randomness, `seed` not propagated
- **Sampling params** (`temperature/do_sample/top_k/top_p`) are entirely user-configured in YAML under `rollout`, the `deterministic` flag does not override them

### 1. Floating-point determinism

RM inference uses `name=vllm` servers (`vLLMReplica` + `vLLMHttpServer`). When `deterministic=true`, monkey-patch `vLLMReplica.launch_servers()` with `_DeterministicServerProxy` which does two things:

**Env vars injection** — `PYTHONHASHSEED` must be set before Python process startup, so it is injected into RM actor `runtime_env`. Rollout actors remain unaffected (`is_reward_model` guard).

**Server class swap** — Replace `vLLMHttpServer` with `_DeterministicRMHttpServer(vLLMHttpServer)` subclass, which calls verl's `enable_full_determinism(seed)` in `_post_init`. This covers all in-process determinism: env vars (`CUBLAS_WORKSPACE_CONFIG`, `FLASH_ATTENTION_DETERMINISTIC`, `VLLM_BATCH_INVARIANT`, `NCCL_LAUNCH_MODE`, `NCCL_PROTO`), Python/NumPy/PyTorch seeds, deterministic algorithms, cuDNN config.

| Setting | Set by |
|---|---|
| `PYTHONHASHSEED=str(seed)` | `runtime_env` (before process startup) |
| `CUBLAS_WORKSPACE_CONFIG=:16:8` | `enable_full_determinism` (in-process) |
| `FLASH_ATTENTION_DETERMINISTIC=1` | `enable_full_determinism` (in-process) |
| `VLLM_BATCH_INVARIANT=1` | `enable_full_determinism` (in-process) |
| `NCCL_LAUNCH_MODE=GROUP` | `enable_full_determinism` (in-process) |
| `NCCL_PROTO=Simple` | `enable_full_determinism` (in-process) |
| `random.seed(seed)` | `enable_full_determinism` (in-process) |
| `np.random.seed(seed)` | `enable_full_determinism` (in-process) |
| `torch.manual_seed(seed)` | `enable_full_determinism` (in-process) |
| `torch.cuda.manual_seed_all(seed)` | `enable_full_determinism` (in-process) |
| `torch.use_deterministic_algorithms(True, warn_only=True)` | `enable_full_determinism` (in-process) |
| `cudnn.deterministic=True, benchmark=False, enabled=False` | `enable_full_determinism` (in-process) |

### 2. Configurable GenRM sampling + seed control

`VisualRewardManager` passes RM rollout config sampling params to `compute_score_ocr` via `sampling_params` kwarg, replacing hardcoded `DEFAULT_SAMPLING_PARAMS`. When `deterministic=true`, `seed` (from `reward_model.seed`) is included in `sampling_params` and propagates to vLLM's `/v1/chat/completions` request. When `deterministic=false`, `seed` is excluded — full randomness.

Sampling params (`temperature/do_sample/top_k/top_p`) are user-configured. Examples:
- `temperature=0, top_p=1.0` → greedy, fully deterministic
- `temperature>0, seed=42` → controlled pseudorandom, reproducible

### 3. VLM scoring path

When RM enabled + no `custom_reward_function.path`, auto-set path to `compute_score_ocr`. This provides the GenRM scoring entry point needed for VLM visual scoring.

### Propagation chain

```
reward.yaml (deterministic: true/false, seed: 42, rollout.temperature/do_sample/top_k/top_p)
  → DiffusionRayTrainer._init_online_rollout_stack()
      ├─ deterministic=true: monkey-patch vLLMReplica → _DeterministicServerProxy
      │     ├─ inject PYTHONHASHSEED into runtime_env
      │     └─ swap server_class → _DeterministicRMHttpServer (calls enable_full_determinism in _post_init)
      ├─ RM enabled + no custom_reward_function: auto-set → compute_score_ocr
  → VisualRewardManager
      ├─ sampling_params from rollout config (temperature, do_sample, top_k, top_p)
      ├─ deterministic=true: include seed from reward_model.seed
      ├─ deterministic=false: exclude seed
      → compute_score_ocr(sampling_params=...)
          → HTTP /v1/chat/completions request uses user-configured params
```

---

### 1. YAML config

**File**: `verl_omni/trainer/config/reward/reward.yaml`

Add `deterministic: false` and `seed: 42` under `reward_model`.

### 2. Floating-point determinism

**File**: `verl_omni/trainer/diffusion/ray_diffusion_trainer.py`

Define `_DeterministicRMHttpServer(vLLMHttpServer)` subclass. When `deterministic=true` + RM enabled, monkey-patch `vLLMReplica` with `_reward_deterministic` class attribute and `launch_servers()` override via `_DeterministicServerProxy`. `_DeterministicServerProxy` injects `PYTHONHASHSEED` into RM actor `runtime_env` and swaps `self.server_class` to `_DeterministicRMHttpServer`.

**File**: `verl_omni/workers/rollout/vllm_rollout/vllm_omni_async_server.py`

Remove `_DETERMINISTIC_RM_ENV_VARS` static dict.

### 3. Configurable GenRM sampling + seed control

**File**: `verl_omni/reward_loop/reward_manager/visual.py`

Pass `sampling_params` from RM rollout config. Include `seed` when `deterministic=true`, exclude when `deterministic=false`.

**File**: `verl_omni/utils/reward_score/genrm_ocr.py`

Add `sampling_params: dict = None` parameter. Use `sampling_params or DEFAULT_SAMPLING_PARAMS` in HTTP request.

---

## Backward Compatibility

- `deterministic=true` affects RM actors only; rollout unaffected
- `deterministic=false` preserves existing behavior
- `sampling_params` defaults to `None`; when `None`, `DEFAULT_SAMPLING_PARAMS` used
- `custom_reward_function` auto-set when RM enabled + no path provided

## Verification

1. `deterministic=true`: `PYTHONHASHSEED` injected via runtime_env + `enable_full_determinism(seed)` called in `_post_init`
2. Rollout actors unaffected (`is_reward_model` guard)
3. `deterministic=true` + `seed=42`: reproducible GenRM scores
4. `deterministic=false`: existing behavior

## Files

| File | Change |
|---|---|
| `verl_omni/trainer/config/reward/reward.yaml` | Add `deterministic: false` and `seed: 42` under `reward_model` |
| `verl_omni/trainer/diffusion/ray_diffusion_trainer.py` | `_DeterministicRMHttpServer` + monkey-patch `vLLMReplica` |
| `verl_omni/workers/rollout/vllm_rollout/vllm_omni_async_server.py` | Remove `_DETERMINISTIC_RM_ENV_VARS` |
| `verl_omni/reward_loop/reward_manager/visual.py` | Pass `sampling_params`; include/exclude `seed` |
| `verl_omni/utils/reward_score/genrm_ocr.py` | Add `sampling_params` param |
| `tests/reward_loop/test_visual_reward_manager.py` | `test_deterministic_reward_reproducibility` |

## Related Issues
Closes #97

## CC Lists
@SamitHuang @zhtmike

KaisennHu:
### What does this PR do?

Make VLM reward-model (GRM) inference bitwise-reproducible for identical inputs, reusing verl's full-determinism machinery (`enable_full_determinism`, batch invariance, deterministic routing). Three gaps are addressed:

1. **No floating-point determinism** — flash attention, CUDA matmul, and NCCL all-reduce introduce nondeterminism across RM inference runs.
2. **No batch invariance / routing determinism** — co-batched RM requests and least-loaded router tie-breaking make outputs depend on batch composition and arrival order.
3. **No GenRM seed control** — sampling in the GRM scorer was not reproducible; there was no way to pass a per-request `seed` through to vLLM.

Determinism is opt-in via `reward.reward_model.rollout.full_determinism` (keys already exist in reward.yaml) and enforced at three layers.

### Related Issues

Closes #111

### Test

`test_deterministic_reward_reproducibility_single_replica` and `_multi_replica`. Input: 3 identical image samples (prompt + `torch.randint` uint8 image + ground-truth text "OCR"), scored by the GRM (qwen3-vl transcribes the image; Levenshtein distance vs ground-truth yields `rm_scores ∈ [0,1]` and `genrm_response` text). Each run spins up its own `RewardLoopManager` (fresh RM server process(es) via `ray.init`/`ray.shutdown` in `try/finally`).

Asserts: two runs with `full_determinism=true` + the same seed produce bitwise-equal `rm_scores` (`torch.equal`) and identical `genrm_response`. `rm_scores` is a Levenshtein text-similarity scalar derived from the GRM's generated text — not logits or logprobs — so text equality ⟺ score equality. Multi-replica skips unless enough GPUs are available.

Batch invariance and multi-replica crc32 routing are exercised by the code path but not separately asserted: `genrm_ocr` sends one image per request serially (no explicit co-batching → `VLLM_BATCH_INVARIANT` not reliably triggerable), and every replica uses the same seed (routing differences are output-preserving).

### API and Usage Example

```yaml
reward:
  custom_reward_function:
    path: verl_omni/utils/reward_score/genrm_ocr.py
    name: compute_score_ocr
  reward_model:
    enable: True
    model_path: ~/models/qwen3-vl
    rollout:
      name: vllm_omni       # RM server whose _post_init applies determinism
      full_determinism: true
      seed: 42              # floating-point + per-request sampling seed
      tensor_model_parallel_size: 2
```

`reward.custom_reward_function` is required when the RM is enabled — the trainer raises `ValueError` otherwise (genrm must name its scorer explicitly).

### Design & Propagation Chain

```
reward.reward_model.rollout.full_determinism / seed (existing keys in reward.yaml)
  → main_diffusion.run_diffusion()
      ├─ _validate_grm_reward_function: RM enabled without custom_reward_function.path → ValueError
      └─ if _determinism_requested(config):   # rollout.full_determinism + RM enabled
            _export_full_determinism_env
            export switch flags VERL_FULL_DETERMINISM / VLLM_BATCH_INVARIANT / PYTHONHASHSEED
            → forwarded to every Ray actor via runtime_env (get_ppo_ray_runtime_env);
            floating-point vars (CUBLAS/FLASH/NCCL_*) are NOT exported here — they are
            set inside each actor's enable_full_determinism(), as in verl's main_ppo
  → NaiveRouter (multiprocessing.Process)
      └─ reads VERL_FULL_DETERMINISM=1 → crc32(request) tie-break among equally-loaded replicas
  → vLLMOmniHttpServer._post_init  (rollout.name=vllm_omni)
      └─ if config.full_determinism: enable_full_determinism(seed)
            + set VERL_SEED (read by rollout worker subprocesses) / VLLM_BATCH_INVARIANT
  → vLLMOmniARStrategy.preprocess_input
      └─ if config.full_determinism: inject per-request seed into SamplingParams
  → VisualRewardManager
      └─ if full_determinism: sampling_params = {max_tokens, seed} → compute_score_ocr
          → HTTP /v1/chat/completions request carries the seed
```

**1. Batch invariance** — the RM server sets `VLLM_BATCH_INVARIANT=1`, making a request's output independent of which other requests are co-batched. This applies to the generative `/v1/chat/completions` path the GRM uses (verl-omni's GRM does not use vLLM's `/classify` pooling endpoint, which batch invariance does not cover).

**2. Floating-point determinism** — `enable_full_determinism(seed)` is applied in `vLLMOmniHttpServer._post_init`, covering PYTHONHASHSEED, CUBLAS, flash-attn, NCCL (`NCCL_DETERMINISTIC`/`NCCL_ALGO=Ring`/`NCCL_PROTO=Simple`), all RNGs, `torch.use_deterministic_algorithms`, cuDNN config, and `VERL_DISABLE_FLASH_ATTN_CE=1` (which routes `logprobs_from_logits` through the pure-PyTorch `log_softmax+gather` path, bypassing the non-deterministic flash-attn Triton cross-entropy kernel). As in verl's `main_ppo`, only the switch flags are exported before `ray.init()` (they must be in the runtime env forward-list so actors and the router subprocess inherit them); the floating-point vars are set inside each actor's `enable_full_determinism()`, not in the main process.

**3. Per-request sampling seed** — `VisualRewardManager` forwards only `max_tokens` (response-length cap) plus, under determinism, the `seed` into each GRM request's `sampling_params`; vLLM's OpenAI-compatible endpoint honors it. The scorer keeps its own tuned sampling defaults (`genrm_ocr`: `temperature=0.7`, `top_p=0.8`, overridable via `GENRM_OCR_*` env vars), and an explicit `seed` opts in to reproducible sampling — without one, any env-inherited seed is dropped, so nondeterministic sampling remains the default.

### Files

| File | Change |
|---|---|
| `verl_omni/trainer/main_diffusion.py` | `_determinism_requested` gate + `_export_full_determinism_env` (switch flags before `ray.init()`) + `_validate_grm_reward_function` (require scorer when RM enabled) |
| `verl_omni/workers/rollout/vllm_rollout/vllm_omni_async_server.py` | `vLLMOmniHttpServer._post_init` determinism hook |
| `verl_omni/workers/rollout/vllm_rollout/vllm_omni_ar_strategy.py` | Per-request seed injection in `preprocess_input` |
| `verl_omni/reward_loop/reward_manager/visual.py` | Pass `sampling_params` (`max_tokens` + determinism `seed`) to the scorer |
| `verl_omni/utils/reward_score/genrm_ocr.py` | Accept `sampling_params` kwarg; seed opt-in semantics |
| `tests/reward_loop/test_visual_reward_manager.py` | Same-seed bitwise-reproducibility tests (single/multi replica); GPU-count skip; `try/finally` around `ray.init` |
| `docs/algo/deterministic_post_training.md` + `docs/index.md` | New doc on enabling and verifying reward determinism; toctree entry |

### DIFF
--- a/docs/algo/deterministic_post_training.md
+++ b/docs/algo/deterministic_post_training.md
@@ -1,0 +1,143 @@
+# Deterministic Post-Training
+
+Last updated: 08/15/2026.
+
+By default, verl-omni RL training is **not bitwise reproducible**: identical configs run twice can produce different reward curves due to nondeterminism in GPU kernels, batch composition, request routing, and sampling. This page documents the effort to make post-training deterministic across multiple aspects (reward inference, rollout generation, diffusion sampling, and eventually end-to-end bitwise-aligned reward curves).
+
+**Supported: reward inference determinism** (VLM reward-model / GRM scoring). The remaining aspects are tracked as future work — see [Scope](#scope).
+
+## Scope
+
+| Component | Status |
+|---|---|
+| GRM inference — floating-point determinism | ✅ Supported |
+| GRM inference — batch invariance (`VLLM_BATCH_INVARIANT`) | ✅ Supported |
+| GRM inference — deterministic multi-replica routing (`NaiveRouter` crc32) | ✅ Supported |
+| GRM inference — per-request sampling seed | ✅ Supported |
+| Actor-rollout generation determinism | 🚧 Future |
+| Diffusion sampling determinism | 🚧 Future |
+| End-to-end bitwise-aligned reward curves | 🚧 Future |
+
+For the upstream full-determinism feature (actor rollout, FSDP/Megatron, trainer backend), see verl's [`docs/advance/determinism.md`](https://github.com/verl-project/verl/blob/main/docs/advance/determinism.md). verl-omni will adopt those layers incrementally, and a single top-level determinism switch (covering both rollout and reward) will replace the current per-component `full_determinism` flags.
+
+## When to use
+
+- **Debugging** — reproduce a reward-score anomaly exactly.
+- **Regression testing** — verify a code change has no silent effect on GRM scores.
+- **Research** — fair comparison of reward-side algorithmic changes.
+
+Leave determinism off for production training — deterministic kernels are slower, cuDNN benchmarking is disabled, and batch invariance can serialize.
+
+## Enable
+
+Set `reward.reward_model.rollout.full_determinism=true` and a `seed` under the `reward_model.rollout` block:
+
+```yaml
+reward:
+  reward_model:
+    enable: true
+    rollout:
+      name: vllm_omni
+      full_determinism: true
+      seed: 42
+```
+
+Or via Hydra overrides:
+
+```bash
+python -m verl_omni.trainer.main_diffusion \
+  reward.reward_model.enable=true \
+  reward.reward_model.rollout.full_determinism=true \
+  reward.reward_model.rollout.seed=42 \
+  [other config overrides...]
+```
+
+Sampling params (`temperature`, `top_k`, `top_p`) remain user-configured under `reward.reward_model.rollout`; the flag does not override them. Greedy decoding uses `temperature=0`; controlled pseudorandom sampling uses `temperature>0` plus the per-request `seed`. The scorer's own tuned defaults (e.g. genrm_ocr's `temperature=0.7`, `top_p=0.8`) apply unless overridden.
+
+## Configuration Reference
+
+| Parameter | Default | Description |
+|---|---|---|
+| `reward_model.rollout.full_determinism` | `false` | Enables reward inference determinism (all layers below) |
+| `reward_model.rollout.seed` | `42` | Floating-point + per-request sampling seed |
+| `reward_model.rollout.name` | — | Use `vllm_omni` (the server with the `_post_init` determinism hook) |
+| `reward_model.rollout.temperature` | — | User-configured sampling temperature |
+| `reward_model.rollout.top_k` | — | User-configured top-k |
+| `reward_model.rollout.top_p` | — | User-configured top-p |
+
+## How it works
+
+verl-omni's GRM scores images via `/v1/chat/completions` (a VLM such as Qwen3-VL transcribes the image). Reproducibility is enforced at three layers.
+
+### Env export before `ray.init()`
+
+When determinism is requested (currently `reward.reward_model.rollout.full_determinism=true`; a single top-level switch covering rollout + reward is future work — see [Scope](#scope)), `main_diffusion.run_diffusion` calls `_export_full_determinism_env` in the main process **before** `ray.init()` to set the determinism **switch** env vars:
+
+- `VERL_FULL_DETERMINISM=1`, `VLLM_BATCH_INVARIANT=1`, `PYTHONHASHSEED=<seed>`
+
+These are forwarded to every Ray actor via verl's `get_ppo_ray_runtime_env` (the runtime_env forward-list). They must be set before `ray.init()` so actors inherit them — too late if only set inside `_post_init`. The `NaiveRouter` subprocess (a `multiprocessing.Process` forked from an actor) reads `VERL_FULL_DETERMINISM` to activate crc32 routing, and the vLLM server reads `VLLM_BATCH_INVARIANT` to activate batch invariance.
+
+Only switch flags are exported here (mirroring verl's `main_ppo`); the floating-point vars (`CUBLAS_WORKSPACE_CONFIG`, `FLASH_ATTENTION_DETERMINISTIC`, `NCCL_DETERMINISTIC`, `NCCL_ALGO`, `NCCL_PROTO`, `VERL_DISABLE_FLASH_ATTN_CE`) are set inside each actor's `enable_full_determinism()`, so they do not pollute the normal training flow's environment.
+
+### 1. Batch invariance
+
+`VLLM_BATCH_INVARIANT=1` makes a request's output independent of which other requests are co-batched with it. This applies to the generative `/v1/chat/completions` path the GRM uses (unlike vLLM's `/classify` pooling endpoint, which is not covered by batch invariance — that path would require serializing with `max_num_seqs=1`; verl-omni's GRM does not use it).
+
+Coverage is model- and hardware-dependent — see the [vLLM batch invariance docs](https://docs.vllm.ai/en/latest/features/batch_invariance/) (and [tested models](https://docs.vllm.ai/en/latest/features/batch_invariance/#tested-models)). If not covered, set `max_num_seqs=1` to serialize.
+
+### 2. Floating-point determinism
+
+The RM server (`vLLMOmniHttpServer`, reached via `rollout.name=vllm_omni`) applies `enable_full_determinism(seed)` in `_post_init`, before model load:
+
+| Setting | Effect |
+|---|---|
+| `PYTHONHASHSEED` | freezes Python `hash()` / dict ordering |
+| `CUBLAS_WORKSPACE_CONFIG=:16:8` | deterministic cuBLAS |
+| `FLASH_ATTENTION_DETERMINISTIC=1` | deterministic flash-attn |
+| `NCCL_DETERMINISTIC=1` / `NCCL_ALGO=Ring` / `NCCL_PROTO=Simple` | deterministic NCCL |
+| `VERL_DISABLE_FLASH_ATTN_CE=1` | bypass the non-deterministic flash-attn Triton cross-entropy kernel in `logprobs_from_logits` (pure-PyTorch `log_softmax+gather`); not covered by `FLASH_ATTENTION_DETERMINISTIC` (attention backward only) or `torch.use_deterministic_algorithms` (Triton custom ops skip `warn_only`) |
+| `VERL_SEED` | read by rollout worker subprocesses (verl `vllm_rollout/utils.py`) to apply `enable_full_determinism` per worker |
+| seeded `random` / `numpy` / `torch` / `cuda` RNGs | fixed RNG state |
+| `torch.use_deterministic_algorithms(True, warn_only=True)` | deterministic ops |
+| `cudnn.deterministic=True, benchmark=False` | deterministic cuDNN |
+
+`full_determinism`/`seed` live on the `reward_model` block (a non-structured config node), so the trainer propagates them onto `reward_model.rollout` via `OmegaConf.open_dict` — the server process's `self.config` is the rollout config and can read them.
+
+Each RM replica uses the same `seed` (not `replica_rank + seed`), so every replica starts from identical state.
+
+### 3. Deterministic routing
+
+The reward router (verl's `NaiveRouter`) is least-loaded; under `VERL_FULL_DETERMINISM=1` it tie-breaks among equally-loaded replicas with `crc32(request_body) % len(candidates)`, so the same request lands on the same replica across runs. Combined with identical per-replica seeds and batch invariance, the same request yields the same output regardless of which replica serves it.
+
+### 4. Per-request sampling seed
+
+`VisualRewardManager` injects `seed = reward_model.rollout.seed` into each GRM request's `sampling_params`. vLLM's OpenAI-compatible endpoint honors the `seed` field, so the same image + same sampling params + same seed yields the same transcription, and therefore the same reward score. This is **controlled pseudorandom** sampling, not greedy — `temperature`/`top_k`/`top_p` are respected, and the seed makes the randomness reproducible.
+
+`do_sample` is **not** a valid vLLM `SamplingParams` key (silently ignored), so it is not passed. `max_tokens` is always supplied (default `4096`) so generation length is stable.
+
+### VLM scoring path
+
+Enabling the RM requires an explicit `reward.custom_reward_function` — the trainer raises `ValueError` when `reward.reward_model.enable=true` without one. Use the OCR GRM scorer (`verl_omni/utils/reward_score/genrm_ocr.py::compute_score_ocr`) or provide your own scorer via `path`/`name`.
+
+## Side Effects
+
+- **Performance**: deterministic kernels are slower and cuDNN benchmarking is disabled; expect throughput loss.
+- **Recommendation**: enable for debugging, regression testing, or research only.
+
+## Limitations
+
+- **Batch invariance not directly asserted**: `genrm_ocr` sends one image per `/v1/chat/completions` request serially, so requests are not explicitly co-batched and `VLLM_BATCH_INVARIANT` cannot be reliably triggered or observed in the test. Batch invariance is still set by `_post_init` for correctness; its coverage is model/hardware-dependent (see [vLLM batch invariance docs](https://docs.vllm.ai/en/latest/features/batch_invariance/)).
+- **Nondeterministic fallbacks**: some GPU ops have no deterministic implementation. `warn_only=True` emits warnings; these are expected.
+- **Multi-replica routing not asserted**: every replica uses the same seed, so replica outputs are identical regardless of routing, and crc32 routing (gated on `VERL_FULL_DETERMINISM`) cannot be distinguished from least-loaded routing by the assertions. Routing correctness is guaranteed by code, not by the test suite.
+- **Hardware**: deterministic ops and batch invariance require specific hardware — see the [vLLM batch invariance docs](https://docs.vllm.ai/en/latest/features/batch_invariance/). On unsupported hardware, set `max_num_seqs=1`.
+- **Generative GRM only**: this covers the generative VLM reward path. Full E2E determinism (rollout + diffusion sampling) is future work.
+
+## Verify
+
+```bash
+pytest tests/reward_loop/test_visual_reward_manager.py -v -s -k deterministic_reward_reproducibility
+```
+
+Each run spins up its own `RewardLoopManager` (fresh RM server process(es) via `ray.init`/`ray.shutdown`). Floating-point determinism (`enable_full_determinism`), `VERL_SEED`, and `VLLM_BATCH_INVARIANT=1` are all set by the RM server's `_post_init` (gated on `full_determinism=true` in the rollout config, which the test propagates via `OmegaConf.open_dict`), so no determinism env vars are needed in the Ray runtime_env. Two variants cover single replica (`test_deterministic_reward_reproducibility_single_replica`) and multi replica (`test_deterministic_reward_reproducibility_multi_replica`). Each asserts:
+
+- **Same seed → bitwise-aligned**: two independent runs with the same seed produce bitwise-equal `rm_scores` (`torch.equal`) and identical `genrm_response`.
--- a/tests/reward_loop/test_visual_reward_manager.py
+++ b/tests/reward_loop/test_visual_reward_manager.py
@@ -14,6 +14,7 @@
 
 import os
 
+import pytest
 import ray
 import torch
 from hydra import compose, initialize_config_dir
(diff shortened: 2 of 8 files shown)


verl_omni/reward_loop/reward_manager/visual.py
```diff
@@ -37,6 +37,9 @@ def __init__(self, config, tokenizer, compute_score, reward_router_address=None,
         self.reward_router_address = reward_router_address
         self.reward_model_tokenizer = reward_model_tokenizer
 
+        deterministic = config.reward.reward_model.get("deterministic", False)
+        self._genrm_sampling_params = {"temperature": 0.0, "top_p": 1.0} if deterministic else None
```
zhtmike: I think we can simply set `"temperature": 0.0, "top_p": 1.0` in the shell scripts to enable reward determinism. It will looks clear.

WDYT?

KaisennHu: hi @zhtmike, shell scripts can set sampling server defaults for name=vllm RM servers, but can't cover:
1. Per-request sampling — vllm_omni RM servers ignore override_generation_config; only sampling_params kwarg works.
2. Env vars + VLM scoring auto-set — must be injected at Ray actor launch time, can't be done from CLI args.

deterministic toggle covers all three levels.

zhtmike: OK, btw why do we use vllm_omni for VLM scoring? it should be vllm for VLM?

KaisennHu: You're right. VLM RM scoring uses name=vllm, not vllm_omni. For name=vllm, override_generation_config already works. The sampling_params per-request override is an extra safeguard, and the only effective path for name=vllm_omni RM servers (future scenario). Shell scripts still can't scope env vars to RM-only actors or auto-set VLM scoring — both require the deterministic toggle.

zhtmike: Understood. Then let us focus on vLLM VLM now, pls drop unnecessary file change

SamitHuang (commented):
Looks clear to me! Pls supplement the test results for 1) multiple runs with same seed, 2) multiple runs with different seeds

SamitHuang:
It's a good starting point for deterministic flowgrpo. It's better to add a doc to illustarte how to config and enable determinism.

verl_omni/reward_loop/reward_manager/visual.py
```diff
@@ -58,11 +58,23 @@ async def run_single(self, data: DataProto) -> dict:
         extra_info["num_turns"] = num_turns
         extra_info["rollout_reward_scores"] = rollout_reward_scores
 
+        rm_rollout = self.config.reward.reward_model.rollout
+        sampling_params = {
+            "temperature": rm_rollout.temperature,
+            "do_sample": rm_rollout.do_sample,
```
wtomin: I suppose the rollout backend for reward is `vllm`. The latest main branch of vllm `SamplingParams` has no key name `do_sample`: https://github.com/vllm-project/vllm/blob/main/vllm/sampling_params.py#L199. `do_sample` maybe silently ignored.

I think `do_sample=False` and `temperature=0` are equivalent. Please correct me if I am wrong.

verl_omni/reward_loop/reward_manager/visual.py
```diff
@@ -58,11 +58,23 @@ async def run_single(self, data: DataProto) -> dict:
...
+            "top_k": rm_rollout.top_k,
+            "top_p": rm_rollout.top_p,
```
wtomin: I notice that in `genrm_ocr.py`, `DEFAULT_SAMPLING_PARAMS = {"temperature": 0.7, "top_p": 0.8, "max_tokens": 4096}`. Here the `sampling_params` has no `max_tokens` argument. Maybe this can cause generation difference, if `max_tokens` default value is not 4096.

verl_omni/trainer/diffusion/ray_diffusion_trainer.py
```diff
@@ -632,6 +632,78 @@ def _init_colocated_workers(self):
 
     def _init_online_rollout_stack(self, actor_rollout_resource_pool):
         """Initialize rollout, reward, and checkpoint engines (online sampling only)."""
+        # Deterministic reward model inference when `deterministic=true`
+        from verl.workers.rollout.vllm_rollout.vllm_async_server import vLLMHttpServer, vLLMReplica
```
wtomin: The readability and maintainability of the monkey-patched solution inside `ray_diffusion_trainer.py‎` is relatively poor.

I suggest to introduce extended classes based on `vLLMReplica / vLLMHttpServer` under `verl_omni/workers/rollout/vllm_rollout`: `enable_full_determinism` in the extended classes. Also hearing suggestions from @SamitHuang @zhtmike @AndyZhou952 .

zhtmike: > The readability and maintainability of the monkey-patched solution inside `ray_diffusion_trainer.py‎` is relatively poor.
> 
> I suggest to introduce extended classes based on `vLLMReplica / vLLMHttpServer` under `verl_omni/workers/rollout/vllm_rollout`: `enable_full_determinism` in the extended classes. Also hearing suggestions from @SamitHuang @zhtmike @AndyZhou952 .

I think this PR is slighted outdated. Per offline discussion, we will follow the design pattern in https://github.com/verl-project/verl/pull/6572

verl_omni/trainer/diffusion/ray_diffusion_trainer.py
```diff
@@ -632,6 +632,78 @@ def _init_colocated_workers(self):
...
+                            """Inject PYTHONHASHSEED + seed signal into runtime_env and swap to deterministic server."""
+
+                            def __init__(self, deterministic_server_class, rm_seed):
+                                self.original = deterministic_server_class
+                                self._rm_seed = rm_seed
+
+                            def options(self, **kwargs):
+                                runtime_env = kwargs.setdefault("runtime_env", {})
+                                env_vars = runtime_env.setdefault("env_vars", {})
+                                env_vars["PYTHONHASHSEED"] = str(self._rm_seed)
```
wtomin: I suggest to clearly explain the global envs related to deterministic settings in the documents. As for the document, I suggest to create a `docs/algo/deterministic_post_training.md`.

verl_omni/reward_loop/reward_manager/visual.py
KaisennHu: We set temperature for pseudorandom

verl_omni/trainer/main_diffusion.py
```diff
@@ -55,6 +93,13 @@ def run_diffusion(config, task_runner_class=None) -> None:
                 settings, model paths, and training hyperparameters.
         task_runner_class: For recipe to change TaskRunner.
     """
+    # Export floating-point determinism env vars BEFORE ray.init() so
+    # get_ppo_ray_runtime_env() forwards them to every Ray actor at startup,
+    # ahead of each actor importing torch / initializing NCCL. Setting them
+    # later inside enable_full_determinism() (after the actor is up) is too
+    # late for NCCL all-reduce and cuBLAS matmul.
+    _export_full_determinism_env(config)
```
knlnguyen1802: I think we should not do this for normal flow of diffusion training. Even thought there is no effect now it can become complicated later when deal with environment variable

verl_omni/trainer/diffusion/ray_diffusion_trainer.py
```diff
@@ -734,6 +780,24 @@ def _init_colocated_workers(self):
...
         """Initialize rollout, reward, and checkpoint engines (online sampling only)."""
+        # Auto-set VLM scoring function when RM enabled + no custom_reward_function provided
+        if self.use_rm:
+            custom_fn_cfg = self.config.reward.custom_reward_function
+            if custom_fn_cfg.path is None:
+                from omegaconf import OmegaConf
+
+                with OmegaConf.open_dict(custom_fn_cfg):
+                    custom_fn_cfg.path = "verl_omni/utils/reward_score/genrm_ocr.py"
+                    custom_fn_cfg.name = "compute_score_ocr"
```
knlnguyen1802: Better to move this into script instead of hardcode here

zhtmike:
can you resolve conflict?

docs/algo/deterministic_post_training.md
```diff
@@ -0,0 +1,143 @@
...
+
+### 4. Per-request sampling seed
+
+`VisualRewardManager` injects `seed = reward_model.seed` into each GRM request's `sampling_params`. vLLM's OpenAI-compatible endpoint honors the `seed` field, so the same image + same sampling params + same seed yields the same transcription, and therefore the same reward score. This is **controlled pseudorandom** sampling, not greedy — `temperature`/`top_k`/`top_p` are respected, and the seed makes the randomness reproducible.
+
+`do_sample` is **not** a valid vLLM `SamplingParams` key (silently ignored), so it is not passed. `max_tokens` is always supplied (default `4096`) so generation length is stable.
+
+### VLM scoring path
+
+When RM is enabled, set the scoring function in the training script (not hardcoded in the trainer): `reward.custom_reward_function.path=verl_omni/utils/reward_score/genrm_ocr.py`, `name=compute_score_ocr`.
```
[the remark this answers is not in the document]
zhtmike: can you fix this @KaisennHu ?

tests/reward_loop/test_visual_reward_manager.py
```diff
@@ -65,52 +70,140 @@ def test_reward_model_genrm():
                 "VLLM_USE_V1": "1",
                 "FLASHINFER_DISABLE_VERSION_CHECK": os.environ.get("FLASHINFER_DISABLE_VERSION_CHECK", "1"),
             }
-        },
-        ignore_reinit_error=True,
+        }
     )
```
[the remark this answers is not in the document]
zhtmike: And this

verl_omni/trainer/main_diffusion.py
```diff
@@ -70,6 +70,42 @@ def main(config):
...
+    Only switch flags (VERL_FULL_DETERMINISM, VLLM_BATCH_INVARIANT, PYTHONHASHSEED);
+    floating-point vars are set in each actor's enable_full_determinism(), as in verl.
+    """
+    rm_cfg = config.reward.reward_model
+    os.environ["VERL_FULL_DETERMINISM"] = "1"
+    os.environ["VLLM_BATCH_INVARIANT"] = "1"
+    os.environ["PYTHONHASHSEED"] = str(rm_cfg.get("seed", 42))
+
+
+def _maybe_default_grm_reward_function(config) -> None:
```
zhtmike: I am not sure it is a good idea, maybe just raise error directly. @knlnguyen1802 pls take a look

zhtmike:
The failing check is `doc (3.10)`, and it is **not** caused by the scary-looking `TypeError: unsupported operand type(s) for |: 'OmniDiffusionRequest' and 'DiffusionRequestBatch'` autodoc warnings — those same warnings also appear in the green doc run on `main` (they are pre-existing, tolerated Sphinx WARNINGs from importing `verl_omni.pipelines` with the pinned vllm-omni on Python 3.10). The real failure comes from the new doc page added by this PR: `docs/algo/deterministic_post_training.md` is not referenced in any toctree, so Sphinx logs `WARNING: document isn't included in any toctree`, and the "Build docs" step in `.github/workflows/doc.yml` explicitly runs `grep -q "WARNING: document isn't included in any toctree" _build/sphinx.log` and exits 1 when it matches.

**How to fix:** add one line to the "Algorithms" toctree in `docs/index.md` (around line 58–65):

```
algo/deterministic_post_training.md
```

e.g. right after `algo/diffusion_opd.md`. Commit and push — the doc CI will then pass (you can verify locally with `cd docs && make clean && make html SPHINXOPTS="--keep-going -w _build/sphinx.log"` and checking that `_build/sphinx.log` contains no "isn't included in any toctree" warning).

knlnguyen1802 (changes requested):
Thanks for pushing this through — the per-request-seed + config-driven determinism is the right direction, and reusing enable_full_determinism via _post_init instead of the old monkey-patch is much cleaner. Before we can merge, pls address:
1. Don't add reward_model.full_determinism/seed — verl already reads reward_model.rollout.full_determinism/seed (they're in reward.yaml). Use those and make _determinism_requested read the rollout key, so the two can't diverge.
2. The unconditional sampling_params injection in visual.py silently changes default GRM sampling (temp 0.7→1.0, top_p 0.8→1.0) for everyone, even with determinism off. Gate it on full_determinism or keep the tuned defaults.

verl_omni/trainer/main_diffusion.py
knlnguyen1802: @KaisennHu I agree with Mike, maybe can just raise error on this.

knlnguyen1802:
@KaisennHu Other part is ok please resolve all remaining comment
And also update with a latest to verify the deterministic of reward

KaisennHu:
> Other part is ok please resolve all remaining comment
> And also update with a latest to verify the deterministic of reward

The test has already passed. Please take another look!

knlnguyen1802 (approved):

zhtmike (approved):
Nice work!

Issue #124, opened 2026-05-31
[RFC]: MFU metric for diffusion RL training


SamitHuang:
## 1. Motivation

Upstream [`verl`](https://github.com/verl-project/verl) reports Model FLOPs Utilization (MFU) for the actor under three keys:

| key                       | what is measured                              |
| ------------------------- | --------------------------------------------- |
| `perf/mfu/actor`          | actor training step (forward + backward)      |
| `perf/mfu/actor_infer`    | actor log-prob recompute (forward only)       |
| `perf/mfu/ref`            | reference policy log-prob (forward only)      |

MFU is the only metric that lets users compare hardware/compiler efficiency across runs independent of model size, batch size, or sequence length.

In `verl-omni` the diffusion training stack (Qwen-Image FlowGRPO / MixGRPO / DanceGRPO / SD3-DPO / DiffusionNFT) currently **does not** emit any MFU metric. The relevant block in `verl_omni/workers/engine_workers.py` reads:

```python
if hasattr(self.model_config, "hf_config"):
    self.flops_counter = FlopsCounter(self.model_config.hf_config)
else:
    # for Diffusion models, FlopsCounter is not supported yet.
    self.flops_counter = None
```

i.e. the moment `model_type` is `diffusion_model` / `diffusion_dp_model` MFU is silently dropped because `verl.utils.flops_counter.FlopsCounter` only understands HuggingFace LLM `PretrainedConfig` schemas (`hidden_size`, `num_hidden_layers`, `intermediate_size`, …).

This RFC proposes a **diffusion-aware MFU counter** that hooks into the same `_postprocess_output` path the LLM trainer already uses, so users get the familiar `actor/mfu`, `actor_infer/mfu`, `ref/mfu` keys for diffusion RL runs as well.

## 2. Why upstream `FlopsCounter` cannot be reused as-is

| Aspect              | LLM (`FlopsCounter`)                                  | Diffusion DiTs (Qwen-Image, SD3, Flux, Wan2.2, Z-Image, …)                  |
| ------------------- | ----------------------------------------------------- | --------------------------------------------------------------------------- |
| Config schema       | `transformers.PretrainedConfig`                       | `diffusers.ConfigMixin` (FrozenDict). Field names diverge across families: `num_layers` vs `n_layers`, `attention_head_dim` vs `dim_head`, `ffn_dim` vs implicit `4·dim`. |
[...]

PR #128, opened 2026-06-01
[diffusion, perf, doc] feat: add MFU metric support for diffusion models

SamitHuang:
### What does this PR do?

Implement #124 

Introduces Model FLOPs Utilization (MFU) tracking for diffusion models in `verl-omni`. It provides a robust, extensible framework for calculating FLOPs across different diffusion architectures (T2I, T2V, Edit, etc.) and accurately reporting MFU during RL training.

Key additions and fixes:
- **Core MFU Counter**: Adds `DiffusionFlopsCounter` and a class-based registry (`DiffusionModelFlops`) for architecture-specific FLOPs estimation.
- **Qwen-Image Support**: Implements `QwenImageFlops` with accurate FLOPs formulas for dual-stream DiTs, including a fix to correctly extract sequence lengths from the
`diffusers._pack_latents` layout (`(B, [T_steps,] L, C')`).
- **Distributed Gathering**: Adds `_allgather_diffusion_flops_meta` in `TrainingWorker` to correctly aggregate sequence lengths across DP ranks, preventing MFU inflation.
- **Documentation & Tuning**: Adds comprehensive documentation (`docs/perf/diffusion_mfu.md`) including an MFU tuning guide, and an optimized H200 launch script
(`examples/flowgrpo_trainer/run_qwen_image_ocr_h200.sh`).

### Checklist Before Starting

### Test

1. **Unit Tests**: Added 44 tests in `tests/utils/test_diffusion_flops_counter_on_cpu.py` covering:
   - Qwen-Image FLOPs formula (dense + attention).
   - DP gathering of sequence lengths (`TestDPGlobalConsistency`).
   - Packed latents shape extraction (`TestQwenImageFlopsPackedLatents`).
   - Device peak env var override (`TestDevicePeakOverride`).
2. **End-to-End Validation**: Ran `flowgrpo` training with Qwen-Image on a 4x H200 cluster.
   - Verified MFU values are physically accurate using the `VERL_OMNI_DEVICE_FLOPS_TFLOPS=989` override.
   - Achieved `actor/mfu=0.223` and `actor_infer/mfu=0.327` using the H200 running script `examples/flowgrpo_trainer/run_qwen_image_ocr_h200.sh`

### API and Usage Example

**1. Pinning the Device Peak (for mis-detected GPUs like H200):**
```bash
export VERL_OMNI_DEVICE_FLOPS_TFLOPS=989
```

**2. Adding support for a new architecture:**
```python
from verl_omni.utils.diffusion_flops_counter import DiffusionModelFlops, register_diffusion_architecture

@register_diffusion_architecture("MyNewDiTPipeline")
class MyNewDiTFlops(DiffusionModelFlops):
    def estimate_flops(self, latent_seqlens, prompt_seqlens, delta_time, num_timesteps, cfg_passes):
        # ... custom FLOPs formula ...
        return flops_per_sample, promised_peak

    def get_latent_seqlens(self, data, config=None):
        # Override if latents are packed or concatenated with references
        return super().get_latent_seqlens(data, config)
```

### Design & Code Changes

- `verl_omni/utils/diffusion_flops_counter.py`: Contains the core `DiffusionFlopsCounter`, the `DiffusionModelFlops` base class, and the `QwenImageFlops` implementation.
- `verl_omni/workers/engine_workers.py`: Integrates the counter into `TrainingWorker._postprocess_output`, handles DP gathering via `_allgather_diffusion_flops_meta`, and
introduces the `record_loss` flag.
- `docs/perf/diffusion_mfu.md`: Detailed documentation on the FLOPs formulas, how the counter is wired, and a guide for tuning MFU on H200.

> [!IMPORTANT]


verl_omni/utils/diffusion_flops_counter.py
```diff
@@ -0,0 +1,676 @@
...
+
+    # Guidance-distilled models (e.g. Flux) run only 1 pass
+    if tcfg.get("guidance_embeds"):
+        return 1
+
+    # True-CFG or standard CFG scales > 1.0 require 2 passes
+    if _gt_one(pcfg.get("true_cfg_scale")) or _gt_one(pcfg.get("guidance_scale")):
+        return 2
+
+    return 1
```
zhtmike: `resolve xxx` is too AI

verl_omni/workers/engine_workers.py
```diff
@@ -62,6 +66,35 @@
 logger.setLevel(os.getenv("VERL_LOGGING_LEVEL", "WARN"))
 
 
+def _load_diffusion_transformer_config(model_config: "DiffusionModelConfig") -> Optional[dict]:
```
zhtmike: too AI ...

verl_omni/trainer/diffusion/ray_diffusion_trainer.py
```diff
@@ -70,6 +70,22 @@
 sys_logger = logging.getLogger(__name__)
 
 
+def _extract_infer_metrics(output: Any) -> dict[str, Any]:
```
zhtmike: too AI again.. check how verl implement this

zhtmike:
btw how to ensure the MFU is calculated correctly? ask another AI to have a review?

zhtmike:
> btw how to ensure the MFU is calculated correctly? ask another AI to have a review?

how about this, write a simple model, like the a small model with 1 or 2 layer of Qwen-image, use `nsys profile` to get the actual FLOPS, then we can check if MFU is correct

verl_omni/utils/diffusion_flops_counter.py
SamitHuang: renamed to get_forward_passes_per_step for clarity

verl_omni/workers/engine_workers.py
SamitHuang: it's quite straightforward

zhtmike: Engine worker is not for diffusers

zhtmike: And why dont parse config under config post_init directly?

verl_omni/trainer/diffusion/ray_diffusion_trainer.py
```diff
@@ -862,7 +869,10 @@ def _compute_ref_log_prob(self, batch: DataProto) -> DataProto:
         ref_log_prob = tu.get_tensordict(
             {"ref_log_prob": log_probs.float(), "ref_prev_sample_mean": prev_sample_mean.float()}
         )
-        return DataProto.from_tensordict(ref_log_prob)
+        # Propagate per-call MFU / loss metrics computed in the worker.
+        infer_metrics = _extract_infer_metrics(output)
+        meta_info = {"metrics": infer_metrics} if infer_metrics else {}
+        return DataProto.from_tensordict(ref_log_prob, meta_info=meta_info)
```
zhtmike: dont calling this functions many times, you can check how verl's calculate mfu

verl_omni/trainer/diffusion/ray_diffusion_trainer.py
```diff
@@ -881,7 +891,9 @@ def _compute_old_log_prob(self, batch: DataProto):
         if prev_sample_mean is not None:
             old_log_prob_dict["old_prev_sample_mean"] = prev_sample_mean.float()
         old_log_prob = tu.get_tensordict(old_log_prob_dict)
-        return DataProto.from_tensordict(old_log_prob)
+        infer_metrics = _extract_infer_metrics(output)
+        meta_info = {"metrics": infer_metrics} if infer_metrics else {}
+        return DataProto.from_tensordict(old_log_prob, meta_info=meta_info)
```
zhtmike: same

verl_omni/trainer/diffusion/ray_diffusion_trainer.py
```diff
@@ -1002,6 +1014,9 @@ def fit(self):
                     else:  # Recompute old_log_probs
                         with marked_timer("old_log_prob", timing_raw, color="blue"):
                             old_log_prob = self._compute_old_log_prob(batch)
+                            metrics.update(
+                                rename_dict(reduce_metrics(old_log_prob.meta_info.get("metrics", {})), "actor_infer/")
+                            )
```
zhtmike: same

verl_omni/trainer/diffusion/ray_diffusion_trainer.py
```diff
@@ -1019,6 +1034,9 @@ def fit(self):
                         # compute reference log_prob
                         with marked_timer(str(Role.RefPolicy), timing_raw, color="olive"):
                             ref_log_prob = self._compute_ref_log_prob(batch)
+                            metrics.update(
+                                rename_dict(reduce_metrics(ref_log_prob.meta_info.get("metrics", {})), "ref/")
+                            )
```
zhtmike: same

verl_omni/trainer/diffusion/ray_diffusion_trainer.py
```diff
@@ -1254,7 +1272,9 @@ def _compute_ref_noise_pred(self, batch: DataProto) -> Optional[DataProto]:
             "noise": noise.float(),
             "timesteps": timesteps.float(),
         }
-        return DataProto.from_tensordict(tu.get_tensordict(ref_output))
+        infer_metrics = _extract_infer_metrics(output)
+        meta_info = {"metrics": infer_metrics} if infer_metrics else {}
+        return DataProto.from_tensordict(tu.get_tensordict(ref_output), meta_info=meta_info)
```
zhtmike: same

verl_omni/trainer/diffusion/ray_diffusion_trainer.py
```diff
@@ -1351,6 +1371,9 @@ def fit(self):
                         with marked_timer(str(Role.RefPolicy), timing_raw, color="olive"):
                             ref_dpo = self._compute_ref_noise_pred(batch)
                             if ref_dpo is not None:
+                                metrics.update(
+                                    rename_dict(reduce_metrics(ref_dpo.meta_info.get("metrics", {})), "ref/")
+                                )
```
zhtmike: same

verl_omni/utils/diffusion_flops_counter.py
```diff
@@ -0,0 +1,674 @@
...
+
+# ---------------------------------------------------------------------------
+# Registry
+# ---------------------------------------------------------------------------
+
+
+_REGISTRY: dict[str, type[DiffusionModelFlops]] = {}
+
+
+def register_diffusion_architecture(
```
zhtmike: why need to register diffusion architecture here???

verl_omni/utils/diffusion_flops_counter.py
```diff
@@ -0,0 +1,674 @@
...
+    ``module.config``, which behaves like a plain ``dict``.  We additionally
+    accept raw dicts for testing.
+    """
+    if isinstance(config, Mapping):
+        return config
+    if hasattr(config, "to_dict"):
+        return config.to_dict()
+    if hasattr(config, "items"):
+        return dict(config.items())
+    raise TypeError(f"DiffusionFlopsCounter expects a dict-like transformer config, got {type(config).__name__}.")
```
zhtmike: ai code

verl_omni/utils/diffusion_flops_counter.py
```diff
@@ -0,0 +1,674 @@
...
+            return getter(key, None)
+        except TypeError:
+            try:
+                return getter(key)
+            except Exception:
+                return None
+    try:
+        return data[key]
+    except (KeyError, IndexError, TypeError):
+        return None
```
zhtmike: ai code

verl_omni/utils/diffusion_flops_counter.py
```diff
@@ -0,0 +1,674 @@
...
+            return int(shape[0])
+        except (TypeError, IndexError):
+            pass
+    # Fall back: peek at any tensor-valued entry.
+    for key in ("image_latents", "all_latents", "prompt_embeds", "prompt_embeds_mask"):
+        value = _safe_get(data, key)
+        value_shape = getattr(value, "shape", None)
+        if value_shape is not None and len(value_shape) >= 1:
+            return int(value_shape[0])
+    return 0
```
zhtmike: ai code

verl_omni/utils/diffusion_flops_counter.py
```diff
@@ -0,0 +1,674 @@
...
+    "register_diffusion_architecture",
+    "get_forward_passes_per_step",
+    "resolve_device_peak_tflops",
+]
+
+
+_DEVICE_PEAK_OVERRIDE_ENV = "VERL_OMNI_DEVICE_FLOPS_TFLOPS"
+
+
+def resolve_device_peak_tflops() -> float:
```
zhtmike: seems like a patch

verl_omni/utils/diffusion_flops_counter.py
```diff
@@ -0,0 +1,674 @@
...
+          concatenated along the sequence dimension (in which case, the count must
+          include both target and reference tokens).
+        - Latents are already flattened to sequence dimensions `(B, L, C)`.
+        """
+        if not isinstance(self, DiffusionModelFlops):
+            # Static call backward-compatibility path: get_latent_seqlens(data, config)
+            data_param = self
+            config_param = data
+            temp_inst = DiffusionModelFlops(config_param or {})
+            return temp_inst.get_latent_seqlens(data_param)
```
knlnguyen1802: This seem to be a bug
If any subclass calls SomeSubclass.get_latent_seqlens(data, config) as a static/unbound call, the shim instantiates the base DiffusionModelFlops and runs the base implementation — so any subclass override (the whole point of the framework: edit/img2vid/audio key handling) is bypassed without error.

SamitHuang:
> btw how to ensure the MFU is calculated correctly? ask another AI to have a review?

yes, cross-checked by opus and gemini

SamitHuang:
> > btw how to ensure the MFU is calculated correctly? ask another AI to have a review?
> 
> how about this, write a simple model, like the a small model with 1 or 2 layer of Qwen-image, use `nsys profile` to get the actual FLOPS, then we can check if MFU is correct

Good idea. I drafted a comparison test, but it requires running with sudo right.

```python
import torch
from diffusers import QwenImageTransformer2DModel
from verl_omni.utils.diffusion_flops_counter import DiffusionFlopsCounter

def main():
    # 1. Instantiate a tiny 2-layer Qwen-Image model
    print("Initializing tiny Qwen-Image model (2 layers)...")
    model = QwenImageTransformer2DModel(
        num_attention_heads=4,
        attention_head_dim=16,
        num_layers=2,
        in_channels=32,
        out_channels=8,
        patch_size=2,
        joint_attention_dim=48,
        axes_dims_rope=(4, 6, 6),
        guidance_embeds=False,
    ).to("cuda", dtype=torch.bfloat16)

    # 2. Setup a dummy batch
    batch_size = 1
    # 4*6*6 = 144 patches
    latent_seqlen = 144
    prompt_seqlen = 256

    # QwenImage expects packed latents (B, L, C)
    hidden_states = torch.randn(batch_size, latent_seqlen, 32, device="cuda", dtype=torch.bfloat16, requires_grad=True)
    encoder_hidden_states = torch.randn(batch_size, prompt_seqlen, 48, device="cuda", dtype=torch.bfloat16, requires_grad=True)
    timestep = torch.tensor([1.0], device="cuda", dtype=torch.bfloat16)

    # Need to pass img_shapes for RoPE (B, 3) where 3 is (frames, height, width)
    # The error "not enough values to unpack (expected 3, got 1)" means it's iterating over the batch dim incorrectly or expects a tuple of tuples
    img_shapes = [(4, 6, 6)]

    # 3. Warmup
    print("Warming up...")
    for _ in range(3):
        out = model(
            hidden_states=hidden_states,
            encoder_hidden_states=encoder_hidden_states,
            timestep=timestep,
            img_shapes=img_shapes
        )
        loss = out.sample.sum()
        loss.backward()
    # Clear grads
    model.zero_grad(set_to_none=True)
    torch.cuda.synchronize()

    # 4. Profile one step (forward + backward)
    print("Running profiled step...")
    torch.cuda.profiler.start()
    out = model(
        hidden_states=hidden_states,
        encoder_hidden_states=encoder_hidden_states,
        timestep=timestep,
        img_shapes=img_shapes
    )
    loss = out.sample.sum()
    loss.backward()
    torch.cuda.synchronize()
    torch.cuda.profiler.stop()

    # 5. Calculate theoretical FLOPs using our counter
    counter = DiffusionFlopsCounter("QwenImagePipeline", dict(model.config))
    # estimate_flops returns TFLOPS (divided by delta_time). We pass delta_time=1.0 to get raw TFLOPs, then multiply by 1e12
    theoretical_tflops, _ = counter.estimate_flops(
        latent_seqlens=[latent_seqlen] * batch_size,
        prompt_seqlens=[prompt_seqlen] * batch_size,
        delta_time=1.0,
        num_timesteps=1,
        num_forward_passes=1
    )
    theoretical_flops = theoretical_tflops * 1e12

    print("-" * 50)
    print(f"Theoretical FLOPs (fwd+bwd): {theoretical_flops:,.0f}")
    print("-" * 50)
    print("To get exact hardware FLOPs, run this script with Nsight Compute:")
    print("ncu --metrics smsp__inst_executed_pipe_tensor.sum,smsp__inst_executed_pipe_fp16.sum python3 verify_flops.py")

if __name__ == "__main__":
    main()
```

Can you help check in your env? @knlnguyen1802 @zhtmike

zhtmike:
> yes, cross-checked by opus and gemini

> Good idea. I drafted a comparison test, but it requires running with sudo right.
> Can you help check in your env? @knlnguyen1802 @zhtmike

Thanks. The codes look more clear and human-readable now. 

I think we cannot trust AI with 100% confidence, a physical evidence is still especially important for benchmarking. Let us do this before moving to the next step. @knlnguyen1802 do you have time?

verl_omni/workers/engine_workers.py
zhtmike: @SamitHuang

zhtmike:
# Code Review: `mfu` branch — Diffusion MFU calculation

**Scope reviewed:** 8 files, ~2,578 insertions (vs `main` merge-base `bab5a057`). Core focus:
- `verl_omni/utils/diffusion_flops_counter.py` (new)
- `verl_omni/workers/engine_workers.py` (integration)
- `verl_omni/trainer/diffusion/ray_diffusion_trainer.py` (metric wiring)
- `tests/utils/test_diffusion_flops_counter_on_cpu.py` (new)

Reference baseline for "verl pattern": `verl/utils/flops_counter.py`.

---

## 1. Is it correct?

**No — there is a correctness blocker that breaks the diffusion MFU path at runtime.**

The worker builds the metadata dict with the key `cfg_passes`:

```python
# engine_workers.py — _collect_diffusion_flops_meta / _allgather_diffusion_flops_meta
return {
    "latent_seqlens": ...,
    "prompt_seqlens": ...,
    "num_timesteps": ...,
    "cfg_passes": cfg_passes,
}
```

then unpacks it into the counter:

```python
estimated_flops, promised_flops = self.flops_counter.estimate_flops(
    delta_time=delta_time, **global_meta
)
```

But `DiffusionFlopsCounter.estimate_flops(...)` has **no** `cfg_passes` parameter and **no** `**kwargs`:

```python
def estimate_flops(self, latent_seqlens, prompt_seqlens, delta_time,
                   *, num_timesteps=1, num_forward_passes=1) -> tuple[float, float]:
```

So `**global_meta` injects `cfg_passes=...` and raises
`TypeError: estimate_flops() got an unexpected keyword argument 'cfg_passes'`
the first time MFU is computed on a real diffusion step. Even if it did not raise,
`num_forward_passes` would silently default to `1`, so True-CFG (2 passes) would be
undercounted by 2×.

**This is masked by the tests.** `TestDPGlobalConsistency` claims to "mirror the wiring
invariant in `TrainingWorker._postprocess_output`," but it calls
`estimate_flops(..., num_forward_passes=1)` directly instead of reconstructing the
`cfg_passes` dict the worker actually passes. The test that exists to pin this wiring is
exactly the one that hides the defect.

**Fix:** rename the dict key to `num_forward_passes` in both `_collect_diffusion_flops_meta`
and `_allgather_diffusion_flops_meta`, or have the test build/unpack the dict the same way
the worker does.

### The FLOPs math itself is sound
- Factor `6` for dense (fwd+bwd) / `12` for full bidirectional attention matches verl's
  `_estimate_qwen3_vit_flop`.
- `forward_only → /= 3.0` correctly reduces 6→2 and 12→4.
- Units are consistent (TFLOPS / TFLOPS).
- Dual-stream (latent/prompt) attribution for Qwen-Image and the packed-latent seqlen
  override (`shape[-1] == in_channels → L`) are reasonable.

---

## 2. Does it strictly follow verl's pattern?

**Partially.**

verl's `verl/utils/flops_counter.py` uses:
- module-level `_estimate_*_flops(config, tokens_sum, batch_seqlens, delta_time)` functions,
- an `ESTIMATE_FUNC` dict keyed by `model_type`,
- a single `FlopsCounter.estimate_flops(...) -> (estimated, promised)`,
- `inspect.signature` used **once**, only to detect a `**kwargs` variant.

This branch instead introduces a class-registry/decorator architecture
(`DiffusionModelFlops` + `@register_diffusion_architecture`). That is a defensible
extension for diffusion, and `DiffusionFlopsCounter` mirroring `FlopsCounter`'s return
contract `(estimated, promised)` is good.

However, the runtime `inspect.signature` dispatch inside `collect_meta` and
`estimate_flops` (branching on whether the first param is `"config"` — the "Legacy
static/class method signature" path) is explicitly **not** verl's style and is not
warranted here (see §3).

---

## 3. Are all changes necessary?

**No — there is speculative generality in brand-new code.**

- **Dead "backward-compatibility" shims.**
  `get_latent_seqlens` / `get_prompt_seqlens` each contain a
  `if not isinstance(self, DiffusionModelFlops): # Static call backward-compatibility path`
  branch, and `DiffusionFlopsCounter.estimate_flops` / `collect_meta` use
  `inspect.signature` to support a "legacy static/class method signature." This code is
  introduced in **this same branch** — there is no legacy caller to be compatible with.
  All real subclasses use the instance signature. These branches are unreachable and
  should be deleted, along with the per-call `import inspect`.

- **`resolve_device_peak_tflops` env override (`VERL_OMNI_DEVICE_FLOPS_TFLOPS`)** —
  **justified.** Relabeled H200/L20X SKUs report a device name that mis-matches verl's
  `_DEVICE_FLOPS` table and produce MFU > 1. Keep it.

- **`record_loss` placeholder-loss suppression** — reasonable and necessary (avoids a
  bogus `actor_infer/loss` / `ref/loss` curve in wandb from the constant `loss=1.0`
  returned on `compute_loss=False` paths). Though strictly it is a separate concern
  bundled into the MFU change.

### Minor / off-pattern
- `_allgather_diffusion_flops_meta` uses `all_gather_object` (pickle) rather than the
  tensor all-reduce verl uses for token counts — fine for small lists, slightly off-pattern.
- Dividing MFU by the global `get_world_size()` while gathering seqlens only over the DP
  group matches the existing LLM `elif` branch, so it is consistent.

---

## Recommended actions

1. **Fix the `cfg_passes` → `num_forward_passes` key mismatch** (the one correctness blocker).
2. **Make `TestDPGlobalConsistency` actually construct and unpack the worker's meta dict**
   so the wiring is genuinely covered.
3. **Remove the unreachable static-call / `inspect.signature` legacy branches** and the
   in-method `import inspect`.
4. Consider whether the class-registry abstraction needs to be this large for the single
   shipped architecture (`QwenImageFlops`), per verl's "no low-value / over-built" guidance.

zhtmike:
Agree with AI comment, it is over designed, drop registry please

zhtmike:
Drop resolve_device_peak_tflops as AI mentioned

SamitHuang:
> Agree with AI comment, it is over designed, drop registry please

the class-based registry is not over-built. It's necessary consider the sequence lengths and attention strategy in diffusion are highly architecture-dependent unlike LLM.
  • Some models use unpacked 4D/5D latents (B, C, H, W).
  • Some (like Qwen-Image) use packed 3D latents (B, L, C').
  • Some concatenate reference images along the sequence dimension (Edit/Inpaint models).

If we used verl's function-dictionary pattern, we would need two separate registries (SEQLEN_EXTRACTOR_FUNC and ESTIMATE_FUNC) that could easily fall out of sync.

SamitHuang:
> Drop resolve_device_peak_tflops as AI mentioned

It's for some edge cases where the hardware type is mislabeled and we have to set the peak flops from ENV. actually, the ai-gen comments suggest "keep it" in the end

zhtmike:
> > Agree with AI comment, it is over designed, drop registry please
> 
> 
> 
> the class-based registry is not over-built. It's necessary consider the sequence lengths and attention strategy in diffusion are highly architecture-dependent unlike LLM.
> 
>   • Some models use unpacked 4D/5D latents (B, C, H, W).
> 
>   • Some (like Qwen-Image) use packed 3D latents (B, L, C').
> 
>   • Some concatenate reference images along the sequence dimension (Edit/Inpaint models).
> 
> 
> 
> If we used verl's function-dictionary pattern, we would need two separate registries (SEQLEN_EXTRACTOR_FUNC and ESTIMATE_FUNC) that could easily fall out of sync.

We can refactor it later if this happens. Now seems unnecessary.

zhtmike:
> > Drop resolve_device_peak_tflops as AI mentioned
> 
> 
> 
> It's for some edge cases where the hardware type is mislabeled and we have to set the peak flops from ENV. actually, the ai-gen comments suggest "keep it" in the end

Change it to a clear name then.

examples/flowgrpo_trainer/run_qwen_image_ocr_h200_mfu_optimized.sh
zhtmike, line 1: what is the difference between current script, if there is no difference, pls drop.

verl_omni/trainer/diffusion/ray_diffusion_trainer.py
```diff
@@ -862,9 +862,11 @@ def _compute_ref_log_prob(self, batch: DataProto) -> DataProto:
         ref_log_prob = tu.get_tensordict(
             {"ref_log_prob": log_probs.float(), "ref_prev_sample_mean": prev_sample_mean.float()}
         )
-        return DataProto.from_tensordict(ref_log_prob)
+        # Propagate per-call MFU / loss metrics computed in the worker.
+        infer_metrics = tu.get(output, "metrics", default={})
+        return DataProto.from_tensordict(ref_log_prob), infer_metrics
```
zhtmike: we can add metrics into ref_log_prob tensordict directly, not recommand to change return value

verl_omni/trainer/diffusion/ray_diffusion_trainer.py
```diff
@@ -881,7 +883,8 @@ def _compute_old_log_prob(self, batch: DataProto):
         if prev_sample_mean is not None:
             old_log_prob_dict["old_prev_sample_mean"] = prev_sample_mean.float()
         old_log_prob = tu.get_tensordict(old_log_prob_dict)
-        return DataProto.from_tensordict(old_log_prob)
+        infer_metrics = tu.get(output, "metrics", default={})
+        return DataProto.from_tensordict(old_log_prob), infer_metrics
```
zhtmike: same

verl_omni/trainer/diffusion/ray_diffusion_trainer.py
```diff
@@ -1001,7 +1004,8 @@ def fit(self):
                         apply_bypass_mode_to_diffusion_batch(batch)
                     else:  # Recompute old_log_probs
                         with marked_timer("old_log_prob", timing_raw, color="blue"):
-                            old_log_prob = self._compute_old_log_prob(batch)
+                            old_log_prob, infer_metrics = self._compute_old_log_prob(batch)
+                            metrics.update(rename_dict(reduce_metrics(infer_metrics), "actor_infer/"))
```
zhtmike: rename_dict is for what? 
and reduce_metric is also for what
make the name clear.

verl_omni/trainer/diffusion/ray_diffusion_trainer.py
```diff
@@ -1018,7 +1022,8 @@ def fit(self):
                     if self.use_reference_policy:
                         # compute reference log_prob
                         with marked_timer(str(Role.RefPolicy), timing_raw, color="olive"):
-                            ref_log_prob = self._compute_ref_log_prob(batch)
+                            ref_log_prob, infer_metrics = self._compute_ref_log_prob(batch)
+                            metrics.update(rename_dict(reduce_metrics(infer_metrics), "ref/"))
```
zhtmike: same

verl_omni/trainer/diffusion/ray_diffusion_trainer.py
```diff
@@ -1220,7 +1225,7 @@ def _update_actor(self, batch: DataProto) -> DataProto:
         actor_output = rename_dict(actor_output, "actor/")
         return DataProto.from_single_dict(data={}, meta_info={"metrics": actor_output})
 
-    def _compute_ref_noise_pred(self, batch: DataProto) -> Optional[DataProto]:
+    def _compute_ref_noise_pred(self, batch: DataProto) -> Optional[tuple[DataProto, dict[str, Any]]]:
```
zhtmike: dont change return, as above

verl_omni/trainer/diffusion/ray_diffusion_trainer.py
```diff
@@ -1254,7 +1259,8 @@ def _compute_ref_noise_pred(self, batch: DataProto) -> Optional[DataProto]:
             "noise": noise.float(),
             "timesteps": timesteps.float(),
         }
-        return DataProto.from_tensordict(tu.get_tensordict(ref_output))
+        infer_metrics = tu.get(output, "metrics", default={})
+        return DataProto.from_tensordict(tu.get_tensordict(ref_output)), infer_metrics
```
zhtmike: same

verl_omni/trainer/diffusion/ray_diffusion_trainer.py
```diff
@@ -1351,6 +1357,8 @@ def fit(self):
                         with marked_timer(str(Role.RefPolicy), timing_raw, color="olive"):
                             ref_dpo = self._compute_ref_noise_pred(batch)
                             if ref_dpo is not None:
+                                ref_dpo, infer_metrics = ref_dpo
+                                metrics.update(rename_dict(reduce_metrics(infer_metrics), "ref/"))
```
zhtmike: same

verl_omni/utils/diffusion_flops_counter.py
zhtmike: make a better name

verl_omni/utils/diffusion_flops_counter.py
```diff
@@ -0,0 +1,567 @@
...
+    Honors the ``VERL_OMNI_DEVICE_FLOPS_TFLOPS`` env var as a manual
+    override and otherwise falls back to upstream
+    :func:`verl.utils.flops_counter.get_device_flops`.
+
+    The override is needed on clusters where ``torch.cuda.get_device_name()``
+    returns a string verl's built-in table doesn't recognize
+    (e.g. relabeled H200 SKUs that report as ``"NVIDIA L20X"`` and would
+    otherwise mis-match the ``"L20"`` entry, producing nonsensical
+    MFU > 1.0).
+    """
```
zhtmike: let us fix it on verl directly, and drop the patch

verl_omni/workers/engine_workers.py
```diff
@@ -62,6 +66,35 @@
...
+            config_path,
+        )
+        return None
+    try:
+        with open(config_path) as f:
+            return json.load(f)
+    except (OSError, json.JSONDecodeError) as exc:
+        logger.warning("Diffusion MFU disabled: failed to read %s: %s", config_path, exc)
+        return None
+
```
zhtmike: read the config under `verl_omni/workers/config/diffusion/model.py`, `DiffusionModelConfig.__post_init__` directly, we should not patch it here.

verl_omni/workers/engine_workers.py
```diff
@@ -151,8 +184,16 @@ def __init__(self, config: TrainingWorkerConfig):
 
         if hasattr(self.model_config, "hf_config"):
             self.flops_counter = FlopsCounter(self.model_config.hf_config)
+        elif self.config.model_type in ("diffusion_model", "diffusion_dp_model"):
+            # Diffusion configs (diffusers ConfigMixin) have a different schema
+            # than HF LLM configs, so we use a dedicated counter that understands
+            # joint-attention DiTs and multi-step denoising.
```
zhtmike: clean

verl_omni/workers/engine_workers.py
```diff
@@ -179,11 +220,32 @@ def reset(self):
...
+                with keys ``latent_seqlens``, ``prompt_seqlens``, ``num_timesteps``,
+                ``cfg_passes``. Provided only on the diffusion path; ignored
+                otherwise.
+            record_loss: when ``False``, drop ``output["loss"]`` without
+                aggregating or emitting it as a metric. Set this for infer
+                paths that pass ``compute_loss=False`` (e.g. old/ref
+                log-prob recompute), where the engine returns a constant
+                placeholder ``loss=1.0`` per micro-batch that would
+                otherwise surface in wandb as a meaningless
+                ``actor_infer/loss`` / ``ref/loss`` curve.
```
zhtmike: make it short, 2 lines for each

verl_omni/workers/engine_workers.py
```diff
@@ -179,11 +220,32 @@ def reset(self):
...
+    def _postprocess_output(
+        self,
+        output,
+        *,
+        global_token_num,
+        delta_time,
+        forward_only,
+        images_seqlens,
+        diffusion_flops_meta: Optional[dict] = None,
+        record_loss: bool = True,
```
zhtmike: this is a patch fix, which should be irrevant to mfu pr.
And it should caused by recent commits, drop this and we need to find the root cause first. we can do it on a seperate PR later

verl_omni/workers/engine_workers.py
```diff
@@ -194,14 +256,20 @@ def _postprocess_output(self, output, *, global_token_num, delta_time, forward_o
...
-            torch.distributed.all_reduce(loss, op=torch.distributed.ReduceOp.AVG, group=dp_group)
-        loss = loss.item()
+        loss = None
+        if record_loss:
+            loss = torch.sum(torch.tensor(output.pop("loss"), device=self.device_name))
+            if dp_group is not None:
+                torch.distributed.all_reduce(loss, op=torch.distributed.ReduceOp.AVG, group=dp_group)
+            loss = loss.item()
+        else:
+            # Drain the placeholder loss so it cannot leak into downstream handling.
```
zhtmike: revert change as mentioned above

verl_omni/workers/engine_workers.py
```diff
@@ -212,7 +280,8 @@ def _postprocess_output(self, output, *, global_token_num, delta_time, forward_o
             final_metrics = allgather_dict_into_dict(data=metrics, group=dp_group)
         else:
             final_metrics = metrics
-        final_metrics["loss"] = loss
+        if record_loss:
+            final_metrics["loss"] = loss
```
zhtmike: same

verl_omni/workers/engine_workers.py
```diff
@@ -423,12 +507,87 @@ def infer_batch(self, data: TensorDict) -> TensorDict:
                 delta_time=delta_time,
                 forward_only=True,
                 images_seqlens=images_seqlens,
+                diffusion_flops_meta=diffusion_flops_meta,
+                # When the caller passed compute_loss=False (old/ref log-prob
+                # recompute on the diffusion path), the engine returns a
+                # constant placeholder loss=1.0 per micro-batch. Suppress
+                # emission so it doesn't surface as a meaningless metric.
+                record_loss=compute_loss,
```
zhtmike: drop this for this pr

verl_omni/workers/engine_workers.py
```diff
@@ -423,12 +507,87 @@ def infer_batch(self, data: TensorDict) -> TensorDict:
...
+
+        Returns ``None`` for non-diffusion engines so the LLM code path keeps
+        flowing through ``self.flops_counter.estimate_flops(global_token_num, ...)``.
+
+        Per-rank ``latent_seqlens`` / ``prompt_seqlens`` come from the
+        architecture class registered for ``DiffusionModelConfig.architecture``
+        (see :class:`DiffusionModelFlops`). ``num_timesteps`` is read
+        from the FlowGRPO/MixGRPO ``all_timesteps`` trajectory (defaulting
+        to 1 for DPO / one-shot paths) and ``cfg_passes`` is resolved from
+        the pipeline + transformer configs via :func:`get_forward_passes_per_step`.
```
zhtmike: apparently it is only for flowgrpo, consider move it to a specific position only. cc @wtomin 

Or make it generic enough.

verl_omni/workers/engine_workers.py
```diff
@@ -423,12 +507,87 @@ def infer_batch(self, data: TensorDict) -> TensorDict:
...
+        cfg_passes = get_forward_passes_per_step(pcfg_view, transformer_config)
+
+        return {
+            "latent_seqlens": seqlens["latent_seqlens"],
+            "prompt_seqlens": seqlens["prompt_seqlens"],
+            "num_timesteps": num_timesteps,
+            "num_forward_passes": cfg_passes,
+        }
+
+    def _allgather_diffusion_flops_meta(self, meta: dict) -> dict:
```
zhtmike: same, need refactor cc @wtomin

zhtmike (commented):
reviewed with comments

zhtmike:
reviewed. The mfu seems for qwen-image specific, but it added too much model-specific methods under public class and APIs. it will break our current refactoring for different algorithm implementation.

@wtomin @AndyZhou952

examples/flowgrpo_trainer/run_qwen_image_ocr_h200_mfu_optimized.sh
SamitHuang: mfu optimized by h200, by setting SP=1 to reduce all-to-all overhead, and layered_summon=False.

verl_omni/trainer/diffusion/ray_diffusion_trainer.py
SamitHuang: it's aligned with verl's design, where _compute_old_log_prob also return the old_log_prob and mfu related metrics.

https://github.com/verl-project/verl/blob/0c36dee1387243ee0c1c57273dd81cd7902422b6/verl/trainer/ppo/ray_trainer.py#L1290-L1291

verl_omni/trainer/diffusion/ray_diffusion_trainer.py
SamitHuang: these two utils functions are from verl, better not to change the function names

  1. reduce_metrics turns {"mfu": [0.31, 0.32, 0.33, 0.31]} into {"mfu": 0.3175} (the mean).
  2. rename_dict turns {"mfu": 0.3175} into {"actor_infer/mfu": 0.3175}.

verl_omni/trainer/diffusion/ray_diffusion_trainer.py
SamitHuang: similar to _compute_old_log_prob, for API design consistency

examples/flowgrpo_trainer/run_qwen_image_ocr_h200_mfu_optimized.sh
zhtmike: change it to a better name, like h200_mfu_optmized, etc

verl_omni/trainer/diffusion/ray_diffusion_trainer.py
zhtmike: only `_compute_old_log_prob` has tuple return, others are not in the verl. Why we have this difference?

verl_omni/trainer/diffusion/ray_diffusion_trainer.py
zhtmike: in verl, it is only used in `fit`, why we used it everywhere?

verl_omni/trainer/diffusion/ray_diffusion_trainer.py
zhtmike: questioned above

examples/flowgrpo_trainer/run_qwen_image_ocr_h200_mfu_optimized.sh
zhtmike: And if I understand correctly, we should use async reward for optimal MFU, but seems this script is not enabled?

verl_omni/trainer/diffusion/ray_diffusion_trainer.py
SamitHuang: _compute_ref_log_prob and _compute_ref_noise_pred are refactored to return one item now

verl_omni/workers/engine_workers.py
```diff
@@ -151,8 +155,12 @@ def __init__(self, config: TrainingWorkerConfig):
 
         if hasattr(self.model_config, "hf_config"):
             self.flops_counter = FlopsCounter(self.model_config.hf_config)
+        elif self.config.model_type in ("diffusion_model", "diffusion_dp_model"):
```
zhtmike: it should be `diffusion_dpo_model`, it is a typo and just fixed in main branch

verl_omni/workers/engine_workers.py
```diff
@@ -151,8 +155,12 @@ def __init__(self, config: TrainingWorkerConfig):
...
+            self.flops_counter = DiffusionFlopsCounter(
+                architecture=getattr(self.model_config, "architecture", None),
+                transformer_config=getattr(self.model_config, "transformer_config", None),
```
zhtmike: let us make it clear. just

architecture=self.model_config.architecture
transformer_config=self.model_config.transformer_config

verl_omni/trainer/diffusion/ray_diffusion_trainer.py
SamitHuang: our mfu metric naming format is different from verl’s actor MFU renaming. it's changed for alignement now.

SamitHuang:
> We can refactor it later if this happens. Now seems unnecessary.

The current design has been considered for extending supoport for other diffusion models and ease-of-use. We should make a good design choice at the beginning.

A. class register (current)
One class per architecture. Seqlen hooks + FLOPs formula live together; reuse via subclassing.

```python
@register_diffusion_architecture(
    "QwenImagePipeline",
    "QwenImagePipelineWithLogProb",
)
class QwenImageFlops(DiffusionModelFlops):
    def get_latent_seqlens(self, data) -> list[int]:
        # packed (B, [T,] L, C') vs unpacked (B, [T,] C, H, W)
        ...

    def estimate_flops(self, latent_seqlens, prompt_seqlens, delta_time, *,
                       num_timesteps, num_forward_passes) -> float:
        ...


@register_diffusion_architecture("QwenImageEditPipeline")
class QwenImageEditFlops(QwenImageFlops):
    def get_latent_seqlens(self, data) -> list[int]:
        base = super().get_latent_seqlens(data)
        ref = data.get("reference_image_latents")
        if ref is None:
            return base
        return [b + int(ref.shape[-2]) for b in base]


@register_diffusion_architecture("WanPipeline", "WanPipelineWithLogProb")
class WanFlops(DiffusionModelFlops):
    # default get_latent_seqlens / get_prompt_seqlens often enough for (B,C,T,H,W)
    def estimate_flops(self, latent_seqlens, prompt_seqlens, delta_time, *,
                       num_timesteps, num_forward_passes) -> float:
        ...
```

B. function register

Like `verl/utils/flops_counter.py` `ESTIMATE_FUNC`, plus separate seqlen dicts (diffusion has two streams; LLM has one).

```python
LATENT_SEQLEN_FUNC = {
    "QwenImagePipeline": _get_qwenimage_latent_seqlens,
    "QwenImageEditPipeline": _get_qwenimageedit_latent_seqlens,
    "WanPipeline": _get_wan_latent_seqlens,
}

PROMPT_SEQLEN_FUNC = {
    # default for most; override per arch
}

ESTIMATE_FUNC = {
    "QwenImagePipeline": _estimate_qwenimage_flops,
    "QwenImageEditPipeline": _estimate_qwenimage_flops,
    "WanPipeline": _estimate_wan_flops,
}


def _get_qwenimageedit_latent_seqlens(data, config):
    base = _get_qwenimage_latent_seqlens(data, config)
    ref = data.get("reference_image_latents")
    ...

def _estimate_qwenimage_flops(config, latent_seqlens, prompt_seqlens, delta_time, *,
                              num_timesteps, num_forward_passes):
    ...


class DiffusionFlopsCounter:
    def collect_meta(self, data):
        get_latent = LATENT_SEQLEN_FUNC.get(self.architecture, _default_latent)
        get_prompt = PROMPT_SEQLEN_FUNC.get(self.architecture, _default_prompt)
        return {"latent_seqlens": get_latent(data, self.config),
                "prompt_seqlens": get_prompt(data, self.config)}

    def estimate_flops(self, ...):
        func = ESTIMATE_FUNC.get(self.architecture, _unknown)
        return func(self.config, ...)
```

I picked because: 
- (A) can handle mode-specific seqlens, FLOPs compute in one class, and easy to inherit e.g. (Edit ⊂ T2I), less code redundancy, 
- (A) avoid setting 2-3 dict for each model, i.e. LATENT_SEQLEN_FUNC, PROMPT_SEQLEN_FUNC, ESTIMATE_FUNC in (B), and (B) is hard to re-use the code of seqlen or flops estimation between models of the same family.

Let's discuss later which one is better for diffusion Flops compute

zhtmike:
OK, let us stick with option A. But it should be user-friendly and easy-to-use:

Consider this:
- Please make a mode base for mfu only, similar to `verl_omni/pipelines/model_base.py::DiffusionModelBase`, may put it under `verl_omni/pipelines/mfu`
- move QwenImage MFU registration under `verl_omni/pipelines/mfu/qwen_image`
- add a doc of how to add new model for MFU under doc/contributing, the detail maths can be mentioned in docs/diffusion_mfu.md as current state

We should decouple normal model interface with mfu model interface to avoid ambiguous.

examples/flowgrpo_trainer/run_qwen_image_ocr_h200_mfu_optimized.sh
zhtmike: And also bypass mode?

SamitHuang: perf/mfu/actor / perf/mfu/actor_infer measure DiT GPU utilization during train_batch / infer_batch only
the additional 1 GPU for async reward doesn't change the delta time for the rollout phase (same as synchronous training)

zhtmike: then the change of `layered_summon=False` seems also has no effect?

zhtmike: And we already set `SP=1` by default

SamitHuang:
> OK, let us stick with option A. But it should be user-friendly and easy-to-use:
> 
> Consider this:
> 
> * Please make a mode base for mfu only, similar to `verl_omni/pipelines/model_base.py::DiffusionModelBase`, may put it under `verl_omni/pipelines/mfu`
> * move QwenImage MFU registration under `verl_omni/pipelines/mfu/qwen_image`
> 
> We should decouple normal model interface with mfu model interface to avoid ambiguous.

1. yes, we have DiffusionModelFlops as a base class for model FLOPs and MFU estimators
2. put into verl_omni/utils/mfu
```
  verl_omni/utils/mfu/
  ├── __init__.py                  # public API + registers built-in archs
  ├── diffusion_flops_counter.py   # registry, base class, helpers
 └── qwen_image.py                # QwenImageFlops
```

zhtmike:
> > OK, let us stick with option A. But it should be user-friendly and easy-to-use:
> > Consider this:
> > 
> > * Please make a mode base for mfu only, similar to `verl_omni/pipelines/model_base.py::DiffusionModelBase`, may put it under `verl_omni/pipelines/mfu`
> > * move QwenImage MFU registration under `verl_omni/pipelines/mfu/qwen_image`
> > 
> > We should decouple normal model interface with mfu model interface to avoid ambiguous.
> 
> 1. yes, we have DiffusionModelFlops as a base class for model FLOPs and MFU estimators
> 2. put into verl_omni/utils/mfu
> 
> ```
>   verl_omni/utils/mfu/
>   ├── __init__.py                  # public API + registers built-in archs
>   ├── diffusion_flops_counter.py   # registry, base class, helpers
>  └── qwen_image.py                # QwenImageFlops
> ```

OK, please address the remaining reviews~

examples/flowgrpo_trainer/run_qwen_image_ocr_h200_mfu_optimized.sh
SamitHuang: https://github.com/verl-project/verl-omni/blob/main/examples/flowgrpo_trainer/run_qwen_image_ocr.sh#L15

SP=2 for full-param training

tests/utils/test_diffusion_flops_counter_on_cpu.py
```diff
@@ -0,0 +1,767 @@
+# Copyright 2026 Bytedance Ltd. and/or its affiliates
+#
```
knlnguyen1802, line 2: Replace this by a test similar to this comment https://github.com/verl-project/verl-omni/pull/128#issuecomment-4594731203 is better

verl_omni/workers/engine_workers.py
SamitHuang: verl's convention: 
- do batch meta data process in engine_worker inline in train_mini_batch, we create _collect_diffusion_flops_meta()() because diffusion needs more metadata than input_ids token counts. it is cleaner.
- DP gather inline all_gather_object, we create _allgather_diffusion_flops_meta() for the same reason
- MFU emit is in _postprocess_output, same as verl-omni.

https://github.com/verl-project/verl/blob/cb8211095e7911b07c3fa5bd4e84c1cff4211d43/verl/workers/engine_workers.py#L219-L225

examples/flowgrpo_trainer/run_qwen_image_ocr_h200_mfu_optimized.sh
zhtmike: ok

verl_omni/utils/mfu/diffusion_flops_counter.py
```diff
@@ -0,0 +1,290 @@
...
+                "verl.utils.flops_counter.get_device_flops().",
+                stacklevel=2,
+            )
+        except ValueError:
+            warnings.warn(
+                f"{_DEVICE_PEAK_OVERRIDE_ENV}={raw!r} is not a valid float; falling back to "
+                "verl.utils.flops_counter.get_device_flops().",
+                stacklevel=2,
+            )
+    return float(get_device_flops("T"))
```
zhtmike, line 72: why need this?

verl_omni/workers/engine_workers.py
zhtmike: move them to a helper function outside of this class, possibly under utils/mfu/helper.py, state clearly it is currently for flowgrpo/mixgrpo only with a TODO statement say it needs refactoring

verl_omni/utils/mfu/diffusion_flops_counter.py
SamitHuang: in some cases the device type are mislabeled. e.g. H200 is labeled as L20X

zhtmike: can we fix it on verl directly? 

and leave a # TODO here to drop it later

SamitHuang: sure, let's add a TODO for now

zhtmike:
Looks good, let me ask AI for a final review

verl_omni/workers/engine_workers.py
```diff
@@ -236,7 +255,21 @@ def _postprocess_output(self, output, *, global_token_num, delta_time, forward_o
...
+                # Counter expects global (DP-allgathered) seqlens, matching the
+                # convention used for ``global_token_num`` in the LLM path.
+                global_meta = allgather_diffusion_flops_meta(
+                    diffusion_flops_meta, self.engine.get_data_parallel_group()
+                )
+                estimated_flops, promised_flops = self.flops_counter.estimate_flops(
+                    delta_time=delta_time, **global_meta
+                )
+                if promised_flops > 0:
+                    final_metrics["mfu"] = estimated_flops / promised_flops / torch.distributed.get_world_size()
```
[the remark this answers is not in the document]
zhtmike: @SamitHuang pls check this

verl_omni/utils/mfu/qwen_image.py
```diff
@@ -0,0 +1,106 @@
...
+        """Handle diffusers ``_pack_latents`` layout ``(B, [T,] L, C')``."""
+        latents, stacked = read_latents(data)
+        if latents is not None and hasattr(latents, "shape"):
+            shape = tuple(int(d) for d in latents.shape)
+            in_channels = int(self.config.get("in_channels") or 0)
+            if in_channels > 0 and shape[-1] == in_channels:
+                # Packed: (B, L, C') or FlowGRPO-stacked (B, T, L, C').
+                # Unpacked NFT/VAE latents (B, C, H, W) also end in C' but
+                # have four dims without a timestep axis — defer to the base
+                # spatial product instead of treating H as the seqlen.
```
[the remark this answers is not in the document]
zhtmike: @SamitHuang pls check this

verl_omni/trainer/diffusion/ray_diffusion_trainer.py
```diff
@@ -869,7 +871,7 @@ def _compute_ref_log_prob(self, batch: DataProto) -> DataProto:
         )
         return DataProto.from_tensordict(ref_log_prob)
 
-    def _compute_old_log_prob(self, batch: DataProto):
+    def _compute_old_log_prob(self, batch: DataProto) -> tuple[DataProto, float]:
```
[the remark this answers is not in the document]
zhtmike: @SamitHuang pls check this

verl_omni/workers/engine_workers.py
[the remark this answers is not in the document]
SamitHuang: the comment sounds. will fix for sp

verl_omni/workers/engine_workers.py
```diff
@@ -236,11 +255,29 @@ def _postprocess_output(self, output, *, global_token_num, delta_time, forward_o
                 flatten_v = [sublist[0] for sublist in v]  # sublist should be single element
                 final_metrics[k] = sum(flatten_v) / len(flatten_v)
         # compute mfu
-        if global_token_num is not None and self.flops_counter is not None:
+        # Seqlens / token counts are gathered over ``dp_group``; normalize MFU by
+        # the same scope (not global world size) so Ulysses/SP ranks are not
+        # double-penalized when ``dp_group`` is smaller than WORLD.
```
zhtmike: clean pls

verl_omni/utils/mfu/diffusion_flops_counter.py
```diff
@@ -0,0 +1,369 @@
...
+    Returns ``None`` when ``flops_counter`` is not a :class:`DiffusionFlopsCounter`
+    so the LLM code path can keep using ``FlopsCounter.estimate_flops(global_token_num, ...)``.
+
+    Per-rank ``latent_seqlens`` / ``prompt_seqlens`` come from the architecture
+    class registered for ``DiffusionModelConfig.architecture`` (see
+    :class:`DiffusionModelFlops`). ``num_timesteps`` is read from
+    ``all_timesteps`` (FlowGRPO / MixGRPO), ``train_timesteps`` (DiffusionNFT),
+    or defaults to ``1`` for one-shot paths such as diffusion DPO.
+    ``num_forward_passes`` is resolved from the pipeline + transformer configs
+    via :func:`get_forward_passes_per_step`.
```
zhtmike: make it human readable, like 2 lines

zhtmike:
perfect, no ai comments

zhtmike (approved):
LGTM

PR #141, opened 2026-06-04
[trainer] feat: add fa3 backend support for diffusers engine

zhtmike:
**###** What does this PR do?
- Add fa3 backend support for diffusers engine
- Requires `kernel == v0.14.1` 
- add a patch to fix the numerical precision of fa3 backend

Result:
螢幕截圖 2026-06-09 上午10 04 15
螢幕截圖 2026-06-09 上午10 03 58
螢幕截圖 2026-06-09 上午10 04 38

----
This note explains the bug fixed by the temporary monkey-patch in
[`verl_omni/pipelines/_patch.py`](../../verl_omni/pipelines/_patch.py)
(`_apply_flash_attention_3_varlen_hub_fix`). 

## TL;DR

In diffusers `0.38.0`, the FA3 varlen backend
(`_flash_attention_3_varlen_hub`, used by verl-omni's `_flash_3_varlen_hub`
attention backend) packs valid key/value tokens with a **contiguous-prefix**
assumption:

```python
valid_len = seqlens_k[b]
key_valid.append(key[b, :valid_len])      # <-- assumes valid tokens are the first `valid_len`
value_valid.append(value[b, :valid_len])
```

This is only correct when each sample's valid positions are a contiguous block
at the start of the sequence. For **Qwen-Image**, the joint text/image attention
mask is **not contiguous**, so `key[b, :valid_len]` selects the *wrong* tokens —
while `cu_seqlens_k` (derived from `attn_mask.sum(dim=1)`) is still computed
correctly. The result is a silent correctness bug: the packed lengths are right,
but the packed *contents* are wrong.

The fix selects the actual valid positions by index:

```python
indices_k = attn_mask.flatten().nonzero(as_tuple=False).flatten()
key_packed = key.reshape(-1, *key.shape[2:])[indices_k]
value_packed = value.reshape(-1, *value.shape[2:])[indices_k]
```

This matches the upstream diffusers `fa3` branch fix.

## Concrete example

Take a batch of 2 samples, KV sequence length 4, 1 head, head-dim 2.
Suppose the attention mask marks valid (`1`) / padding (`0`) positions as:

```text
attn_mask = [[1, 0, 1, 1],     # sample 0: valid positions {0, 2, 3}, NOT contiguous
             [0, 1, 1, 0]]     # sample 1: valid positions {1, 2},    NOT contiguous
```

So:

- `seqlens_k = attn_mask.sum(dim=1) = [3, 2]`
- `cu_seqlens_k = [0, 3, 5]` (5 valid tokens total) — **correct in both versions**

Let `key[b, t]` be identified by its `(sample, position)` index.

### Buggy behavior (`key[b, :valid_len]`)

```text
sample 0, valid_len=3 -> positions {0, 1, 2}   # WRONG: position 1 is padding, position 3 dropped
sample 1, valid_len=2 -> positions {0, 1}       # WRONG: position 0 is padding, position 2 dropped
```

Packed K/V tokens: `[(0,0), (0,1), (0,2), (1,0), (1,1)]`

These do not correspond to the masked-in tokens at all. Note the lengths (3, 2)
*look* right, so nothing crashes — the attention output is just computed over the
wrong keys/values.

### Patched behavior (`nonzero` index gather)

```text
indices_k = nonzero(flatten([1,0,1,1, 0,1,1,0])) = [0, 2, 3, 5, 6]
```

Packed K/V tokens: `[(0,0), (0,2), (0,3), (1,1), (1,2)]`

These are exactly the masked-in tokens, in per-sample order, consistent with
`cu_seqlens_k = [0, 3, 5]`.

### Verification snippet

```python
import torch

mask = torch.tensor([[1, 0, 1, 1], [0, 1, 1, 0]], dtype=torch.bool)
b, skv = mask.shape
key = torch.arange(b * skv * 1 * 2, dtype=torch.float32).reshape(b, skv, 1, 2)

# patched gather
idx = mask.flatten().nonzero(as_tuple=False).flatten()
packed = key.reshape(-1, *key.shape[2:])[idx]

# reference: per-sample concat of masked-in rows
expected = torch.cat([key[i][mask[i]] for i in range(b)], dim=0)

assert torch.equal(packed, expected)   # passes
```

## When does this matter?

- Triggered whenever an attention mask with **non-contiguous** valid positions is
  passed to the FA3 varlen backend. Qwen-Image joint text+image attention is the
  main case in verl-omni.
- With a fully-valid mask (or `attn_mask=None`), the contiguous-prefix assumption
  happens to hold, so the bug does not appear — which is why it can go unnoticed.

## Why a patch instead of bumping diffusers?

The fix lives upstream in the diffusers `fa3` pr, but we cannot easily pin to
an unreleased diffusers. The monkey-patch in
[`verl_omni/pipelines/_patch.py`](../../verl_omni/pipelines/_patch.py):

- Only activates on diffusers `>= 0.38.0`.
- Replaces the registered `_flash_3_varlen_hub` backend with a signature-identical
  function, so the dispatcher's argument filtering is unaffected.
- Is idempotent and logs a one-time warning.

**Remove this patch once the fix is released upstream in diffusers.**

> Add **concise** overview of what this PR aims to achieve or accomplish. Reference related GitHub issues and PRs that help with the review.

### Checklist Before Starting

### Test

### API and Usage Example

```python
```

### Design & Code Changes

> [!IMPORTANT]


examples/flowgrpo_trainer/run_qwen_image_ocr_lora_fsdp2_fa3.sh
```diff
@@ -0,0 +1,76 @@
...
+set -x
+
+# Set WORKSPACE to any writable directory; defaults to $HOME
+WORKSPACE=${WORKSPACE:-$HOME}
+
+ocr_train_path=$WORKSPACE/data/ocr/qwen_image/train.parquet
+ocr_test_path=$WORKSPACE/data/ocr/qwen_image/test.parquet
+
+model_name=Qwen/Qwen-Image
+reward_model_name=Qwen/Qwen3-VL-8B-Instruct
```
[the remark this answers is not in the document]
zhtmike: 。。。

SamitHuang:
this is helpful for vllm-omni 0.22 upgrade, which has also used fa3 kernel

examples/flowgrpo_trainer/run_qwen_image_ocr_lora_fsdp2_fa3.sh
```diff
@@ -0,0 +1,74 @@
...
+    data.train_files=$ocr_train_path \
+    data.val_files=$ocr_test_path \
+    data.train_batch_size=32 \
+    data.max_prompt_length=256 \
+    actor_rollout_ref.model.algorithm=flow_grpo \
+    actor_rollout_ref.model.path=$model_name \
+    actor_rollout_ref.model.lora_rank=64 \
+    actor_rollout_ref.model.lora_alpha=128 \
+    actor_rollout_ref.model.target_modules="['to_q','to_k','to_v','to_out.0','add_q_proj','add_k_proj','add_v_proj','to_add_out','img_mlp.net.0.proj','img_mlp.net.2','txt_mlp.net.0.proj','txt_mlp.net.2']" \
+    actor_rollout_ref.model.attn_backend="_flash_3_varlen_hub" \
```
SamitHuang, line 32: better mention how to configure actor attention backend in the doc as well

verl_omni/pipelines/_patch.py
```diff
@@ -76,4 +76,97 @@ def _patched_forward(
     QwenImageTransformer2DModel.forward = _patched_forward
 
 
+def _apply_flash_attention_3_varlen_hub_fix() -> None:
```
SamitHuang, line 81: can add a TODO for future drop

SamitHuang (commented):
should we enable fa3 in smoking test?

verl_omni/pipelines/_patch.py
zhtmike: done

examples/flowgrpo_trainer/run_qwen_image_ocr_lora_fsdp2_fa3.sh
zhtmike: done, added a doc

zhtmike:
> should we enable fa3 in smoking test?

It needs to have an internet access to download the kernel from huggingface hub. Since the internet in ci server is not stable, I prefer not to do this. It should be guarded already in diffusers side.

AndyZhou952 (approved):

SamitHuang (approved):

PR #211, opened 2026-06-26
[rollout, model] fix: guard rollout_attn_backend for AR rollouts and refresh hf_processor re-export

qinganrice:
## What

1. **AR rollout crash.** `vLLMOmniHttpServer` unconditionally read `self.config.rollout_attn_backend`, a field that only exists on the *diffusion* rollout config (added in #200). AR (text) rollouts hit `AttributeError: 'RolloutConfig' object has no attribute 'rollout_attn_backend'`. It is now guarded with `getattr`, and the engine arg is only set when the field is present.

2. **`chat_template is not set`.** `patch_hf_processor_for_qwen3_omni` patched `verl.utils.tokenizer.hf_processor`, but not the snapshot re-export captured by `from verl.utils import hf_processor`. Callers on that import path got the unpatched function, so the processor came back `None`, the loader fell through to the tokenizer branch which has no chat_template, and training failed. The package-level binding is now refreshed too.

## Not a duplicate

#200 *introduced* the diffusion-only `rollout_attn_backend`; this guards the AR path it broke rather than re-implementing it. No open PR touches the `hf_processor` re-export.

## Testing

An AR GSPO+LoRA rollout that previously crashed at server launch now starts and trains end-to-end. A collaborator's `chat_template is not set` failure via `from verl.utils import hf_processor` was reproduced and confirmed fixed (the refresh returns the Qwen3-Omni processor with a non-None `chat_template`).

### DIFF
--- a/verl_omni/models/transformers/qwen3_omni_thinker.py
+++ b/verl_omni/models/transformers/qwen3_omni_thinker.py
@@ -165,6 +165,12 @@ def patch_hf_processor_for_qwen3_omni() -> None:
             return None
 
     _vt.hf_processor = _patched_hf_processor
+    # Also refresh verl.utils's stale re-export (callers use `from verl.utils import hf_processor`).
+    import sys as _sys
+
+    _utils_mod = _sys.modules.get("verl.utils")
+    if _utils_mod is not None and hasattr(_utils_mod, "hf_processor"):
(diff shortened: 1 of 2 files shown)


knlnguyen1802 (approved):
LGTM, thanks for the fix

SamitHuang (approved):
LGTM

PR #227, opened 2026-07-02
[recipe, diffusion] fix: let Ray set Ascend visible devices in Qwen-Image NPU script

Sky-Trigger:
## What does this PR do?

This PR removes `RAY_EXPERIMENTAL_NOSET_ASCEND_RT_VISIBLE_DEVICES=1` from the Qwen-Image OCR LoRA NPU FlowGRPO example script:

```bash
examples/flowgrpo_trainer/qwen_image/run_qwen_image_ocr_lora_npu.sh
```

With this environment variable set, Ray skips rewriting `ASCEND_RT_VISIBLE_DEVICES` for Ascend/NPU workers. In this FlowGRPO Qwen-Image NPU recipe, this can lead to device placement mismatch across workers.

Observed error:

```text
AssertionError: Expects tensor to be on the compute device npu:15, was on npu:0
```

Removing this override allows Ray to manage the per-worker Ascend device visibility normally, which fixes the device mismatch in the NPU FlowGRPO training script.

## Why is this needed?

`RAY_EXPERIMENTAL_NOSET_ASCEND_RT_VISIBLE_DEVICES=1` is useful in some vLLM-Ascend distributed serving setups where device mapping is handled outside Ray. However, for this verl-omni FlowGRPO training recipe, keeping it enabled prevents Ray from setting the expected per-worker device visibility and may cause tensors to be created on the wrong NPU.

This change keeps the example script aligned with Ray-managed NPU placement and avoids the runtime device assertion failure.

## Test

Tested by running the Qwen-Image OCR LoRA NPU FlowGRPO example.

Before this change, the script failed with:

```text
AssertionError: Expects tensor to be on the compute device npu:15, was on npu:0
```

After removing `RAY_EXPERIMENTAL_NOSET_ASCEND_RT_VISIBLE_DEVICES=1`, the script can proceed without this device mismatch error.


SamitHuang (approved):
LGTM

PR #224, opened 2026-07-02
[omni,model] fix: load Qwen3OmniMoeProcessor in AgentLoopWorker via stale hf_processor refresh

wtomin:
### What does this PR do?

Addressing #223, fixes a crash in the Qwen3-Omni Thinker GSPO recipe when async rollout reaches
`AgentLoopWorker.generate_sequences()`:

```bash
ValueError: Cannot use chat template functions because tokenizer.chat_template is not set
```

The failure happens because `HFModelConfig` loads `processor=None` on Ray workers,
so `SingleTurnAgentLoop` falls back to a bare tokenizer without `chat_template`.

### Checklist Before Starting

- [x] Search for similar PRs. Paste at least one query link here: ...
- [x] Format the PR title as `[{modules}] {type}: {description}` (This will be checked by the CI)
  - `{modules}` include `fsdp`, `vllm_omni`, `rollout`, `trainer`, `ci`, `training_utils`, `recipe`, `ray`, `worker`, `single_controller`, `misc`, `perf`, `model`, `algo`, `env`, `tool`, `ckpt`, `doc`, `data`, `cfg`, `reward`, `diffusion`, `omni`, `tests`, `docker`
  - If this PR involves multiple modules, separate them with `,` like `[diffusion, doc]`
  - `{type}` is in `feat`, `fix`, `refactor`, `chore`, `test`
  - If this PR breaks any API (CLI arguments, config, function signature, etc.), add `[BREAKING]` to the beginning of the title.
  - Example: `[BREAKING][diffusion, fsdp] feat: new rollout scheduler`

On Ray AgentLoopWorker processes the import order is:

1. `agent_loop.py` imports `HFModelConfig` → `model.py` binds the unpatched hf_processor
2. `HFModelConfig.__post_init__` calls `import_external_libs(external_lib)` → patch replaces `verl.utils.tokenizer.hf_processor`
3. `model.py` still calls its stale local binding → processor stays `None`, and falls back to a bare tokenizer without `chat_template`.

### Test
```
bash examples/gspo_trainer/qwen3_omni/run_qwen3_omni_thinker_gspo_lora.sh
```

Before fix:

```bash
ray::AgentLoopWorker.generate_sequences() (pid=1944996, ip=192.168.17.94, actor_id=ee2bad47b10989e292891da501000000, repr=<verl.experimental.agent_loop.agent_loop.AgentLoopWorker object at 0x7fb65c232210>)
[...]

PR #208, opened 2026-06-24
[model, rollout, cfg, tests, doc] feat: Qwen3-Omni Thinker LoRA FSDP on transformers 5.x

qinganrice:
### What does this PR do?

Stacked on #211,  will rebase onto main after it merges

Adds **transformers 5.x** support to the Qwen3-Omni Thinker GSPO + LoRA recipe so the full MoE FFN (expert `gate_proj` / `up_proj` / `down_proj`) is LoRA-trainable on both 4.x and 5.x.

transformers 5.x stores the Thinker MoE experts as fused 3-D tensors (`gate_up_proj` / `down_proj` `nn.Parameter`), which PEFT cannot target for LoRA. This PR hooks `peft.get_peft_model` to unfuse those experts into per-expert `nn.Linear` right before LoRA injection (weights cloned from the fused params, forward semantics preserved), and no-ops PEFT 5.x's `gate_up_proj` name-remapping for Qwen3-Omni. On 4.x the experts are already `nn.Linear`, so the same `target_modules` config works unchanged.

Follow-up to the base Qwen3-Omni Thinker recipe (already merged); builds on latest `main`.

Related issues / PRs: https://github.com/verl-project/verl-omni/pull/113

### Checklist Before Starting

- [x] Search for similar PRs. Paste at least one query link here: <FILL: e.g. https://github.com/verl-project/verl-omni/pulls?q=qwen3-omni+thinker>
- [x] Format the PR title as `[{modules}] {type}: {description}`
  - `[model, rollout, cfg, tests, doc] feat: Qwen3-Omni Thinker MoE LoRA on transformers 5.x`

### Test

Validated end-to-end with the GSPO + LoRA smoke test (tiny random Qwen3-Omni, 2 steps, vLLM-Omni AR rollout, FSDP LoRA sync) and full model test:

Result: training completes both steps; rollout↔actor logprob `pearson ≈ 0.9998`. Verified PEFT targets the unfused MoE experts (48 expert LoRA adapters: 4 experts × {gate, up, down} × 2 layers × {A, B}). Re-added to `gpu_smoke` as **Test 8** (2 GPUs).

Full model training reward results (val is 0.887):
Screenshot 2026-06-26 at 12 26 22 PM

### API and Usage Example

No API change. The recipe config now lists the MoE FFN modules explicitly (`all-linear` cannot resolve the unfused experts on the multimodal wrapper):

```bash
[...]

PR #230, opened 2026-07-02
[ci] fix: update config vision_start_token_id for patched hf_processor

wtomin:
### What does this PR do?

Fix [ci error](https://github.com/verl-project/verl-omni/actions/runs/28571906249/job/84714570241) for qwen3-omni smoke test. After #224 was merged, the `verl.workers.config.model` has the patched hf_processor, so AgentLoop won't skip `get_rope_index`.

After #208 updates the transformers version to `5.x`,`get_rope_index` also uses `vision_start_token_id` from `talker_config`. However, the `processor.config` was overwritten with `thinker_config`, and has no access to `talker_config.vision_start_token_id`.

This PR fixes this problem by overwritting `processor.config.vision_start_token_id` with `config.talker_config.vision_start_token_id`

### Test

```bash
bash tests/special_e2e/run_gspo_qwen3_omni_thinker_lora_smoke.sh
```


SamitHuang (commented):
LGTM

SamitHuang (approved):

PR #189, opened 2026-06-18
[omni, rollout] feat: add Ascend NPU support for Qwen3-Omni Thinker GSPO training

panshaowu:
### What does this PR do?

Add Ascend NPU support for Qwen3-Omni Thinker model GSPO (Group Relative Policy Optimization) training, on top of the [PR #113](https://github.com/verl-project/verl-omni/pull/113).

Qwen3-Omni consists of three stages (Thinker, Talker, Code2Wav). For Thinker-only RL (GSPO/GRPO), a single stage must use ALL available NPUs of an Atlas 800T A3 server for tensor parallelism.

This PR provides:

1. A **Thinker-only stage config** (`qwen3_omni_thinker_only_npu.yaml`) for NPU deployments
2. A **full training script** (`run_qwen3_omni_thinker_gspo_npu.sh`) using 16 NPUs
3. **Device-aware worker extension selection** — GPU (any mode) and NPU+AR mode use the base extension; NPU+non-AR mode uses the NPU-extended extension with memory-pool / sleep / wake_up
4. **Tokenizer patch** (`verl_omni/models/transformers/qwen3_omni_thinker.py`) to load `chat_template` from `chat_template.json` for Qwen3-Omni models
5. **Bug fix** for `ar_mode` generate call — use `sampling_params_list` instead of `sampling_params`

### Checklist Before Starting

- [x] Search for similar PRs. Paste at least one query link here: https://github.com/verl-project/verl-omni/pull/113
- [x] Format the PR title as `[{modules}] {type}: {description}` (This will be checked by the CI)
  - `{modules}` include `fsdp`, `vllm_omni`, `rollout`, `trainer`, `ci`, `training_utils`, `recipe`, `ray`, `worker`, `single_controller`, `misc`, `perf`, `model`, `algo`, `env`, `tool`, `ckpt`, `doc`, `data`, `cfg`, `reward`, `diffusion`, `omni`, `tests`, `docker`
  - If this PR involves multiple modules, separate them with `,` like `[diffusion, doc]`
  - `{type}` is in `feat`, `fix`, `refactor`, `chore`, `test`
  - If this PR breaks any API (CLI arguments, config, function signature, etc.), add `[BREAKING]` to the beginning of the title.
  - Example: `[BREAKING][diffusion, fsdp] feat: new rollout scheduler`

### Test

[...]

PR #231, opened 2026-07-02
[recipe, doc, reward] feat: add MMK12 image2text recipe for Qwen3-Omni Thinker GSPO based on npu

wsc16:
### What does this PR do?
Adds an end-to-end **image → text** post-training recipe for `Qwen3-Omni-30B-A3B Thinker` GSPO on top of #189 (NPU support). Uses the [MMK12](https://www.modelscope.cn/datasets/SKYLENAGE/MMK12) K12 visual-math dataset, a `math_verify`-based rule reward, and reuses the existing NPU launcher (`run_qwen3_omni_thinker_gspo_npu.sh`, 16 × Atlas 800T A3, full-parameter FSDP + TP=2 rollout). The GPU / LoRA launcher from #208 works against MMK12 too after preprocessing.

Follow-up to #189 (Ascend NPU support) and #208 (Thinker LoRA on transformers 5.x), and [verl #6923](https://github.com/verl-project/verl/pull/6923)(Thinker on transformers 5.x); no dependency on other unmerged PRs.

Key additions:

- MMK12 → verl RL parquet converter  (`examples/gspo_trainer/data_process/mmk12.py`).
- `math_verify`-based rule reward with MC-content partial credit and a progressive `<answer>…\boxed{…}…</answer>` format ladder (`verl_omni/utils/reward_score/mmk12_reward.py`).
- Multimodal-input plumbing on the async rollout server: `audio_data` and `mm_processor_kwargs` accepted alongside `image_data` / `video_data`.
- NPU worker-thread device pin: fixes `Expects tensor to be on the compute device npu:N, was on npu:0` when FSDP offload / LoRA gather runs inside `asyncio.to_thread` on multi-visible-device NPU hosts.
- README restructured to document both modalities (text→text /image→text) and both platforms (GPU / NPU).

### Checklist Before Starting

- [x] Search for similar PRs. Paste at least one query link here: ...

### Test
Validated on **16 × Ascend NPU (Atlas 800T A3)** with `Qwen/Qwen3-Omni-30B-A3B-Instruct`, MMK12 image → text, full-parameter FSDP actor, TP=2 rollout, `rollout.n=8`.
The training result:
actor
image

critic/reward
image

val
image

### API and Usage Example
No new public API. Two additive optional args on `vLLMOmniHttpServer.generate(...)` (`audio_data`, `mm_processor_kwargs`), both defaulting to `None`.

```bash
# 1. Install the reward-scorer dependency (not pulled transitively).
pip install math-verify

# 2. Preprocess MMK12 shards into verl RL parquet.
python examples/gspo_trainer/data_process/mmk12.py \
    --input_dir  /path/to/mmk12/raw \
    --output_dir ~/data/mmk12

# 3. Launch training on 16 × Atlas 800T A3.
bash examples/gspo_trainer/qwen3_omni/run_qwen3_omni_thinker_gspo_npu.sh \
    reward.custom_reward_function.path=verl_omni/utils/reward_score/mmk12_reward.py \
    data.train_batch_size=32 \
    data.max_response_length=4096 \
    actor.ppo_mini_batch_size=32 \
    actor.optim.lr=2e-6
```

### Design & Code Changes
Main changed files:

- `examples/gspo_trainer/data_process/mmk12.py` — new. Raw MMK12 shards → verl RL parquet: `data_source="math_dapo"`, system prompt constrains the model to `<answer>…\boxed{…}…</answer>`, image bytes carried inline in the `images` column, in-question option lines parsed into `extra_info["options"]` for MC-content partial credit.
- `verl_omni/utils/reward_score/mmk12_reward.py` — new. Rule reward for MMK12, aligned with MM-EUREKA's `accuracy_reward_func`:
  - unified `math_verify` correctness check with `String + Latex + Expr` extraction configs (handles both letters and numeric / LaTeX uniformly),
  - MC-content fallback → `accuracy = 0.5` when the model gave the right option value but not the letter,
  - progressive format ladder on `<answer>` tag + `\boxed{}` presence (default `format_score = 0.3` → `0.0 / 0.15 / 0.30`),
  - total `= (accuracy + format) / (1 + format_score)`, normalized to `[0, 1]`.
- `verl_omni/workers/rollout/vllm_rollout/vllm_omni_async_server.py` — `vLLMOmniHttpServer.generate(...)`, `_build_multi_modal_data(...)`, `_preprocess_input(...)` extended to also accept `audio_data` and thread `mm_processor_kwargs` through to the AR engine prompt (needed for Qwen3-Omni's image / video / audio preprocessor on the rollout side). Backward-compatible: existing image-only / text-only call sites are unaffected.
- `verl_omni/workers/engine_workers.py` — worker-thread device pin. `ActorRolloutRefWorker._offload_actor_and_empty_cache(...)` and `_gather_lora_and_offload_actor(...)` run inside `asyncio.to_thread` from the async rollout path; the pool thread does not inherit the per-process default device set at worker init, so FSDP collectives materialized in those threads land on device 0 and trigger `Expects tensor to be on the compute device npu:N, was on npu:0` on NPU when `RAY_EXPERIMENTAL_NOSET_ASCEND_RT_VISIBLE_DEVICES=1`
  keeps all devices visible. Added a small `_set_worker_thread_device()` helper and call it at the top of both
  paths. GPU is unaffected.
- `examples/gspo_trainer/README.md` — restructured to cover both modalities (text→text / image→text) and both platforms (GPU / NPU), with a `Training with MMK12` section that walks through preprocessing + reward wiring + launch.
- `verl_omni/models/transformers/qwen3_omni_thinker.py` - **weight loader**: registers `Qwen3OmniMoeThinkerForConditionalGeneration` on the vllm-moe-model weight-loader patch list, so the fused 3-D MoE expert tensors are split and loaded into per-expert `nn.Linear` correctly under transformers 5.x. Without this, the Thinker loads with silently-mismatched expert weights. Depends on [verl #6923](https://github.com/verl-project/verl/pull/6923); **get_rope_index**: cast dtype of position_ids from fp32 to int64 to avoid BF16 large-int rounding in root-input casting of FSDP.

> [!IMPORTANT]

### DIFF
--- a/examples/gspo_trainer/README.md
+++ b/examples/gspo_trainer/README.md
@@ -1,8 +1,16 @@
-# Qwen3-Omni Thinker GSPO + LoRA Trainer
+# Qwen3-Omni Thinker GSPO Trainer
 
 This example shows how to post-train the **Qwen3-Omni-30B-A3B Thinker** with
-**GSPO + LoRA** on a math-reasoning task, using FSDP for the actor and
-`vllm-omni` as the async rollout backend.
+**GSPO** on a math-reasoning task, using FSDP for the actor and `vllm-omni` as
+the async rollout backend. Two input modalities are supported: **text → text**
+(e.g. `MATH-lighteval`) and **image → text** (e.g. `MMK12`).
+
+Both **GPU** and **NPU** training platforms are supported via two launch scripts:
+
+- [`run_qwen3_omni_thinker_gspo_lora.sh`](qwen3_omni/run_qwen3_omni_thinker_gspo_lora.sh)
+  — **GPU**, **LoRA (r=64)** on a single node with **4 × H100/H200 80GB**.
+- [`run_qwen3_omni_thinker_gspo_npu.sh`](qwen3_omni/run_qwen3_omni_thinker_gspo_npu.sh)
+  — **NPU**, **full-parameter** on a single **Atlas 800T A3** node with **16 NPUs**.
 
 For the base environment setup, see the [installation guide](../../docs/start/install.md).
 
@@ -71,10 +79,11 @@ so verl loads both on the driver via its `VERL_USE_EXTERNAL_MODULES` hook:
 patches before the driver's dataset loader runs. The GPU workers load the same
 model patch via `actor_rollout_ref.model.external_lib`.
 
-The provided script is configured for a single node with **4 × H100/H200 80GB**:
-the actor (FSDP, 30B + LoRA r=64 with param/optimizer offload) and the
-`vllm-omni` rollout (TP=4) colocate on the same 4 GPUs. Multi-node is not yet
-validated.
+The two launch scripts colocate the FSDP actor and the `vllm-omni` rollout on
+the same devices. `run_qwen3_omni_thinker_gspo_lora.sh` targets a single node with **4 × H100/H200 80GB**
+(30B + LoRA r=64 with param/optimizer offload, rollout TP=4). `run_qwen3_omni_thinker_gspo_npu.sh` targets
+a single **Atlas 800T A3** node with **16 NPUs** (full-parameter FSDP actor,
+rollout TP=2). Multi-node is not yet validated on either platform.
 
 > **Where the rollout engine's memory/batching is set.** When
 > `stage_configs_path` is provided, vLLM-Omni **ignores** the top-level engine
@@ -90,20 +99,6 @@ validated.
 > the codepaths used here are numpy-2 compatible, so the pip resolver warning is
 > safe to ignore.
 
-## Prepare the dataset
-
-A parquet dataset of math problems with `prompt` and `answer` fields, defaulting
-to `~/data/math/{train,test}.parquet`. The example was tested on
-`MATH-lighteval`; any standard RL math dataset works. To convert HuggingFace
-datasets into verl's parquet format, see
-[`verl/examples/data_preprocess/`](https://github.com/verl-project/verl/tree/main/examples/data_preprocess).
-
-```bash
-mkdir -p ~/data/math
-# … place train.parquet and test.parquet here …
-ls ~/data/math/   # train.parquet  test.parquet
-```
-
 ## Prepare the model
 
 The script uses the HuggingFace Hub ID `Qwen/Qwen3-Omni-30B-A3B-Instruct`
@@ -118,12 +113,32 @@ export MODEL_PATH=/path/to/local/Qwen3-Omni-30B-A3B-Instruct
 > `tokenizer.chat_template`; verl's dataset loader calls
 > `tokenizer.apply_chat_template(...)` and fails without it.
 
-## Run training
+## Training with `MATH-lighteval`
 
-Launch from the repository root:
+### Prepare the dataset
+
+A parquet dataset of math problems with `prompt` and `answer` fields, defaulting
+to `~/data/math/{train,test}.parquet`. The example was tested on
+`MATH-lighteval`; any standard RL math dataset works. To convert HuggingFace
+datasets into verl's parquet format, see
+[`verl/examples/data_preprocess/`](https://github.com/verl-project/verl/tree/main/examples/data_preprocess).
 
 ```bash
+mkdir -p ~/data/math
+# … place train.parquet and test.parquet here …
+ls ~/data/math/   # train.parquet  test.parquet
+```
+
+### Run training
+
+Launch from the repository root — pick the flavor that matches your hardware:
+
+```bash
+# GPU, LoRA (r=64), 4 × H100/H200
 bash examples/gspo_trainer/qwen3_omni/run_qwen3_omni_thinker_gspo_lora.sh
+
+# NPU, full-parameter, 16 × Atlas 800T A3
+bash examples/gspo_trainer/qwen3_omni/run_qwen3_omni_thinker_gspo_npu.sh
 ```
 
 The recipe config lives in
@@ -152,21 +167,16 @@ which trains on a tiny random-weight model built by
 [`build_qwen3_omni_tiny_random.py`](../../tests/special_e2e/build_qwen3_omni_tiny_random.py)
 (no 60 GB download). It is wired into the `tests/gpu_smoke` CI suite as Test 8.
 
-## Logging
-
-W&B logging is enabled by default:
-
-```bash
-export WANDB_API_KEY=<your_wandb_api_key>
-# trainer.project_name / experiment_name are already set in the script
-```
-
-## What is trained
+### What is trained
 
 Only the **Thinker** (`Qwen3OmniMoeThinkerForConditionalGeneration`):
 
-- LoRA rank 64, alpha 32, on `target_modules="q_proj,k_proj,v_proj,o_proj,gate_proj,up_proj,down_proj"`
-  (explicit names so the unfused MoE expert `gate/up/down_proj` are targeted, not just attention).
+- **GPU (LoRA)** — rank 64, alpha 32, on
+  `target_modules="q_proj,k_proj,v_proj,o_proj,gate_proj,up_proj,down_proj"`
+  (explicit names so the unfused MoE expert `gate/up/down_proj` are targeted,
+  not just attention).
+- **NPU (full-parameter)** — LoRA is disabled (`lora_rank=0`); all Thinker
+  parameters are updated under FSDP.
 - `exclude_modules` strips talker / code2wav / code_predictor / visual /
   audio_tower; `freeze_vision_tower=True` keeps the vision encoder cold.
 - The non-Thinker heads are dropped at FSDP-wrap time via `_verl_strip_modules`.
@@ -182,10 +192,11 @@ sampled response length):
   (`actor/perf/max_memory_allocated_gb` < 65).
 - `val-core/.../acc/mean@1` rising with steps.
 
-## Performance
+### Performance (GPU + LoRA)
 
 > Measured on a single node of **4 × H100/H200 80GB**, actor and rollout
-> colocated, MATH-lighteval, `dapo` reward.
+> colocated, MATH-lighteval, `dapo` reward, LoRA r=64
+> (`run_qwen3_omni_thinker_gspo_lora.sh`).
 
 | Script | Model | Algorithm | # Cards (colocate) | Batch × `rollout.n` | lr | Steps | Throughput (tok/gpu/s) | Time / Step (s) | val acc/mean@1 | rollout↔actor pearson |
 | --- | --- | --- | --- | --- | --- | --- | --- | --- | --- | --- |
@@ -196,9 +207,9 @@ response length (mean `response_length` ranged ~0.7k–3.6k tokens over the run,
 to `max_response_length=8192` with `rollout.n=8`); `actor/perf/max_memory_allocated`
 peaks at ~64 GB.
 
-## Preliminary results
+### Preliminary results (GPU + LoRA)
 
-Over a 60-step run with the default config, the training reward
+Over a 60-step run with the default LoRA config, the training reward
 (`critic/rewards/mean`) rises from **~0.75 to ~0.90** and validation accuracy on
 MATH-lighteval is **~0.90**. Treat this
 as a plumbing-correctness signal (finite loss, reasonable grad norm, rollout↔actor
@@ -209,15 +220,84 @@ low-variance advantages on a high-baseline policy.
 
 ![training reward](reward.png)
 
+## Training with `MMK12`
+
+For visual math reasoning we ship an end-to-end pipeline on top of the
+[MMK12](https://huggingface.co/datasets/FangqingM/MMK12) dataset (image
+input + text output, K12 math). It reuses the same GSPO recipe as the
+text-only path — only the data preprocessing and the reward scorer differ.
+The launch example below uses the NPU / full-parameter flavor
+(`run_qwen3_omni_thinker_gspo_npu.sh`), but the GPU / LoRA flavor works too
+after pointing it at the MMK12 parquet.
+
+### Prepare the dataset
+
+Download the raw MMK12 parquet shards (from ModelScope or HuggingFace) into a
+local directory — the loader expects filenames like `train-*.parquet` and
+`test-*.parquet` — and convert them into the verl RL parquet layout with:
+
+```bash
+python examples/gspo_trainer/data_process/mmk12.py \
+    --local_dataset_path /path/to/mmk12/ \
+    --local_save_dir ~/data/mmk12
+```
+
+The converter emits one verl RL row per problem, with
+`data_source="math_dapo"`, a system prompt that constrains the model to emit
+`<answer>…\boxed{…}…</answer>`, and the image bytes carried inline in the
+`images` column so the parquet stays self-contained. Input / kept / dropped
+counts and answer-type tallies are printed at the end. See the module docstring
+in [`data_process/mmk12.py`](data_process/mmk12.py) for the exact output schema.
+
+### Run training
+
+The MMK12 reward scorer grades responses with
+[`math_verify`](https://github.com/huggingface/math-verify), which is **not**
+pulled in transitively by verl or verl-omni. Install it explicitly first —
+otherwise the scorer falls back to `accuracy = 0` for every sample:
+
+```bash
+pip install math-verify
+```
+
+Then point the launcher at the MMK12 parquet and register the custom reward
+scorer via CLI overrides (no yaml edits required):
+
+```bash
+TRAIN_FILE=$HOME/data/mmk12/train.parquet \
+VAL_FILE=$HOME/data/mmk12/test.parquet \
+bash examples/gspo_trainer/qwen3_omni/run_qwen3_omni_thinker_gspo_npu.sh \
+    reward.custom_reward_function.path=verl_omni/utils/reward_score/mmk12_reward.py \
+    actor_rollout_ref.actor.optim.lr=3e-6
+```
+
+The scorer combines `math_verify` accuracy with a progressive format reward on
+the `<answer>…\boxed{}…</answer>` template; see
+[`verl_omni/utils/reward_score/mmk12_reward.py`](../../verl_omni/utils/reward_score/mmk12_reward.py)
+for the full formula.
+
+## Logging
+
+W&B logging is enabled by default:
+
+```bash
+export WANDB_API_KEY=<your_wandb_api_key>
+# trainer.project_name / experiment_name are already set in the script
+```
+
 ## File map
 
 ```
 examples/gspo_trainer/
 ├── qwen3_omni/
-│   ├── run_qwen3_omni_thinker_gspo_lora.sh   ← launch script (volatile overrides only)
+│   ├── run_qwen3_omni_thinker_gspo_lora.sh   ← launch script (GPU, LoRA r=64)
+│   ├── run_qwen3_omni_thinker_gspo_npu.sh    ← launch script (NPU, full-parameter)
 │   ├── config/
 │   │   └── qwen3_omni_thinker_gspo.yaml      ← recipe config (inherits verl ppo_trainer)
-│   └── qwen3_omni_thinker_only.yaml          ← vllm-omni stage config
-├── reward.png                            ← preliminary reward curve
-└── README.md                             ← (this file)
+│   ├── qwen3_omni_thinker_only.yaml          ← vllm-omni stage config (GPU)
+│   └── qwen3_omni_thinker_only_npu.yaml      ← vllm-omni stage config (NPU)
+├── data_process/
+│   └── mmk12.py                              ← MMK12 → verl RL parquet converter
+├── reward.png                                ← preliminary reward curve
+└── README.md                                 ← (this file)
 ```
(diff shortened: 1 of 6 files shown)


verl_omni/utils/reward_score/mmk12_reward.py
SamitHuang, line 1: how to test the correctness of this reward function?

verl_omni/workers/engine_workers.py
```diff
@@ -775,6 +775,21 @@ def ema_update_adapter(self, source: str = "default", target: str = "old", decay
         assert "actor" in self.role, "ema_update_adapter only supports actor role"
         self.actor.ema_update_adapter(source=source, target=target, decay=decay)
 
+    def _set_worker_thread_device(self):
```
knlnguyen1802: This is quite weird, can we move the fix to initialize actor instead of doing it here ?

zhtmike:
A reward curve should be provided, thx

verl_omni/workers/engine_workers.py
wsc16: The worker thread has initialized device already, while the 'gather' thread does not inherit it. The function "set_device" is thread-local (the current device of torch.cuda or torch_npu is thread-specific), and the "to_thread" submits the task to run in another thread of the default executor - the device set by the main thread is not passed along.

zhtmike:
seems not converging. what is you validation curve

verl_omni/utils/reward_score/mmk12_reward.py
wsc16: The reward method refers to the open-source MM-EUREKA. And I tested the reward method on the inference results generated by vllm directly, almost all the data could be processed correctly.

wsc16:
> seems not converging. what is you validation curve

Probably, or it just seems to be growing at a relatively slow pace. The format of the test set and training set in the mmk12 dataset is different. And the training effect on the training set does not perform well on the test set which consists entirely of multiple-choice questions. Perhaps I should sample a part of the training set as the test data? Or maybe there are some issues, I'm not quite sure.

zhtmike:
yes, we need a plausible result before review, thx

verl_omni/models/transformers/qwen3_omni_thinker.py
```diff
@@ -323,12 +323,102 @@ def _patched_hf_tokenizer(name_or_path, *args, **kwargs):
...
+
+
 def apply_qwen3_omni_thinker_patches() -> None:
     """Apply all Qwen3-Omni Thinker patches (idempotent registrations)."""
     _register_qwen3_omni_automodel()
     patch_hf_processor_for_qwen3_omni()
     _patch_unfuse_qwen3_omni_thinker_experts()
     patch_hf_tokenizer_for_qwen3_omni()
+    patch_register_vllm_moe_model_weight_loader()
+    patch_agent_loop_position_ids_for_qwen3_omni()
```
zhtmike: this is just get crazy

wsc16: patch for weight loader is needed, while the modification of position_ids may be  a mistake and i will check it again. thx~

wsc16:
> > seems not converging. what is you validation curve
> 
> Probably, or it just seems to be growing at a relatively slow pace. The format of the test set and training set in the mmk12 dataset is different. And the training effect on the training set does not perform well on the test set which consists entirely of multiple-choice questions. Perhaps I should sample a part of the training set as the test data? Or maybe there are some issues, I'm not quite sure.

position_id is cast to bf16 in fsdp because MixedPrecision set cast_root_forward_inputs=True by default. This will lead to accuracy errors, especially when the values are relatively large. e.g. 2779 will become 2880 when converting from the fp32 type to the bf16 type.

zhtmike:
> > > seems not converging. what is you validation curve
> > 
> > 
> > Probably, or it just seems to be growing at a relatively slow pace. The format of the test set and training set in the mmk12 dataset is different. And the training effect on the training set does not perform well on the test set which consists entirely of multiple-choice questions. Perhaps I should sample a part of the training set as the test data? Or maybe there are some issues, I'm not quite sure.
> 
> position_id is cast to bf16 in fsdp because MixedPrecision set cast_root_forward_inputs=True by default. This will lead to accuracy errors, especially when the values are relatively large. e.g. 2779 will become 2880 when converting from the fp32 type to the bf16 type.

we can use fsdp1 for this (it is much safe); or check how verl handles with rope under fsdp2. It is using `cast_root_forward_inputs=True` as well.

wsc16:
> > > > seems not converging. what is you validation curve
> > > 
> > > 
> > > Probably, or it just seems to be growing at a relatively slow pace. The format of the test set and training set in the mmk12 dataset is different. And the training effect on the training set does not perform well on the test set which consists entirely of multiple-choice questions. Perhaps I should sample a part of the training set as the test data? Or maybe there are some issues, I'm not quite sure.
> > 
> > 
> > position_id is cast to bf16 in fsdp because MixedPrecision set cast_root_forward_inputs=True by default. This will lead to accuracy errors, especially when the values are relatively large. e.g. 2779 will become 2880 when converting from the fp32 type to the bf16 type.
> 
> we can use fsdp1 for this (it is much safe); or check how verl handles with rope under fsdp2. It is using `cast_root_forward_inputs=True` as well.

yeah, the get_rope_index method returns position_id with dtype fp32 (and this looks like a bug of transformers: the return type of the function declaration is long, but the actual type is fp32). In agent_loop, it is concatenated with text_position_ids (int64). Since torch's cat operator performs implicit type conversion (type promotion), the resulting concatenated position_id ends up being fp32.

zhtmike:
> > > > > seems not converging. what is you validation curve
> > > > 
> > > > 
> > > > Probably, or it just seems to be growing at a relatively slow pace. The format of the test set and training set in the mmk12 dataset is different. And the training effect on the training set does not perform well on the test set which consists entirely of multiple-choice questions. Perhaps I should sample a part of the training set as the test data? Or maybe there are some issues, I'm not quite sure.
> > > 
> > > 
> > > position_id is cast to bf16 in fsdp because MixedPrecision set cast_root_forward_inputs=True by default. This will lead to accuracy errors, especially when the values are relatively large. e.g. 2779 will become 2880 when converting from the fp32 type to the bf16 type.
> > 
> > 
> > we can use fsdp1 for this (it is much safe); or check how verl handles with rope under fsdp2. It is using `cast_root_forward_inputs=True` as well.
> 
> yeah, the get_rope_index method returns position_id with dtype fp32 (and this looks like a bug of transformers: the return type of the function declaration is long, but the actual type is fp32). In agent_loop, it is concatenated with text_position_ids (int64). Since torch's cat operator performs implicit type conversion (type promotion), the resulting concatenated position_id ends up being fp32.

yeah, I think fp32 is fine, but need to aware to avoid the cast_foward_inputs=True ..

wsc16:
> > > > > > seems not converging. what is you validation curve
> > > > > 
> > > > > 
> > > > > Probably, or it just seems to be growing at a relatively slow pace. The format of the test set and training set in the mmk12 dataset is different. And the training effect on the training set does not perform well on the test set which consists entirely of multiple-choice questions. Perhaps I should sample a part of the training set as the test data? Or maybe there are some issues, I'm not quite sure.
> > > > 
> > > > 
> > > > position_id is cast to bf16 in fsdp because MixedPrecision set cast_root_forward_inputs=True by default. This will lead to accuracy errors, especially when the values are relatively large. e.g. 2779 will become 2880 when converting from the fp32 type to the bf16 type.
> > > 
> > > 
> > > we can use fsdp1 for this (it is much safe); or check how verl handles with rope under fsdp2. It is using `cast_root_forward_inputs=True` as well.
> > 
> > 
> > yeah, the get_rope_index method returns position_id with dtype fp32 (and this looks like a bug of transformers: the return type of the function declaration is long, but the actual type is fp32). In agent_loop, it is concatenated with text_position_ids (int64). Since torch's cat operator performs implicit type conversion (type promotion), the resulting concatenated position_id ends up being fp32.
> 
> yeah, I think fp32 is fine, but need to aware to avoid the cast_foward_inputs=True ..

Yes, I completely agree with you. And I modified the data type to long in `qwen3_omni_moe_thinker`, thanks for reviewing.

examples/gspo_trainer/README.md
```diff
@@ -152,21 +167,16 @@ which trains on a tiny random-weight model built by
...
 Only the **Thinker** (`Qwen3OmniMoeThinkerForConditionalGeneration`):
 
-- LoRA rank 64, alpha 32, on `target_modules="q_proj,k_proj,v_proj,o_proj,gate_proj,up_proj,down_proj"`
-  (explicit names so the unfused MoE expert `gate/up/down_proj` are targeted, not just attention).
+- **GPU (LoRA)** — rank 64, alpha 32, on
+  `target_modules="q_proj,k_proj,v_proj,o_proj,gate_proj,up_proj,down_proj"`
+  (explicit names so the unfused MoE expert `gate/up/down_proj` are targeted,
+  not just attention).
+- **NPU (full-parameter)** — LoRA is disabled (`lora_rank=0`); all Thinker
+  parameters are updated under FSDP.
```
zhtmike, line 179: why we have this distinction?

verl_omni/models/transformers/qwen3_omni_thinker.py
```diff
@@ -132,7 +135,18 @@ def _patched_hf_processor(name_or_path, **kwargs):
             processor.spatial_merge_size = config.thinker_config.vision_config.spatial_merge_size
             processor.config.vision_start_token_id = config.talker_config.vision_start_token_id
             model_class = Qwen3OmniMoeThinkerForConditionalGeneration
-            processor.get_rope_index = types.MethodType(model_class.get_rope_index, processor)
+            # upstream get_rope_index returns float32. cast to int64 so agent_loop's
+            # cat(text_position_ids=int64, vision_position_ids) yields int64 and FSDP
+            # root-input casting (floating dtypes only) leaves positions exact -
+            # avoids BF16 large-int rounding (e.g. 2879 -> 2880) that breaks SDPA
+            # packed-sequence detection under use_remove_padding.
```
zhtmike: pls make markdown simple

verl_omni/utils/reward_score/reward_utils.py
```diff
@@ -32,3 +43,74 @@ def pil_image_to_base64(image: Image.Image) -> str:
...
+    """
+    gt_boxed = "\\boxed{" + ground_truth + "}"
+    try:
+        future = _get_math_verify_pool().submit(_math_verify_worker, gt_boxed, model_output)
+        return future.result(timeout=timeout)
+    except (FuturesTimeoutError, TimeoutException):
+        return 0.0
+    except Exception as e:
+        print(f"math_verify_score error: {e}", flush=True)
+        return 0.0
```
zhtmike: what happens here? can you provide a simple explanation?

verl_omni/workers/rollout/vllm_rollout/utils.py
```diff
@@ -144,6 +144,13 @@ def update_weights_from_ipc(self, peft_config: dict = None, base_sync_done=False
                 # model.load_weights (no per-bucket finalize), then run the single
                 # post-load processing pass once all buckets are received.
                 model, model_config = standard
+                # Re-attach weight_loader on Ascend FusedMoE params via verl's
+                # built-in patch (handles ACLGraph unwrap + SUPPORTED_MOE_MODELS
+                # whitelist, which Qwen3-Omni is registered into via
+                # patch_register_vllm_moe_model_weight_loader).
+                from verl.utils.vllm.patch import patch_vllm_moe_model_weight_loader
+
+                patch_vllm_moe_model_weight_loader(model)
```
zhtmike, line 153: If I am right, we can simply inherit the vllm/vllm-omni class with overwriting the `load_weights` method? won't it looks more clear?

verl_omni/workers/rollout/vllm_rollout/vllm_omni_async_server.py
```diff
@@ -306,6 +320,62 @@ def _map_stop_reason(finish_reason: Optional[str]) -> Optional[str]:
...
+    @staticmethod
+    def _dedup_multimodal_pad_tokens(prompt_ids: list[int], processor) -> list[int]:
+        """Collapse consecutive multimodal pad tokens back to a single sentinel.
+
+        When we call ``self.processor(text=..., images=...)`` in
+        ``AgentLoopBase.apply_chat_template`` the HF processor expands a single
+        ``<|image_pad|>`` / ``<|video_pad|>`` / ``<|audio_pad|>`` placeholder into
+        ``N`` consecutive pad tokens (one per vision/audio patch). If we then
+        ship those expanded ids to vLLM-Omni together with raw
+        ``multi_modal_data``, vLLM-Omni's ``_apply_prompt_updates`` will
```
zhtmike: so vllm-omni's _apply_prompt_updates is buggy?

verl_omni/workers/rollout/vllm_rollout/vllm_omni_async_server.py
```diff
@@ -306,6 +320,62 @@ def _map_stop_reason(finish_reason: Optional[str]) -> Optional[str]:
...
+        if arr.size == 0:
+            return prompt_ids
+        is_pad = np.isin(arr, list(pad_ids))
+        # Keep position i if: not a pad token, OR it is a pad token whose
+        # predecessor is not the *same* pad token (i.e. collapse runs of
+        # identical pad tokens to one).
+        keep = np.ones(arr.size, dtype=bool)
+        same_as_prev = is_pad[1:] & is_pad[:-1] & (arr[1:] == arr[:-1])
+        keep[1:] &= ~same_as_prev
+        return arr[keep].tolist()
```
zhtmike: pls follow `verl`'s convention, moving the patch into model specific folder

verl_omni/workers/engine_workers.py
```diff
@@ -786,6 +801,7 @@ def _offload_actor_and_empty_cache(self, timings: Optional[dict] = None):
         tensors live in separate allocations that are unaffected by moving the
         base param storage to CPU.
         """
+        self._set_worker_thread_device()
```
zhtmike: revert this according to our discussion?

examples/gspo_trainer/README.md
```diff
@@ -209,15 +220,85 @@ low-variance advantages on a high-baseline policy.
 
 ![training reward](reward.png)
 
+## Training with `MMK12`
+
+For visual math reasoning we ship an end-to-end pipeline on top of the
+[MMK12](https://www.modelscope.cn/datasets/SKYLENAGE/MMK12) dataset (image
```
ZihaoW123: The link is inaccessible. Please provide the correct link.

verl_omni/utils/reward_score/mmk12_reward.py
```diff
@@ -0,0 +1,147 @@
...
+    """MMK12 reward entrypoint.
+
+    Reward formula (additive, normalized to [0, 1]):
+        score = (accuracy_reward + format_reward) / (1 + format_score)
+
+    * ``accuracy_reward`` ∈ {0, 1} — math_verify with String+Latex+Expr configs.
+    * ``format_reward`` ∈ [0, ``format_score``] — progressive (answer_tag + boxed).
+
+    Config:
+        reward.custom_reward_function.path=verl_omni/utils/reward_score/mmk12.py
```
wtomin:
```suggestion
        reward.custom_reward_function.path=verl_omni/utils/reward_score/mmk12_reward.py
```

verl_omni/workers/rollout/vllm_rollout/utils.py
knlnguyen1802: Actually, it can put here because we define a new class WorkerExtension for vllm-omni, it's not inherit from vllm

zhtmike: @knlnguyen1802 you can check the implementation of bagel here~

examples/gspo_trainer/README.md
wsc16: > why we have this distinction?

moe-lora is not supported in vllm-ascend for now. [issue](https://github.com/vllm-project/vllm-ascend/issues/9756)

zhtmike: i see

verl_omni/workers/rollout/vllm_rollout/vllm_omni_async_server.py
wsc16: both of the hf processor and vllm's `_apply_prompt_updates` expand pad ids. while it has no effect on the inference results, revert this change.

zhtmike: I see

verl_omni/utils/reward_score/reward_utils.py
wsc16: Run math_verify in a ProcessPoolExecutor to isolate its signal.alarm() from
reward-manager threads; worker sits at module top-level because file-loaded
scorers get a synthetic module name the subprocess cannot re-import.

zhtmike: Don't get it,

simply set 
```bash 
# for example
reward.reward_manager.source=register \
reward.reward_manager.name=naive \
```

not working?

zhtmike:
is it ready for review? if yes, please drop `[draft]`

zyxzyxzyx:
## Response to "not converging": PR#231 fix validated, 100-step converges with val +8.5%

@zhtmike @wsc16 Regarding the "seems not converging" feedback, I (PR #254) validated that PR#231's `position_id` fix resolves the pearson 0.13 misalignment, and ran a healthy 100-step training (val +8.5%, no collapse).

### 1. "not converging" root cause = pearson 0.13 (rollout<->actor logprob misalignment)

The root cause and fix come from PR#231 (thanks @wsc16): `training/rollout_actor_probs_pearson_corr = 0.13` (should be >0.95; text->text is 0.989) made the training signal noise (log_ppl_diff 6.30 nats), so val didn't converge. PR#231 identified the root cause as **position_id precision (bf16 non-integer)** — bf16 large-position (T~8k) rounding causes (1) `find_packed_sequence_indices` misjudgment -> wrong attention mask, (2) rotary cos/sin drift, (3) actor bf16 vs vllm int decode drift. text->text unaffected (monotonic positions -> 0.989).

### 2. Validated PR#231's `processor .long()` fix

PR#231's `_patched_hf_processor` wraps `get_rope_index` to return `vision_position_ids.long()` (int64), covering agent_loop + actor at the processor layer. I validated it:

| metric | before fix | step-1 after fix |
|---|---|---|
| pearson | 0.133 | **0.9886** |
| log_ppl_diff | 6.30 | -0.025 (approx 0) |
| training_ppl | 888 | 1.36 (approx rollout_ppl 1.40) |

### 3. 50-step passed (converges, vs text->text PR#189)

50-step GSPO (10.5h, 0 crash): pearson 0.945->0.977, val acc 72.3%->78.7% (+6.4%), val reward 74.2%->83.3% (+9.1%). image->text validated.

### 4. 100-step: degradation located + tf5 anti-collapse (val +8.5%)

**100-step old config (lr=3e-5/clip=0.02/batch=8) degraded after step 50**: entropy collapse (0.28->0.07), val drop (77.7%->69.7%, step 50 crashed to 59.6%), format_reward->0. Alignment still OK (pearson 0.915) — degradation is RL instability, not alignment.

**tf5 tuned config (lr=2e-6 (down 15x) + clip=1e-3 (down 20x) + batch=32 (up 4x) + max_resp=4096 + shuffle) successfully prevented collapse** (18h29m, 0 crash):

| val point | tf5 anti-collapse | old degraded |
|---|---|---|
| step 0 | 68.1% | 77.7% |
| step 25 | 75.0% | 75.5% |
| step 50 | 71.3% (held) | 59.6% (crashed) |
| step 75 | 72.9% | 70.2% |
| step 100 | **76.6% (+8.5%)** | 69.7% (-8%) |

- val accuracy +8.5% (68.1%->76.6%), reward +9.0% (70.7%->79.7%)
- entropy stable 0.27-0.39 throughout (no collapse), format_reward rising (0.238->0.270)
- pearson 0.988, log_ppl_diff approx 0, 0 crash

Training curve plots in PR#254 comments (part1/part2).

### 5. tf5 env upgrade (removed compatibility workarounds)

transformers 4.57.6->5.5.3, accelerate 1.12.0->1.13.0 (fix `_is_hf_initialized`), verl 8a694930 + verl PR#6923 (transformers 5 packed MoE -> vLLM per-expert weight sync), verl-omni reverted to PR#231 clean version (8a694930 verl monkey_patch uses `get_text_config` fallback, replacing the text_config property workaround). step-1 tf5 pearson 0.9887 (= 4.57.6 + compat-fix 0.9886).

### Contribution summary (@zyxzyxzyx, PR #254)

Note: the root cause (position_id bf16) and fix (processor `.long()`) are from PR#231 (@wsc16). My contributions:
1. **Validated PR#231's `processor .long()` fix** effective (pearson 0.13->0.99)
2. **Passed 50-step** (val 78.7%, vs PR#189 standard)
3. **Located 100-step degradation** (entropy collapse) and **tf5 tuning prevented collapse** (val +8.5%)
4. **tf5 env upgrade** (transformers 5.5.3 + verl PR#6923, removed compat workarounds, PR#231 clean version validated)

Env: Ascend 910C x16, transformers 5.5.3, accelerate 1.13.0, verl 8a694930 + PR#6923, vllm-omni 0.22.1. Full investigation and training curves in PR #254.

SamitHuang:
Nice job.
1. pls fix linting
2. the main branch has updated vllm-omni to 0.24 (#257 ). Can you rebase and check whether this PR is still runnable?

zhtmike:
https://github.com/vllm-project/vllm-omni/issues/5369 @wsc16

zhtmike:
pls address ai review

SamitHuang (approved):
LGTM. Thanks for the patience.

PR #242, opened 2026-07-07
[reward] fix: drop duplicate critic/score metric for single-reward training

cr-gao:
## What

Fixes #121. For single-reward flow-grpo training, `critic/score/mean` was always identical to `critic/rewards/mean`.

Root cause: `VisualRewardManager.run_single` copies **every** field of a dict reward result into `reward_extra_info`, including the primary `score`. Since `score` is the reward itself (already reported as `critic/rewards/mean` by `compute_data_metrics_diffusion`), re-exporting it made `compute_reward_extra_metrics_diffusion` emit a duplicate `critic/score/mean`. The fix skips the `score` key.

## Not duplicating existing work

No open PR touches `visual.py` / `diffusion_metric_utils.py`, and no PR references #121. Since the issue was filed, the only change to those files (#106) did not address the duplication.

## Testing

- Ran `ruff check` on both changed files — passed.
- A/B verification against the real code (git HEAD vs this branch), driving the real `run_single` -> real `compute_reward_extra_metrics_diffusion`:
  - **before:** `critic/score/mean` present and equal to `critic/rewards/mean` (duplicate)
  - **after:** `critic/score/mean` no longer produced; `critic/rewards/mean` unchanged

## Related but not addressed (kept out of scope)

- Non-dict branch sets `reward_extra_info["acc"] = score`, which similarly duplicates `critic/rewards/mean`. Left as-is because `acc` is used as the validation core variable (`ray_diffusion_trainer.py`); changing it risks a validation regression.
- `MultiVisualRewardManager` emits `reward/combined` == `reward/{key}` for a single sub-reward — same class of issue, better as a follow-up.

Issue #121, opened 2026-05-29
[chore]: Duplicated critic reward/score metrics for flow-grpo training with single-reward


SamitHuang:
### System Info

<img width="2122" height="1206" alt="Image" src="https://github.com/user-attachments/assets/bba03fbe-2c2b-4e3d-af8c-c55f05c2d298" />

critic reward mean and critic score mean are the same when using single reward for training.  Incurred from Incurred by https://github.com/verl-project/verl-omni/pull/109. 

### Information

- [ ] The official example scripts
- [ ] My own modified scripts

### Tasks

- [ ] An officially supported task in the `examples` folder (such as QwenImage, ...)
- [ ] My own task or dataset (give details below)

### Reproduction

bash examples/flowgrpo_trainer/run_qwen_image_ocr_lora.sh

### Expected behavior

When using single reward, we can remove critic score mean

PR #246, opened 2026-07-08
[reward] fix: drop duplicate reward/<key>/score metric for multi-reward training

cr-gao:
# [reward] fix: drop duplicate reward/<key>/score metric for multi-reward training

Follow-up to #242

## What & why

#242 removed the duplicate `critic/score/mean` metric in
`VisualRewardManager` by not copying a dict reward's primary `score` field into
`reward_extra_info`. `MultiVisualRewardManager.run_single` has the same issue:
for a dict sub-reward it copies **every** field — including `score` — into the
per-reward namespace, so it emits both

- `reward/<key>/score` (from the field-copy loop), and
- `reward/<key>` (the primary score, set explicitly right after)

which are identical. This mirrors #121/#242 exactly, so the fix mirrors it:
skip `score` when copying the auxiliary fields.

## Testing
- Ran a local cpu end-to-end test that drives the real `run_single`
  output through the real `compute_reward_extra_metrics_diffusion`, asserting the
  logged metrics contain no `critic/reward/ocr/score/mean` (the duplicate) while
  `critic/reward/ocr/mean` and the auxiliary `reward/ocr/genrm_response` field
  are intact — end-to-end proof that the duplicate metric is gone.
- `ruff check` and `ruff format --check` pass on all changed files.

This is the multi-reward counterpart explicitly requested on #242; no open PR
addresses multi-reward metric deduplication.

### DIFF
--- a/tests/reward_loop/test_multi_reward_manager_on_cpu.py
+++ b/tests/reward_loop/test_multi_reward_manager_on_cpu.py
@@ -135,6 +135,7 @@ class TestMultiVisualRewardManagerRunSingle:
         assert result["reward_extra_info"]["reward/fixed"] == pytest.approx(0.5)
         assert result["reward_extra_info"]["reward/dict_result"] == pytest.approx(1.0)
         assert result["reward_extra_info"]["reward/dict_result/detail"] == "perfect"
+        assert "reward/dict_result/score" not in result["reward_extra_info"]
         assert result["reward_extra_info"]["reward/combined"] == pytest.approx(2.0)
 
     def test_exception_contributes_zero(self):
--- a/tests/reward_loop/test_visual_reward_manager.py
+++ b/tests/reward_loop/test_visual_reward_manager.py
@@ -102,7 +102,7 @@ def test_reward_model_genrm():
 
     for idx, output in enumerate(outputs):
         print(f"GRM Response {idx}:\n{output.non_tensor_batch['genrm_response']}\n")
-        print(f"Score:\n{output.non_tensor_batch['score']}\n")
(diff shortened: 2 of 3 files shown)


zhtmike:
we should expect a `score` metric as a ensemble of different rewards, right?

cr-gao:
@zhtmike Thanks for the comment. It looks like reward/combined and critic/rewards/mean represent the ensemble of different rewards, and neither is affected. Please let me know if I got it wrong!

zhtmike:
> @zhtmike Thanks for the comment. It looks like reward/combined and critic/rewards/mean represent the ensemble of different rewards, and neither is affected. Please let me know if I got it wrong!

ok, got it

SamitHuang:
```
        for idx, output in enumerate(outputs):
            print(f"GRM Response {idx}:\n{output.non_tensor_batch['genrm_response']}\n")
>           print(f"Score:\n{output.non_tensor_batch['score']}\n")
                             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
E           KeyError: 'score'
```
please take a look at the related ci error https://github.com/verl-project/verl-omni/actions/runs/28915141780/job/85780495154

cr-gao:
Thanks @SamitHuang, I have pushed a fix. `test_reward_model_genrm` printed `output.non_tensor_batch['score']`, but it's removed in this PR. I have checked that no downstream code in the repo reads `output.non_tensor_batch['score']`, so I modified the test to read the score from `batch['rm_scores']`. End-to-end training on GPU is not affected.

zhtmike (approved):

Issue #219, opened 2026-06-30
[RFC] Improve the profiling feature of verl omni


knlnguyen1802:
### Feature request

Improve the profiling feature of verl omni

### Motivation

Currently verl omni has profiling feature inherit from verl at https://verl-omni.readthedocs.io/en/latest/perf/profiler.html

However, the profiling result only represent actor train and the result is too large 

### Your contribution

It can be improved by profiling with fewer step, and each step will run a smaller batch size, smaller denoising iteration and smaller image size

Also, the profiling must show result from rollout and reward.

cr-gao:
Hi @knlnguyen1802, I'd like to work on this. I've split it into two parts,
have part 1 validated on hardware, and want to align on part 2's boundary
before implementing it.

## Part 1 (PR ready): smaller footprint + rollout in the trace

Two findings while digging in:

- **Rollout profiling is already 95% built.** Upstream verl ships the whole
  server-side stack (`rollout.profiler` → `vLLMHttpServer.profiler_controller`
  → vLLM's built-in torch profiler), and verl-omni inherits it through
  `vLLMOmniHttpServer` — but `ray_diffusion_trainer` never calls
  `llm_server_manager.start_profile()`. Adding the same
  `curr_step_profile`-gated calls upstream uses around generation
  (`verl/trainer/ppo/ray_trainer.py`) makes the rollout servers write their
  own traces. Verified on 3×RTX 4090: the replica trace shows
  `pipeline_forward` and the denoising kernels.
- **The profiler doc's quick recipes don't run as written**: they use
  `+key=value` on keys that already exist in the composed config, which Hydra
  rejects. Fixed to plain overrides (verified via `hydra.compose`).

For the footprint, the PR makes the size knobs of the SD3.5 recipe
env-overridable (defaults unchanged) and adds a small profiling wrapper
(2 rollouts, 4 denoising steps, 256px, batch 4, 1 step). Measured effect:
traces 163 MB → 32 MB, profiled step 616 s → 70 s.

## Part 2 (proposal): reward in the trace

The reward-model servers are the same `RolloutReplica`/`vLLMHttpServer` stack,
so `start_profile` fan-out already exists there too. What's missing is a
caller, but `RewardLoopManager`/`RewardModelManager` live in upstream verl
(`verl.experimental.reward_loop`), so there are two ways to wire it:

1. **verl-omni subclass** — `OmniRewardLoopManager(RewardLoopManager)` adding
[...]

PR #252, opened 2026-07-09
[trainer, doc] feat: profile rollout servers + lightweight profiling recipe for FlowGRPO

cr-gao:
Addresses #219 (footprint + rollout coverage; reward coverage is a proposed
follow-up — see "Scope"). Not `Closes`.

## What & why

Issue #219 raises two problems with the current profiling feature:
1. traces are too large / slow to capture (full step: `rollout.n=8`, 10
   denoising steps, 384px), and
2. the profiling result only represents actor train — rollout and reward
   never show up.

This PR fixes (1) fully and (2) for the rollout half:

### Rollout servers now appear in the profiler output

Generation runs in vLLM-Omni server processes launched through the agent
loop, not in the actor worker, so the actor-side profiler can never see it.
Upstream verl already ships the full server-side profiling stack
(`rollout.profiler` → `vLLMHttpServer.profiler_controller` → vLLM's built-in
torch profiler, fan-out via `LLMServerManager.start_profile`), and verl-omni
inherits it through `vLLMOmniHttpServer(vLLMHttpServer)` — but the diffusion
trainer never called it.

- **`verl_omni/trainer/diffusion/ray_diffusion_trainer.py`** — call
  `llm_server_manager.start_profile()/stop_profile()` around the generation
  phase of profiled steps, in both `fit()` variants. This is a faithful
  transplant of the pattern upstream uses in
  `verl/trainer/ppo/ray_trainer.py` (the `curr_step_profile` gate inside
  `marked_timer("gen")`). Traces land in
  `{save_path}/agent_loop_rollout_replica_{rank}`.

### Lightweight profiling recipe (docs-only, reworked per review)

An earlier revision added a wrapper script under `examples/`; per
@knlnguyen1802's review this is now purely documentation in
`docs/perf/profiler.md`, and `examples/` is untouched. Every recipe passes
`"$@"` through to the same `diffusion_trainer` config and Hydra resolves
duplicate overrides last-wins, so the doc shows the lightweight run
(2 rollouts, 4 denoising steps, 256px, batch 4, 1 profiled step) as
overrides appended to the stock SD3.5 recipe, plus an "Adapting to other
recipes" subsection that makes it portable to any model recipe. It covers
the four couplings hardware runs surfaced:

- the SDE window must fit the reduced denoising schedule (else rollout
  post-processing fails on ragged per-sample tensors);
- the micro batch size must divide the per-GPU sample count (the engine
  chunks statically);
- the run's last step force-triggers checkpoint saving and validation —
  `save_freq=-1`/`test_freq=-1`;
- a leftover checkpoint would auto-resume past the profiled step and
  silently skip profiling — `resume_mode=disable`.

### Doc fixes (`docs/perf/profiler.md`)

- **The four existing quick recipes did not work as written**: they use
  `+key=value` appends on keys that already exist in the composed config
  (per-role configs inherit `profiler/profiler.yaml`), which Hydra rejects
  with `Could not append to config. An item is already at '...'`. Verified
  via `hydra.compose` against the real config tree; recipes now use plain
  overrides.
- Corrected the claim that the rollout shares the actor worker and that the
  actor profiler captures the full step (it does not — generation is in
  separate server processes).
- New recipe 5 (rollout servers).

## Scope

The remaining half of (2) is **reward**: `RewardLoopManager` /
`RewardModelManager` live in upstream verl (`verl.experimental.reward_loop`),
and the reward-model servers are the same `RolloutReplica` stack used here —
so the same treatment applies, but it either needs a verl-omni subclass
(`OmniRewardLoopManager`) or an upstream change. I posted the design in #219
for maintainers to pick the boundary before implementing.

At the time of writing, #219 has no assignee and no linked PR.
`gh pr list --search "219 in:body"` and `--search "profil"` return nothing
besides this PR.

## Testing

Local (no GPU needed):
- Full Hydra composition of the documented lightweight command (the
  recipe's emitted command line captured via a `python3` shim, 97
  overrides) against `verl_omni/trainer/config/diffusion_trainer.yaml`:
  composes cleanly; 21 keys asserted at their intended values, including
  rollout profiler `enable=True, discrete=True,
  save_path=./outputs/profile_sd35`.
- Same composition check demonstrates the old `+`-prefixed doc recipes fail
  and the fixed plain-override recipes pass.
- `ruff check` / `ruff format --check` pass on the trainer.

On 3×RTX 4090 (24GB; 2 actor+rollout, 1 reward pool, SD3.5-Medium LoRA OCR),
lightweight config vs recipe defaults, both with the same profiler
overrides. (Captured with this PR's earlier wrapper revision; the documented
command composes to the identical config, verified key-by-key via
`hydra.compose`.)

Results (lightweight run, single profiled step):
- Both traces written under `./outputs/profile_sd35`:
  - `e2e/prof_rank-0_<ts>.json.gz` (24 MB) — actor e2e trace;
  - `agent_loop_rollout_replica_0/<ts>_stage_0_rep_0_diffusion_<id>/trace_rank0.json.gz`
    (7.7 MB) plus `profiler_out_0.txt`, vLLM's built-in op summary table.
- The rollout trace does contain the generation compute: the summary's top
  entry is `pipeline_forward` (12.8 s CUDA over 4 calls), with
  flash-attention, cutlass GEMM and cuDNN convolution kernels below it;
  verified in the raw trace as well.
- Profiled step total 70 s (`gen` 52 s, actor trace export 28 s); the same
  footprint without the profiler runs `gen` in ~15 s, so capture+export
  overhead stays in tens of seconds. No `testing`/`save_checkpoint` timings
  appear, confirming the save/validation disable.
- Lightweight vs original footprint:

  | | lightweight | original | ratio |
  |---|---|---|---|
  | traces total (gzipped) | 32 MB | 163 MB | 5.1x |
  | actor e2e trace | 24 MB | 44 MB | 1.8x |
  | rollout replica trace | 7.7 MB | 120 MB | 15.6x |
  | `gen` (profiled) | 52 s | 578 s | 11.1x |
  | profiled step total | 70 s | 616 s | 8.8x |

  The rollout trace dominates the original footprint (`pipeline_forward`:
  45 s CUDA over 32 calls vs 12.8 s over 4), which is exactly the part the
  lightweight knobs shrink.
- Note: vllm-omni's profiler additionally tries to export an Excel workbook
  and logs a harmless warning if `openpyxl` is not installed.


examples/flowgrpo_trainer/sd35/run_sd35_medium_ocr_lora_profile.sh
```diff
@@ -0,0 +1,59 @@
+#!/usr/bin/env bash
```
knlnguyen1802: I think you can add this scripts to profiler.md docs instead of put it here.
And also we can make instruction to do it for other model

cr-gao: Thanks for the suggestions. Reworked as suggested:

- Removed `run_sd35_medium_ocr_lora_profile.sh` and reverted the recipe
  edits, so `examples/` is untouched by this PR now.
- The lightweight run lives in `docs/perf/profiler.md` as plain Hydra
  overrides appended to the stock recipe: every recipe passes `"$@"` into
  the same `diffusion_trainer` config and duplicate overrides are
  last-wins, so no script is needed.
- For other models, an "Adapting to other recipes" subsection states which
  override lines work unchanged and how to re-derive the two
  recipe-specific values (SDE window vs. denoising steps, micro batch vs.
  per-GPU sample count).

Verified with `hydra.compose` that the documented command resolves to the
exact config the wrapper produced, so the hardware numbers in the PR
description carry over unchanged. Let me know if you'd rather also see
per-model example commands spelled out in the doc.

Btw, to finish #219 I still need your call on the reward-profiling
boundary (verl-omni `OmniRewardLoopManager` subclass vs. an upstream verl
change) — details in [my comment on #219](https://github.com/verl-project/verl-omni/issues/219). Happy to start on it once you pick.

wtomin (approved):
LGTM

PR #256, opened 2026-07-10
[trainer, reward, cfg, doc] feat: profile reward-model rollout servers

### LINKED ISSUE #219: [RFC] Improve the profiling feature of verl omni

cr-gao:
Closes #219 — completes the reward half of the issue; the footprint and
rollout halves were merged in #252.

## What & why

Issue #219 asks for profiling results that also cover reward. The
reward-model servers are the same `RolloutReplica`/`vLLMHttpServer` stack as
the actor rollout servers, so the per-server profiler plumbing (vLLM's
built-in torch profiler, `--profiler_config` injection, per-replica
`start_profile` fan-out) already exists — what was missing is a config
surface and a caller. Design was posted in #219 before implementing.

### Config surface (`verl_omni/trainer/config/reward/reward.yaml`)

`reward.reward_model.rollout` previously had no `profiler` key: it silently
fell back to the `RolloutConfig` dataclass default (`None`, profiler
disabled) and rejected plain CLI overrides. The shared per-role profiler
group is now imported under `reward.reward_model.rollout.profiler`
(disabled by default — no behavior change for existing users), mirroring
how `diffusion_rollout.yaml` imports it. Its `save_path` defaults to a
`reward_model/` subdirectory because vLLM's trace directory is named
`agent_loop_rollout_replica_{rank}` for reward and actor rollout replicas
alike — without the subdirectory, replica 0 of both would write into the
same place.

### Caller (`verl_omni/reward_loop/reward_loop.py`, new)

`OmniRewardLoopManager(RewardLoopManager)` adds `start_profile`/
`stop_profile` fan-out over the reward-model rollout replicas (no-op
without a reward model). Subclassing keeps the change inside verl-omni.

### Trainer windows (`ray_diffusion_trainer.py`, both `fit()` variants)

Reward computation has two homes, so the profiling window follows it:

- **streaming** (`enable_resource_pool=True`): scoring runs inside the
  agent loop, concurrent with generation — start/stop next to the existing
  `llm_server_manager` calls in the gen block;
- **colocate** (default for most recipes): scoring runs in
  `compute_rm_score` after generation — start/stop around
  `_compute_reward_colocate` in the reward block.

The two windows are mutually exclusive by construction (the colocate branch
is skipped when `rm_scores` already streamed in, and the gen-block calls are
gated on the streaming flag).

### Docs (`docs/perf/profiler.md`)

New recipe 6 (reward-model servers), reward overrides added to the
lightweight profiling command, implementation note.

`_generated_*.yaml` regenerated via `scripts/generate_trainer_config.sh`.

At the time of writing, #219 has no assignee and no other linked PR;
`gh pr list --search "219 in:body"` / `--search "reward profil"` /
`--search "profiler"` return nothing open.

## Testing

Local (no GPU):
- Full Hydra composition of the documented lightweight command (102
  overrides captured from the recipe's real emitted command line):
  `reward.reward_model.rollout.profiler` resolves to `enable=True,
  discrete=True, save_path=./outputs/profile_sd35/reward_model`; plain
  overrides accepted.
- Replicated the exact server-side consumption
  (`omega_conf_to_dataclass(rollout)` → `RolloutConfig` with
  `ProfilerConfig`, then the `vLLMHttpServer.__init__` tool-config lookup →
  `TorchProfilerToolConfig(discrete=True, contents=[cpu, cuda])`).
- Fan-out smoke test with fake replicas: start (kwargs forwarded) / stop
  broadcast to all replicas; no-op without a reward model.
- Regression: the colocate `qwen_image` recipe composes unchanged
  (profiler present, disabled).
- `pre-commit run` passes (ruff, ruff format, mypy, generated-config
  verification, license, doc checks).

On 3×RTX 4090 (24GB; SD3.5-Medium LoRA OCR + Qwen2.5-VL-3B GenRM reward),
lightweight footprint from the profiler doc plus the recipe-6 overrides:

**Streaming run** (`enable_resource_pool=True`, the recipe default):
- Three traces, each in its own directory under the save path: actor e2e
  (24.8 MB), rollout replica (8.1 MB), reward replica
  (`reward_model/agent_loop_rollout_replica_0`, 4.5 MB).
- The reward trace contains the scoring compute: vLLM decode iterations
  (`execute_context_*_generation_*`), gemv/cutlass bf16 GEMMs,
  `_vllm_fa2_C::varlen_fwd`, `fused_add_rms_norm`,
  `reshape_and_cache_flash` — the VL reward model's forward, distinct from
  the diffusion `pipeline_forward` in the rollout trace.
- Timings confirm the window choice: `gen` 100 s vs a `reward` phase of
  ~0.1 ms — scoring streams inside the gen window, and the reward trace
  still captures it.

**Colocate run** (`enable_resource_pool=False`,
`gpu_memory_utilization=0.25`, `free_cache_engine=True`): completes
normally and writes the same three trace directories, confirming the
reward-phase window and that `start_profile` is safe on a slept reward
server (scoring wakes it inside `compute_rm_score`).

### DIFF
--- a/verl_omni/reward_loop/reward_loop.py
+++ b/verl_omni/reward_loop/reward_loop.py
@@ -1,0 +1,46 @@
+# Copyright 2026 Bytedance Ltd. and/or its affiliates
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+import asyncio
+
+from verl.experimental.reward_loop import RewardLoopManager
+
+
+class OmniRewardLoopManager(RewardLoopManager):
+    """RewardLoopManager that can start/stop the profiler on the reward-model rollout servers.
+
+    The reward-model servers are the same ``RolloutReplica`` stack as the actor rollout
+    servers, whose per-server profiler fan-out already exists (``RolloutReplica.start_profile``);
+    upstream ``RewardLoopManager`` just exposes no caller for it. The trainer invokes these
+    around the phase where the servers actually score: the generation phase when reward
+    computation streams with rollout, or ``compute_rm_score`` in colocate mode. Configured
+    via ``reward.reward_model.rollout.profiler``.
+    """
+
+    def start_profile(self, **kwargs) -> None:
+        """Start profiling on all reward-model rollout servers. No-op without a reward model."""
+        self._run_on_replicas("start_profile", **kwargs)
+
+    def stop_profile(self) -> None:
+        """Stop profiling on all reward-model rollout servers. No-op without a reward model."""
+        self._run_on_replicas("stop_profile")
+
+    def _run_on_replicas(self, method: str, **kwargs) -> None:
+        if self.reward_model_manager is None:
+            return
+        replicas = self.reward_model_manager.rollout_replicas
+
+        async def run_all():
+            await asyncio.gather(*[getattr(replica, method)(**kwargs) for replica in replicas])
+
+        asyncio.run(run_all())
--- a/docs/perf/profiler.md
+++ b/docs/perf/profiler.md
@@ -176,6 +176,29 @@ actor_rollout_ref.rollout.profiler.tool_config.torch.discrete=True
 Combine with recipe 1 to capture the actor train phase and the rollout in the
 same step.
 
+### 6. Reward-model servers
+
+When `reward.reward_model.enable=True`, the reward model runs in its own
+vLLM server processes — the same server stack as recipe 5, driven by
+`reward.reward_model.rollout.profiler`:
+
+```bash
+global_profiler.tool=torch \
+global_profiler.steps=[1] \
+reward.reward_model.rollout.profiler.enable=True \
+reward.reward_model.rollout.profiler.ranks=[0] \
+reward.reward_model.rollout.profiler.tool=torch \
+reward.reward_model.rollout.profiler.tool_config.torch.contents=[cpu,cuda] \
+reward.reward_model.rollout.profiler.tool_config.torch.discrete=True
+```
+
+The trainer starts/stops it around the phase where the servers actually
+score: the generation phase when reward computation streams with the rollout
+(`reward.reward_model.enable_resource_pool=True`), or the reward phase in
+colocate mode. Each profiled replica writes to
+`{save_path}/reward_model/agent_loop_rollout_replica_{rank}`, keeping reward
+traces apart from the actor rollout ones.
+
 ## Lightweight profiling recipe
 
 Profiling a full FlowGRPO step produces a large trace that is slow to open.
@@ -185,7 +208,8 @@ last-wins — so appending overrides to any recipe shrinks its footprint
 without editing the script. The following profiles a single lightweight step
 of the SD3.5 OCR recipe (2 rollouts instead of 8, 4 denoising steps instead
 of 10, 256px instead of 384px, train batch 4 instead of 8), capturing the
-actor train phase and the rollout servers (recipes 1 and 5 combined):
+actor train phase, the rollout servers and the reward-model servers
+(recipes 1, 5 and 6 combined):
 
 ```bash
 bash examples/flowgrpo_trainer/sd35/run_sd35_medium_ocr_lora.sh \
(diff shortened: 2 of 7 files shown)


verl_omni/reward_loop/reward_loop.py
```diff
@@ -0,0 +1,46 @@
...
+
+    def _run_on_replicas(self, method: str, **kwargs) -> None:
+        if self.reward_model_manager is None:
+            return
+        replicas = self.reward_model_manager.rollout_replicas
+
+        async def run_all():
+            await asyncio.gather(*[getattr(replica, method)(**kwargs) for replica in replicas])
+
+        asyncio.run(run_all())
```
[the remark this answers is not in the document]
cr-gao: `rollout_replicas` holds plain local `RolloutReplica` objects, not Ray actor
handles: `RewardModelManager._initialize_llm_servers` instantiates
`rollout_replica_class(...)` directly, and its own
`wake_up`/`sleep` invoke `replica.wake_up()` without `.remote()` through the
same `_run_all` pattern that `_run_on_replicas` mirrors. The Ray boundary is
one level down: `RolloutReplica.start_profile` itself calls
`server.start_profile.remote()` on its server actor handles.

This path also executed successfully on hardware: both validation runs in
the PR description (streaming and colocate) wrote reward traces, which
requires these calls to have gone through. Adding `.remote()` here would
itself raise `AttributeError` on a plain object, so I would temporarily keep it as is.

SamitHuang (commented):
Nice work. 1. Can you show the profiling results on SD3.5 LoRA OCR? 2. should we add a unit test to cover the reward profiling feature

SamitHuang:
@knlnguyen1802 PTAL

verl_omni/reward_loop/reward_loop.py
```diff
@@ -0,0 +1,46 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+import asyncio
+
+from verl.experimental.reward_loop import RewardLoopManager
+
+
+class OmniRewardLoopManager(RewardLoopManager):
```
zhtmike, line 19: suggest to  make a PR in verl directly.
We do not need to a separated reward manager only because of `start_profile` and `stop_profile`

cr-gao: Agreed. The natural home is upstream `RewardModelManager` next to `wake_up`/`sleep`, and it already has the `_run_all` helper, plus a None-guarded passthrough on `RewardLoopManager`. I'll open a PR in verl.

One constraint: verl-omni pins verl at `8a69493`, so this PR can't call the upstream method until that merges and the pin bumps. Would you prefer keeping this subclass as a temporary bridge, or holding this PR until then?

cr-gao:
> Nice work. 1. Can you show the profiling results on SD3.5 LoRA OCR? 2. should we add a unit test to cover the reward profiling feature

Thanks!

1. The numbers in the description are from exactly that recipe (SD3.5-Medium LoRA OCR + Qwen2.5-VL-3B GenRM, 3×4090), attaching perfetto screenshots of the reward-server trace below.
2. Sure, will add a CPU unit test for the start/stop fan-out in this repo if the subclass stays as a bridge, or in the upstream verl PR if we go that route.
image

verl_omni/reward_loop/reward_loop.py
zhtmike: The original design of verl-omni is to reuse verl's components if possible. So naturally the new feature of verl can be smoothly integrated into verl-omni with the minimum effort. And also we can reduce the extra maintenance burden for the hard fork (like in your PR) during the upstream synchronisations in future.

Cc @SamitHuang

zhtmike: Another problem is that it is so called OmniRewardManager, it naturally has no difference with verl's reward manager of the Omni thing.

Later do we need a DiffusionRewardManager? It is appreciated to draft a RFC to introduce new components first for this if you think it is really necessary.

cr-gao: Make sense, thanks for the clarification. I will open a PR to verl directly.

SamitHuang (approved):
LGTM

wtomin:
@cr-gao Hi, could you update the related docs:
- `docs/api/reward.rst`
- `docs/start/http_scorer.md`
- `docs/algo/async_reward.md`

Thank you for your contributions!

cr-gao:
> @cr-gao Hi, could you update the related docs:
> 
> * `docs/api/reward.rst`
> * `docs/start/http_scorer.md`
> * `docs/algo/async_reward.md`
> 
> Thank you for your contributions!

Sure!

PR #238, opened 2026-07-06
[omni, tests] feat: add general image editing interface

NancyFyong:
## Summary

This PR adds the model-agnostic pieces needed by image-editing pipelines:

- `ImageGenerationRequest` normalizes text-to-image and image-conditioned rollout payloads.
- `DiffusionI2IModelBase` defines condition preparation, concat/crop injection, metadata unwrapping, and sequence-parallel validation.
- `prepare_processor_files()` runs before both driver-side and worker-side processor loading and may return a writable prepared path for read-only model mounts.
- CPU tests cover request parsing, T2I compatibility, I2I condition dispatch, processor load order, and external adapter overrides.

The branch was rebuilt from `upstream/main` at `f81ee27`. It contains no Qwen-Image-Edit, FLUX, reward-service, training-recipe, or long-form documentation changes. The Qwen implementation is isolated on `NancyFyong/pr/qwen-image-edit` and will be opened against `main` after this PR merges.

## Compatibility

Existing T2I adapters keep the original `prepare_model_inputs` path. I2I adapters opt in by subclassing `DiffusionI2IModelBase`; missing or malformed condition latents fail before model execution. Driver and worker processor resolution use the same registry and external-library override order.

## Duplicate-work check

Checked before rebuilding:

- `gh issue view 237 --repo verl-project/verl-omni --comments`
- `gh pr list --repo verl-project/verl-omni --state open --search "237 in:body"`
- `gh pr list --repo verl-project/verl-omni --state open --search "image editing interface in:title,body"`
- `gh pr list --repo verl-project/verl-omni --state open --search "Qwen Image Edit FlowGRPO in:title,body"`

No separate open PR implements this interface.

## Tests

- `PYTHONPATH=/tmp/kilo/verl-omni-general pytest -q tests/pipelines/test_image_edit_interface_on_cpu.py tests/pipelines/test_i2i_model_base_on_cpu.py tests/pipelines/test_diffusion_model_base_on_cpu.py tests/workers/config/test_diffusion_config_on_cpu.py` - 59 passed
- `ruff check` and `ruff format --check` on all changed Python files - passed
- full pre-commit hooks on every commit - passed
- `git diff --check upstream/main..HEAD` - passed

## AI assistance

AI assistance was used while restructuring and testing this change. I reviewed the final diff and test results.

Issue #237, opened 2026-07-05
[RFC] Image-to-Image (I2I) Diffusion Model Integration for FlowGRPO


NancyFyong:
## Summary

This RFC proposes a first-class, model-agnostic **contract for integrating
image-conditioned (image-to-image / editing) diffusion models** into FlowGRPO
training in `verl-omni`. It formalizes the condition-image pathway
(`DiffusionI2IModelBase`), the rollout↔training correctness invariants that keep
the PPO importance ratio valid, and proposes an **edit-aware reward interface**
so edit runs optimize the right signal. Two models already ship on this contract
(`qwen_image_edit_flow_grpo`, `flux2_klein_flow_grpo`); this RFC documents the
contract, closes known gaps, and defines what a new I2I model must implement.

## Motivation

FlowGRPO in `verl-omni` was designed for text-to-image (T2I). Image editing adds
a **condition image** on top of the text prompt, which introduces three problems
that are currently solved ad hoc per model:

1. **Condition injection is model-specific but must reuse T2I machinery.** Each
   edit model concatenates condition latents differently (token concat, RoPE
   position ids, cross-attention), yet must reuse the shared scheduler, SDE loop,
   and log-prob recomputation without forking them.
2. **Rollout and training run in separate adapters and can silently diverge.**
   FlowGRPO recomputes the rollout trajectory's log-prob on the training side.
   Any mismatch in the effective sampling distribution — most notably
   **classifier-free guidance (CFG)** — makes the importance ratio meaningless
   and stalls reward, with no crash and no error. This class of bug is expensive
   to diagnose and has recurred.
3. **The reward is blind to the edit.** The current visual reward path scores
   only the *generated* image against the instruction text. It never sees the
   *input* image, so it cannot judge whether the edit was applied or whether
   unedited regions were preserved, and CLIP-style scorers (e.g. PickScore) do
   not reliably separate degenerate outputs from real edits.

A documented contract + closing (3) makes I2I integrations reproducible and
prevents the recurring alignment bugs.

## Guide-level explanation

An I2I model reuses the **same three-file pipeline package** as T2I:

```
verl_omni/pipelines/<model>_flow_grpo/
  ├── __init__.py                      # registers the pipeline
  ├── diffusers_training_adapter.py    # training-side (FSDP) log-prob recompute
  └── vllm_omni_rollout_adapter.py     # rollout-side (vLLM-Omni) sampling
```

The only additions over T2I are a **condition-image pathway** on both sides:

[...]

PR #235, opened 2026-07-05
[vllm_omni, rollout, trainer, reward, fsdp, algo, tests, doc] feat: add Qwen-Image-Edit-2511 FlowGRPO training support

NancyFyong:
## What does this PR do?

Adds end-to-end FlowGRPO training support for Qwen-Image-Edit-2511 / Qwen-Image-Edit-Plus style image-editing diffusion models in VeRL-Omni. This is a re-submission of #148, previously closed because the original implementation did not converge. The convergence issue has been fixed and validated with a successful training run.

### Main changes

- A Qwen-Image-Edit FlowGRPO pipeline with diffusers training and vLLM-Omni rollout adapters.
- Image-edit-specific prompt, condition-image, latent, log-prob, and timestep handling.
- Qwen-Image-Edit LoRA training and smoke-test entrypoints.
- Image-edit reward utilities, including PickScore reward (local inference mode) and vLLM-backed quality reward support.
- Trainer, rollout, FSDP, config, test, and documentation updates needed by the new path.
- Auto-patch for Hugging Face processor configs that ship without `model_type` (Qwen-Image-Edit variants).

## Checklist Before Starting

- [x] Search for similar PRs. Previous attempts #148 and #147 were closed by maintainer @zhtmike with "open the PR once it is ready". No other open PR addresses Qwen-Image-Edit FlowGRPO training.
- [x] Format the PR title as `[{modules}] {type}: {description}` (This will be checked by the CI)
  - `{modules}`: `vllm_omni, rollout, trainer, reward, fsdp, algo, tests, doc`
  - `{type}`: `feat`
  - No breaking API changes.

## Test

> Validated by a real training run on 8×H200 with Qwen-Image-Edit-2511 + LoRA (r=64) + PickScore reward. Reward is monotonically increasing over 140+ steps, confirming the pipeline converges.

### Convergence evidence (addresses #148 review feedback)

| step | reward mean |
|------|------------|
| 16   | 0.775      |
| 100  | 0.81       |
| 138  | 0.83       |

### Training curves

image

### Validation comparison (same Ghibli-style prompt across steps)

| step 20   | reward 0.78 |
image
| step 60   | 0.75 |
image
| step 100  | 0.84 |
image
| step 130  | 0.89 |
image

### Eval curves

image
### Test commands run

```bash
# End-to-end smoke test (tiny model, 1 GPU, 1 step)
NUM_GPUS=1 MODEL_PATH=tiny-random/qwen-image-edit-plus \
[...]

PR #217, opened 2026-06-29
[diffusion, tests, doc] feat: add Qwen-Image SFT example

zyd-ustc:
### What does this PR do?

This PR adds a standalone Qwen-Image SFT example for CoRT-style `t2i` / `edit` image data.

Main changes:

- Add `examples/qwen_image_sft_trainer/qwen_image_sft.py`, a PyTorch/diffusers SFT entrypoint with LoRA, FSDP, CUDA/NPU/CPU device selection, checkpoint save/resume, and CoRT-style dataset expansion.
- Add GPU and Ascend NPU LoRA launch scripts:
  - `run_qwen_image_sft_lora.sh`
  - `run_qwen_image_sft_lora_npu.sh`
- Add README documentation for supported CoRT k-turn rows, atomic rows, weights, LoRA launch, NPU launch, and checkpoints.
- Add CPU unit tests covering temporary dummy CoRT/atomic data expansion and the loss path.

This is a standalone recipe-style trainer, not a Ray/Hydra worker integration.

AI assistance was used to draft and prepare this change. The submitter has reviewed the changed files.

### Checklist Before Starting

- [x] Search for similar PRs. Query links:
  - https://github.com/verl-project/verl-omni/pulls?q=is%3Apr+is%3Aopen+Qwen-Image+SFT
  - https://github.com/verl-project/verl-omni/pulls?q=is%3Apr+is%3Aopen+qwen+image+sft
  - https://github.com/verl-project/verl-omni/issues?q=is%3Aissue+is%3Aopen+qwen+image+sft
- [x] Format the PR title as `[{modules}] {type}: {description}`.

### Test

```bash
PATH=/Users/zyd/miniconda3/bin:$PATH pre-commit run --files \
  examples/qwen_image_sft_trainer/README.md \
  examples/qwen_image_sft_trainer/qwen_image_sft.py \
  examples/qwen_image_sft_trainer/run_qwen_image_sft_lora.sh \
  examples/qwen_image_sft_trainer/run_qwen_image_sft_lora_npu.sh \
  tests/trainer/diffusion/test_qwen_image_sft_example_on_cpu.py

PR #260, opened 2026-07-13
[omni, rollout, cfg, tests] feat: add Qwen-Image-Edit FlowGRPO support

NancyFyong:
### What does this PR do?

This PR adds end-to-end FlowGRPO training support for Qwen-Image-Edit models, including `QwenImageEditPlusPipeline` checkpoints such as Qwen-Image-Edit-2511.

It is the model-specific follow-up to the general image-editing interface merged in #238, as requested in #237. It reuses `ImageGenerationRequest`, `DiffusionI2IModelBase`, and the model-scoped processor preparation hook from #238.

Main changes:

- Add a diffusers training adapter for condition-latent injection and FlowGRPO log-probability recomputation.
- Add a vLLM-Omni rollout adapter for Qwen2.5-VL prompt processing, condition-image encoding, true CFG, and replay tensor collection.
- Support fixed non-square aspect ratios such as 16:9 and 4:3 while requiring every sample in a batch to produce the same condition-latent length.
- Validate target and condition token alignment before Ulysses sequence-parallel execution.
- Prepare missing processor metadata in a writable cache without modifying read-only model snapshots.
- Add image-edit dataset preparation and a Qwen-Image-Edit LoRA recipe using the existing PickScore reward.
- Add `qwen-vl-utils` to the `vllm-omni` extra for Qwen2.5-VL image preprocessing in fresh environments.
- Add CPU tests, a tiny random checkpoint builder, and single-GPU and sequence-parallel GPU smoke tests.

This is the cleaned-up model implementation from the closed attempts #147, #148, and #235. Those PRs were not merged. The general interface was separated and merged through #238; this PR contains the Qwen-specific implementation only.

No existing T2I pipeline behavior or public API is changed.

### Checklist Before Starting

- [x] Searched for similar open PRs: https://github.com/verl-project/verl-omni/pulls?q=is%3Apr+is%3Aopen+%22Qwen-Image-Edit%22+FlowGRPO
- [x] Checked issue #237 and searched open PRs for `237 in:body`, `Qwen Image Edit FlowGRPO`, and `image editing`.
- [x] No open PR implements Qwen-Image-Edit FlowGRPO training. Related PRs #149, #214, and #217 cover different models or training paths.
- [x] The title follows `[{modules}] {type}: {description}`.
- [x] This PR does not introduce a breaking API change.

#### CPU tests

```bash
PYTHONPATH=. python -m pytest -q tests/pipelines/test_qwen_image_edit_on_cpu.py
```

Result:

```text
12 passed
```

The tests cover registration, processor preparation, condition metadata, true CFG validation, fixed non-square aspect ratios, and sequence-parallel token alignment.

#### GPU end-to-end tests

The special E2E test builds a tiny random Qwen-Image-Edit-Plus checkpoint and runs rollout, reward, replay, log-probability recomputation, and the FSDP actor update.

| Configuration | Target | Condition | Result |
| --- | --- | --- | --- |
| 1 GPU, SP=1 | 256x144 (16:9) | 512x288 (16:9) | Passed, one complete training step |
| 2 GPUs, Ulysses SP=2 | 256x144 (16:9) | 512x288 (16:9) | Passed, one complete training step |

The SP=2 run completed with:

```text
training/global_step: 1
actor/loss: 3.11e-06
actor/grad_norm: 4.24e-04
perf/total_num_images: 8
timing_s/update_actor: 2.92
Training Progress: 100% (1/1)
FlowGRPO Qwen-Image-Edit e2e test passed.
```

The combined target and condition sequence contained 4,272 tokens and was sharded across two Ulysses ranks without NCCL or all-to-all errors. For `ACTOR_SP > 1`, the smoke script selects `native` for actor attention and `TORCH_SDPA` for rollout attention; SP=1 retains the existing FA3 defaults.

#### Full-model convergence

The same Qwen-Image-Edit FlowGRPO training path was previously validated during #235 using Qwen-Image-Edit-2511, LoRA rank 64, PickScore reward, and 8 H200 GPUs.

| Step | Mean reward |
| --- | --- |
| 16 | 0.775 |
| 100 | 0.810 |
| 138 | 0.830 |

image

The full validation images and experiment discussion are available in #235.

image

some cases:

20 step reward(0.789):
image

100 step reward(0.901):
image

160 step reward(1.01):
image

#### Static checks

```bash
python -m pre_commit run --all-files --show-diff-on-failure --color=always
```

All hooks passed, including ruff, ruff-format, mypy, generated config verification, license checks, test structure validation, naming checks, and Python compilation.

### API and Usage Example

Prepare an image-edit dataset containing `train.jsonl`, `test.jsonl`, and an `images/` directory:

```bash
python examples/flowgrpo_trainer/qwen_image_edit/prepare_data.py \
    --input_dir /path/to/image_edit_dataset \
    --output_dir ./data/qwen_image_edit \
    --image_size 512
```

Launch LoRA training:

```bash
MODEL_PATH=Qwen/Qwen-Image-Edit-2511 \
TRAIN_FILES=./data/qwen_image_edit/train.parquet \
VAL_FILES=./data/qwen_image_edit/test.parquet \
bash examples/flowgrpo_trainer/qwen_image_edit/run_qwen_image_edit_lora.sh
```

Build the tiny checkpoint and run the E2E test:

```bash
python tests/special_e2e/build_qwen_image_edit_plus_tiny_random.py \
    --source-model Qwen/Qwen-Image-Edit-2511 \
    --output-dir ~/models/tiny-random/qwen-image-edit-plus

MODEL_PATH=~/models/tiny-random/qwen-image-edit-plus \
NUM_GPUS=1 \
ACTOR_SP=1 \
bash tests/special_e2e/run_flowgrpo_qwen_image_edit.sh
```

Run a non-square sequence-parallel smoke test:

```bash
CUDA_VISIBLE_DEVICES=0,1 \
MODEL_PATH=~/models/tiny-random/qwen-image-edit-plus \
NUM_GPUS=2 \
ACTOR_SP=2 \
IMAGE_HEIGHT=144 \
IMAGE_WIDTH=256 \
COND_HEIGHT=288 \
COND_WIDTH=512 \
bash tests/special_e2e/run_flowgrpo_qwen_image_edit.sh
```

#### Training adapter

`QwenImageEditPlusFlowGRPO` combines `DiffusionI2IModelBase` with the existing Qwen-Image FlowGRPO adapter. It reuses Qwen-Image timestep and log-probability computation, injects condition-image latents into transformer hidden states, forwards condition-aware shapes for 2D RoPE, and validates sequence-parallel divisibility before model execution.

#### Rollout adapter

`QwenImageEditPlusPipelineWithLogProb` extends the vLLM-Omni Qwen-Image-Edit-Plus pipeline. It receives the condition image through the existing multimodal request path, processes Qwen2.5-VL prompts containing `<image>`, encodes and packs condition latents, applies true CFG consistently, and returns all tensors and metadata required for training replay.

#### Condition-image dimensions

All condition images in a training batch must produce the same latent sequence length. When target dimensions are available, conditions must match the target aspect ratio after the Qwen VAE dimension calculation. This supports fixed non-square datasets without allowing mixed-aspect batches. When target dimensions are unavailable, the adapter retains the square-image fallback.

#### Processor preparation

Some Qwen-Image-Edit checkpoints omit `processor/config.json`. The adapter creates a deterministic writable processor cache containing the required `qwen2_vl` metadata. The original checkpoint remains unchanged, including when mounted read-only or shared between workers.

#### Tests and recipe

The PR includes focused CPU tests, a tiny random checkpoint builder, dummy parquet generation, single-GPU and Ulysses SP=2 E2E coverage, GPU smoke-test registration, dataset preparation, and a LoRA training entrypoint.

- [x] Ran the relevant CPU tests.
- [x] Ran the full pre-commit suite.
- [x] Added runnable dataset preparation and training examples.
- [x] Added CPU and end-to-end GPU tests.
- [x] Registered the Qwen-Image-Edit test in the GPU smoke-test entrypoint.
- [x] Verified single-GPU, Ulysses SP=2, and fixed non-square execution.

### AI Assistance

AI assistance was used while restructuring the earlier implementation, adding tests, diagnosing the sequence-parallel backend requirement, and drafting this PR description. I reviewed the final diff and test results and can explain and maintain the change end to end.

### DIFF
--- a/examples/flowgrpo_trainer/qwen_image_edit/prepare_data.py
+++ b/examples/flowgrpo_trainer/qwen_image_edit/prepare_data.py
@@ -31,12 +31,13 @@ SYSTEM_PROMPT = (
 
 
 def _load_image(image_path: Path, image_size: int) -> bytes:
-    image = Image.open(image_path).convert("RGB")
-    image = ImageOps.contain(image, (image_size, image_size), method=Image.Resampling.LANCZOS)
-    canvas = Image.new("RGB", (image_size, image_size), color=(255, 255, 255))
-    canvas.paste(image, ((image_size - image.width) // 2, (image_size - image.height) // 2))
-    buffer = io.BytesIO()
-    canvas.save(buffer, format="PNG")
+    with Image.open(image_path) as source:
+        image = source.convert("RGB")
+        image = ImageOps.contain(image, (image_size, image_size), method=Image.Resampling.LANCZOS)
+        canvas = Image.new("RGB", (image_size, image_size), color=(255, 255, 255))
+        canvas.paste(image, ((image_size - image.width) // 2, (image_size - image.height) // 2))
+        buffer = io.BytesIO()
+        canvas.save(buffer, format="PNG")
     return buffer.getvalue()
 
 
@@ -44,7 +45,7 @@ def _convert_split(input_dir: Path, split: str, max_samples: int, image_size: in
     jsonl_path = input_dir / f"{split}.jsonl"
     image_dir = input_dir / "images"
     rows = []
-    with open(jsonl_path) as source:
+    with open(jsonl_path, encoding="utf-8") as source:
         for index, line in enumerate(source):
             if max_samples >= 0 and index >= max_samples:
                 break
--- a/examples/flowgrpo_trainer/qwen_image_edit/prepare_data.py
+++ b/examples/flowgrpo_trainer/qwen_image_edit/prepare_data.py
@@ -1,0 +1,101 @@
+# Copyright 2026 Bytedance Ltd. and/or its affiliates
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+"""Convert an image-edit JSONL dataset to the parquet format used by training."""
+
+import argparse
+import io
+import json
+from pathlib import Path
+
+import pandas as pd
+from PIL import Image, ImageOps
+
+SYSTEM_PROMPT = (
+    "Describe the key features of the input image "
+    "(color, shape, size, texture, objects, background), then explain how the user's "
+    "text instruction should alter or modify the image. Generate a new image that meets "
+    "the user's requirements while maintaining consistency with the original input where "
+    "appropriate."
+)
+
+
+def _load_image(image_path: Path, image_size: int) -> bytes:
+    image = Image.open(image_path).convert("RGB")
+    image = ImageOps.contain(image, (image_size, image_size), method=Image.Resampling.LANCZOS)
+    canvas = Image.new("RGB", (image_size, image_size), color=(255, 255, 255))
+    canvas.paste(image, ((image_size - image.width) // 2, (image_size - image.height) // 2))
+    buffer = io.BytesIO()
+    canvas.save(buffer, format="PNG")
+    return buffer.getvalue()
+
+
+def _convert_split(input_dir: Path, split: str, max_samples: int, image_size: int) -> pd.DataFrame:
+    jsonl_path = input_dir / f"{split}.jsonl"
+    image_dir = input_dir / "images"
+    rows = []
+    with open(jsonl_path) as source:
+        for index, line in enumerate(source):
+            if max_samples >= 0 and index >= max_samples:
+                break
+            example = json.loads(line)
+            instruction = str(example["prompt"])
+            image_name = str(example["image"])
+            image_path = image_dir / image_name
+            if not image_path.is_file():
+                raise FileNotFoundError(f"condition image not found: {image_path}")
+
+            rows.append(
+                {
+                    "data_source": "image_edit",
+                    "prompt": [
+                        {"role": "system", "content": SYSTEM_PROMPT},
+                        {"role": "user", "content": f"Picture 1: <image>{instruction}"},
+                    ],
+                    "negative_prompt": [
+                        {"role": "system", "content": SYSTEM_PROMPT},
+                        {"role": "user", "content": "Picture 1: <image> "},
+                    ],
+                    "ability": "image_edit",
+                    "images": [{"bytes": _load_image(image_path, image_size)}],
+                    "reward_model": {"style": "model", "ground_truth": instruction},
+                    "extra_info": {
+                        "split": split,
+                        "index": index,
+                        "instruction": instruction,
+                        "image": image_name,
+                    },
+                }
+            )
+    return pd.DataFrame(rows)
+
+
+def main() -> None:
+    parser = argparse.ArgumentParser(description=__doc__)
+    parser.add_argument("--input_dir", type=Path, required=True)
+    parser.add_argument("--output_dir", type=Path, required=True)
+    parser.add_argument("--train_size", type=int, default=-1)
+    parser.add_argument("--val_size", type=int, default=-1)
+    parser.add_argument("--image_size", type=int, default=512)
+    args = parser.parse_args()
+
+    args.output_dir.mkdir(parents=True, exist_ok=True)
+    train = _convert_split(args.input_dir.expanduser(), "train", args.train_size, args.image_size)
+    validation = _convert_split(args.input_dir.expanduser(), "test", args.val_size, args.image_size)
+    train.to_parquet(args.output_dir / "train.parquet", row_group_size=500)
+    validation.to_parquet(args.output_dir / "test.parquet", row_group_size=500)
+    print(f"Wrote {len(train)} training and {len(validation)} validation samples to {args.output_dir}")
+
+
+if __name__ == "__main__":
+    main()
--- a/examples/flowgrpo_trainer/qwen_image_edit/run_qwen_image_edit_lora.sh
+++ b/examples/flowgrpo_trainer/qwen_image_edit/run_qwen_image_edit_lora.sh
@@ -2,7 +2,6 @@
 set -x
 
 export RAY_DEDUP_LOGS=0
-export PYTORCH_CUDA_ALLOC_CONF=expandable_segments:True
 
 # Enable reward model on GPU: Ray num_gpus=0 actors can still see CUDA devices.
 export RAY_ACCEL_ENV_VAR_OVERRIDE_ON_ZERO=0
--- a/examples/flowgrpo_trainer/qwen_image_edit/run_qwen_image_edit_lora.sh
+++ b/examples/flowgrpo_trainer/qwen_image_edit/run_qwen_image_edit_lora.sh
@@ -1,8 +1,6 @@
 # Qwen-Image-Edit-2511 LoRA RL with PickScore reward.
 set -x
 
-export RAY_DEDUP_LOGS=0
-
 # Enable reward model on GPU: Ray num_gpus=0 actors can still see CUDA devices.
 export RAY_ACCEL_ENV_VAR_OVERRIDE_ON_ZERO=0
 
--- a/examples/flowgrpo_trainer/qwen_image_edit/run_qwen_image_edit_lora.sh
+++ b/examples/flowgrpo_trainer/qwen_image_edit/run_qwen_image_edit_lora.sh
@@ -1,0 +1,99 @@
+# Qwen-Image-Edit-2511 LoRA RL with PickScore reward.
+set -x
+
+export RAY_DEDUP_LOGS=0
+export PYTORCH_CUDA_ALLOC_CONF=expandable_segments:True
+
+# Enable reward model on GPU: Ray num_gpus=0 actors can still see CUDA devices.
+export RAY_ACCEL_ENV_VAR_OVERRIDE_ON_ZERO=0
+
+model_name=${MODEL_PATH:-Qwen/Qwen-Image-Edit-2511}
+reward_function_path=${REWARD_FUNCTION_PATH:-pkg://verl_omni.utils.reward_score.pickscore_reward}
+
+NUM_GPUS_ACTOR_ROLLOUT_REWARD=${NUM_GPUS_ACTOR_ROLLOUT_REWARD:-8}
+ACTOR_SP=${ACTOR_SP:-1}
+ROLLOUT_TP=${ROLLOUT_TP:-1}
+REWARD_WORKERS=${REWARD_WORKERS:-4}
+IMAGE_RESOLUTION=${IMAGE_RESOLUTION:-512}
+MAX_PROMPT_LENGTH=${MAX_PROMPT_LENGTH:-8192}
+
+ENGINE=vllm_omni
+
+WORKSPACE=${WORKSPACE:-$(cd "$(dirname "$0")/../../.." && pwd)}
+train_path=${TRAIN_FILES:-$WORKSPACE/data/qwen_image_edit/train.parquet}
+test_path=${VAL_FILES:-$WORKSPACE/data/qwen_image_edit/test.parquet}
+
+output_dir=$WORKSPACE/outputs/qwen_image_edit_lora
+checkpoint_dir=$output_dir/checkpoints
+run_timestamp=$(date +"%Y%m%d_%H%M")
+log_file=$output_dir/logs/$run_timestamp/${NODE_RANK:-0}.log
+rollout_data_dir=$output_dir/logs/$run_timestamp/rollout_images
+val_data_dir=$output_dir/logs/$run_timestamp/val_images
+mkdir -p "$checkpoint_dir" "$(dirname "$log_file")"
+exec > >(tee -a "$log_file") 2>&1
+echo "Logging to $log_file"
+
+python3 -m verl_omni.trainer.main_diffusion \
+    data.train_files=$train_path \
+    data.val_files=$test_path \
+    data.train_batch_size=32 \
+    data.max_prompt_length=$MAX_PROMPT_LENGTH \
+    data.seed=42 \
+    actor_rollout_ref.model.algorithm=flow_grpo \
+    actor_rollout_ref.model.attn_backend=_flash_3_varlen_hub \
+    algorithm.global_std=false \
+    actor_rollout_ref.model.path=$model_name \
+    actor_rollout_ref.model.lora_rank=64 \
+    actor_rollout_ref.model.lora_alpha=128 \
+    actor_rollout_ref.model.target_modules="['to_q','to_k','to_v','to_out.0','add_q_proj','add_k_proj','add_v_proj','to_add_out','img_mlp.net.0.proj','img_mlp.net.2','txt_mlp.net.0.proj','txt_mlp.net.2']" \
+    actor_rollout_ref.actor.optim.lr=3e-4 \
+    actor_rollout_ref.actor.optim.weight_decay=0.0001 \
+    actor_rollout_ref.actor.ppo_mini_batch_size=16 \
+    actor_rollout_ref.actor.ppo_micro_batch_size_per_gpu=16 \
+    actor_rollout_ref.actor.fsdp_config.param_offload=True \
+    actor_rollout_ref.actor.fsdp_config.optimizer_offload=True \
+    actor_rollout_ref.actor.fsdp_config.model_dtype=bfloat16 \
+    actor_rollout_ref.model.lora_dtype=float32 \
+    actor_rollout_ref.actor.fsdp_config.ulysses_sequence_parallel_size=$ACTOR_SP \
+    actor_rollout_ref.actor.diffusion_loss.clip_ratio=0.0001 \
+    actor_rollout_ref.rollout.log_prob_micro_batch_size_per_gpu=32 \
+    actor_rollout_ref.rollout.seed=42 \
+    actor_rollout_ref.rollout.tensor_model_parallel_size=$ROLLOUT_TP \
+    actor_rollout_ref.rollout.name=$ENGINE \
+    actor_rollout_ref.rollout.n=16 \
+    actor_rollout_ref.rollout.agent.num_workers=$((NUM_GPUS_ACTOR_ROLLOUT_REWARD / ROLLOUT_TP)) \
+    actor_rollout_ref.rollout.load_format=safetensors \
+    actor_rollout_ref.rollout.layered_summon=True \
+    +actor_rollout_ref.rollout.engine_kwargs.vllm_omni.mm_processor_cache_gb=0 \
+    actor_rollout_ref.rollout.prompt_length=$MAX_PROMPT_LENGTH \
+    actor_rollout_ref.rollout.pipeline.num_inference_steps=12 \
+    actor_rollout_ref.rollout.pipeline.true_cfg_scale=4.0 \
+    actor_rollout_ref.rollout.pipeline.height=$IMAGE_RESOLUTION \
+    actor_rollout_ref.rollout.pipeline.width=$IMAGE_RESOLUTION \
+    actor_rollout_ref.rollout.pipeline.max_sequence_length=$MAX_PROMPT_LENGTH \
+    actor_rollout_ref.rollout.algo.noise_level=0.7 \
+    actor_rollout_ref.rollout.algo.sde_type="sde" \
+    actor_rollout_ref.rollout.algo.sde_window_size=3 \
+    actor_rollout_ref.rollout.algo.sde_window_range="[0,6]" \
+    actor_rollout_ref.rollout.val_kwargs.pipeline.num_inference_steps=40 \
+    actor_rollout_ref.rollout.val_kwargs.algo.noise_level=0.0 \
+    actor_rollout_ref.ref.log_prob_micro_batch_size_per_gpu=32 \
+    reward.num_workers=$REWARD_WORKERS \
+    reward.reward_model.enable=False \
+    reward.custom_reward_function.path=$reward_function_path \
+    reward.custom_reward_function.name=compute_score_pickscore \
+    trainer.logger='["console", "tensorboard", "wandb"]' \
+    trainer.project_name=flow_grpo \
+    trainer.experiment_name=qwen_image_edit_lora_pickscore \
+    trainer.default_local_dir=$checkpoint_dir \
+    +trainer.rollout_data_dir=$rollout_data_dir \
+    +trainer.validation_data_dir=$val_data_dir \
+    trainer.log_val_generations=8 \
+    trainer.val_before_train=False \
+    trainer.n_gpus_per_node=$NUM_GPUS_ACTOR_ROLLOUT_REWARD \
+    trainer.nnodes=1 \
+    trainer.save_freq=5 \
+    trainer.test_freq=20 \
+    trainer.total_training_steps=300 \
+    trainer.total_epochs=100 \
+    trainer.resume_mode=auto "$@"
--- a/tests/gpu_smoke/run_gpu_smoke_tests.sh
+++ b/tests/gpu_smoke/run_gpu_smoke_tests.sh
@@ -183,7 +183,7 @@ run_selected_test() {
 
 # ── Determine which tests to run ───────────────────────────────────────────────
 declare -A RUN_TEST=(
-    [0]=1 [1]=1 [2]=1 [3]=1 [4]=1 [5]=1 [6]=1 [7]=1 [8]=1 [9]=0
+    [0]=1 [1]=1 [2]=1 [3]=1 [4]=1 [5]=1 [6]=1 [7]=1 [8]=1
 )
 
 # If explicit IDs were passed on the CLI, override to run only those.
--- a/tests/gpu_smoke/run_gpu_smoke_tests.sh
+++ b/tests/gpu_smoke/run_gpu_smoke_tests.sh
@@ -183,7 +183,7 @@ run_selected_test() {
 
 # ── Determine which tests to run ───────────────────────────────────────────────
 declare -A RUN_TEST=(
-    [0]=1 [1]=1 [2]=1 [3]=1 [4]=1 [5]=1 [6]=1 [7]=1 [8]=1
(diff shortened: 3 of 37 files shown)


verl_omni/pipelines/qwen_image_edit_flow_grpo/diffusers_training_adapter.py
```diff
@@ -0,0 +1,144 @@
...
+        cls,
+        model_inputs: dict,
+        negative_model_inputs: Optional[dict],
+        condition: Optional[dict],
+    ) -> tuple[dict, Optional[dict]]:
+        if condition and condition.get("image_latents") is not None:
+            target_seq_len = model_inputs["hidden_states"].shape[1]
+            condition_seq_len = condition["image_latents"].shape[1]
+            sp_size = condition.get("sp_size")
+            if isinstance(sp_size, int) and sp_size > 1 and (target_seq_len + condition_seq_len) % sp_size:
```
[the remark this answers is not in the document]
NancyFyong: Not applied: `prepare_condition()` already uses `tu.get(micro_batch, "sp_size")`, which unwraps the training engine `NonTensorData(int)` to a Python `int`. I added a regression test in e6b0840 that passes `NonTensorData(2)` through `prepare_condition()` and verifies that non-divisible sequence lengths still raise the sequence-parallel validation error. Adding broad `hasattr(..., "item"/"data")` handling would be redundant and would weaken the existing metadata boundary.

examples/flowgrpo_trainer/qwen_image_edit/prepare_data.py
```diff
@@ -0,0 +1,101 @@
...
+    buffer = io.BytesIO()
+    canvas.save(buffer, format="PNG")
+    return buffer.getvalue()
+
+
+def _convert_split(input_dir: Path, split: str, max_samples: int, image_size: int) -> pd.DataFrame:
+    jsonl_path = input_dir / f"{split}.jsonl"
+    image_dir = input_dir / "images"
+    rows = []
+    with open(jsonl_path) as source:
```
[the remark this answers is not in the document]
NancyFyong: Adopted in e6b0840. The JSONL reader now explicitly uses `encoding="utf-8"`, matching the JSON Lines format and avoiding locale-dependent parsing.

verl_omni/pipelines/qwen_image_edit_flow_grpo/diffusers_training_adapter.py
```diff
@@ -0,0 +1,144 @@
...
+
+
+def _processor_cache_key(processor_dir: Path) -> str:
+    digest = hashlib.sha256(str(processor_dir.resolve()).encode())
+    for path in sorted(processor_dir.rglob("*")):
+        if path.is_file():
+            stat = path.stat()
+            digest.update(str(path.relative_to(processor_dir)).encode())
+            digest.update(f"{stat.st_size}:{stat.st_mtime_ns}".encode())
+    return digest.hexdigest()[:16]
```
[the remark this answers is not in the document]
NancyFyong: Not applied as suggested. The cache is intentionally keyed by the resolved local source path as well as stat metadata; hashing contents while retaining that path would not provide cross-checkout or cross-node deduplication. The default processor also includes approximately 16 MB of tokenizer assets, so reading every file on every worker/cache lookup is not free. A shared content-addressed cache would need a separate design that removes the path identity and uses streaming hashes. The current local-cache semantics retain the lightweight stat key.

examples/flowgrpo_trainer/qwen_image_edit/prepare_data.py
```diff
@@ -0,0 +1,101 @@
...
+
+
+def _load_image(image_path: Path, image_size: int) -> bytes:
+    image = Image.open(image_path).convert("RGB")
+    image = ImageOps.contain(image, (image_size, image_size), method=Image.Resampling.LANCZOS)
+    canvas = Image.new("RGB", (image_size, image_size), color=(255, 255, 255))
+    canvas.paste(image, ((image_size - image.width) // 2, (image_size - image.height) // 2))
+    buffer = io.BytesIO()
+    canvas.save(buffer, format="PNG")
+    return buffer.getvalue()
```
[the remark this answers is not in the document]
NancyFyong: Adopted in e6b0840. `_load_image()` now uses `with Image.open(...)` so source files are deterministically closed while preserving the conversion, resize, padding, and PNG serialization behavior.

verl_omni/pipelines/qwen_image_edit_flow_grpo/diffusers_training_adapter.py
```diff
@@ -0,0 +1,144 @@
...
+        if (prepared_dir / "config.json").is_file():
+            return str(prepared_dir)
+
+        cache_root.mkdir(parents=True, exist_ok=True)
+        temporary_root = Path(tempfile.mkdtemp(prefix=f".{cache_key}-", dir=cache_root))
+        temporary_processor = temporary_root / "processor"
+        try:
+            shutil.copytree(processor_dir, temporary_processor)
+            with open(temporary_processor / "config.json", "w") as config_file:
+                json.dump({"model_type": "qwen2_vl"}, config_file)
```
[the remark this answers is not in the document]
NancyFyong: Adopted in e6b0840. After `copytree`, the cache copy now preserves its existing mode while adding owner-write permission before creating `config.json`. The read-only-source test now asserts that the resulting cache directory has owner-write permission, so it catches this regression even when tests run as root.

verl_omni/pipelines/qwen_image_flow_grpo/common.py
```diff
@@ -36,7 +36,11 @@ def apply_true_cfg(
 ) -> torch.Tensor:
     comb_pred = negative_noise_pred + true_cfg_scale * (noise_pred - negative_noise_pred)
     cond_norm = torch.norm(noise_pred, dim=-1, keepdim=True)
-    noise_norm = torch.norm(comb_pred, dim=-1, keepdim=True)
+    # Clamp the denominator: when comb_pred collapses to near-zero under bf16
+    # + few-inference-steps the unclamped form produces inf/NaN that downstream
+    # get cached as the rollout's old_log_prob and silently NaN out the FSDP
+    # grad. Harmless for T2I (noise_norm is never near-zero there).
+    noise_norm = torch.norm(comb_pred, dim=-1, keepdim=True).clamp_min(1e-6)
```
zhtmike: why `comb_pred` will be 0?

verl_omni/pipelines/qwen_image_edit_flow_grpo/diffusers_training_adapter.py
```diff
@@ -0,0 +1,146 @@
...
+from tensordict import TensorDict
+from verl.utils import tensordict_utils as tu
+
+from verl_omni.pipelines.model_base import DiffusionI2IModelBase, DiffusionModelBase
+from verl_omni.pipelines.qwen_image_flow_grpo.diffusers_training_adapter import QwenImage
+
+__all__ = ["QwenImageEditPlusFlowGRPO"]
+
+
+def _processor_cache_key(processor_dir: Path) -> str:
```
zhtmike: why we need this?

verl_omni/pipelines/qwen_image_edit_flow_grpo/diffusers_training_adapter.py
```diff
@@ -0,0 +1,146 @@
...
+            temporary_processor.chmod(stat.S_IMODE(temporary_processor.stat().st_mode) | stat.S_IWUSR)
+            with open(temporary_processor / "config.json", "w") as config_file:
+                json.dump({"model_type": "qwen2_vl"}, config_file)
+            try:
+                temporary_processor.rename(prepared_dir)
+            except OSError:
+                if not (prepared_dir / "config.json").is_file():
+                    raise
+        finally:
+            shutil.rmtree(temporary_root, ignore_errors=True)
```
zhtmike: why do we need this, like copy tree, rmtree, etc?

verl_omni/pipelines/qwen_image_edit_flow_grpo/diffusers_training_adapter.py
```diff
@@ -0,0 +1,146 @@
...
+        """
+        processor_dir = Path(model_path) / "processor"
+        if not processor_dir.is_dir():
+            raise FileNotFoundError(f"Qwen-Image-Edit processor directory not found: {processor_dir}")
+        if (processor_dir / "config.json").is_file():
+            return str(processor_dir)
+
+        cache_root = Path(
+            os.environ.get(
+                "VERL_OMNI_PROCESSOR_CACHE",
```
zhtmike: `VERL_OMNI_PROCESSOR_CACHE` is coming from nowhere?

verl_omni/pipelines/qwen_image_edit_flow_grpo/vllm_omni_rollout_adapter.py
```diff
@@ -0,0 +1,625 @@
...
+    # Every sample in a training batch must resolve to the SAME condition VAE
+    # size; otherwise the cross-sample concatenation of ``condition_image_latents``
+    # in the agent loop (and the FSDP micro-batch) fails on mismatched sequence
+    # lengths. When the pipeline target height/width are known we anchor every
+    # sample to their aspect ratio through the same upstream
+    # ``calculate_dimensions`` that produced ``vae_image_sizes``, so any fixed
+    # aspect ratio (e.g. 16:9, 4:3) is allowed as long as every source image
+    # shares it. A square target reduces to the historical 1024x1024
+    # requirement. When the target size is unknown we fall back to requiring
+    # square conditions.
```
zhtmike: clean it

verl_omni/pipelines/qwen_image_edit_flow_grpo/vllm_omni_rollout_adapter.py
```diff
@@ -0,0 +1,625 @@
...
+
+@VllmOmniPipelineBase.register("QwenImageEditPlusPipeline", algorithm="flow_grpo")
+class QwenImageEditPlusPipelineWithLogProb(QwenImageTokenIdPromptMixin, QwenImageEditPlusPipeline):
+    """Rollout pipeline for Qwen-Image-Edit-Plus that captures per-step log-probabilities.
+
+    Extends :class:`~vllm_omni.diffusion.models.qwen_image.pipeline_qwen_image_edit_plus.QwenImageEditPlusPipeline`
+    with a custom SDE-based scheduler and additional output fields required
+    for RL training (e.g. FlowGRPO).  In addition to the final generated image
+    the pipeline returns all intermediate latents, their log-probabilities,
+    the corresponding timesteps, and the condition image latents.
```
zhtmike: may clean

verl_omni/pipelines/qwen_image_edit_flow_grpo/vllm_omni_rollout_adapter.py
```diff
@@ -0,0 +1,625 @@
...
+        word embedding, prompt features become garbage, and the diffusion
+        transformer denoises noise into noise.  With them, the encoder fuses
+        real image features in.
+
+        IMPORTANT: ``condition_images`` here MUST be the raw PIL images
+        forwarded by the agent loop (i.e. ``multi_modal_data["image"]``),
+        NOT the resized list that vllm-omni's ``pre_process_func`` writes to
+        ``additional_information["condition_images"]``.  The latter is already
+        resized to ``CONDITION_IMAGE_SIZE`` (384x384) and would produce a
+        mismatched patch grid.
```
zhtmike: clean it. If it is important, just use a assert to guard it

verl_omni/pipelines/qwen_image_edit_flow_grpo/vllm_omni_rollout_adapter.py
```diff
@@ -0,0 +1,625 @@
...
+        """Encode pre-tokenized prompt IDs with optional condition images.
+
+        The agent loop expanded every ``<image>`` placeholder in the prompt
+        text into a long run of ``<|vision_start|><|image_pad|>...<|vision_end|>``
+        tokens by calling ``processor(text=..., images=raw_images, ...)`` on
+        the same raw PIL list.  Without ``pixel_values`` / ``image_grid_thw``
+        the Qwen2.5-VL text_encoder treats every ``<|image_pad|>`` as an empty
+        word embedding, prompt features become garbage, and the diffusion
+        transformer denoises noise into noise.  With them, the encoder fuses
+        real image features in.
```
zhtmike: make it short and easy to read

verl_omni/pipelines/qwen_image_edit_flow_grpo/vllm_omni_rollout_adapter.py
```diff
@@ -0,0 +1,625 @@
...
+            prompt_embeds.dtype,
+            self.device,
+            generator,
+            latents,
+        )
+
+        # Use pre-encoded image_latents from the request if provided, otherwise
+        # fall back to what prepare_latents returned (None when images=None).
+        # Rename to condition_image_latents to distinguish from the MFU FLOPs
+        # counter's "image_latents" key (which denotes the denoised latent).
```
zhtmike: do we need fallback?

verl_omni/pipelines/qwen_image_edit_flow_grpo/vllm_omni_rollout_adapter.py
```diff
@@ -0,0 +1,625 @@
...
+                # ``vae_image_sizes`` is populated by the upstream
+                # ``pre_process_func`` whenever ``vae_images`` are encoded.
+                # Without it we cannot recover the condition image's aspect
+                # ratio from the packed sequence length alone (sqrt(seq_len)
+                # only equals the side for perfect squares and otherwise
+                # silently produces wrong RoPE positions). Fail closed for
+                # both ``None`` and the empty-list case. An empty list
+                # would still pass the old ``is None`` check and produce
+                # an ``img_shapes`` entry with no condition regions, which
+                # silently corrupts the transformer's RoPE positions.
```
zhtmike: clean

examples/flowgrpo_trainer/qwen_image_edit/run_qwen_image_edit_lora.sh
```diff
@@ -0,0 +1,99 @@
+# Qwen-Image-Edit-2511 LoRA RL with PickScore reward.
+set -x
+
+export RAY_DEDUP_LOGS=0
+export PYTORCH_CUDA_ALLOC_CONF=expandable_segments:True
```
zhtmike: why do we need this?

tests/gpu_smoke/run_gpu_smoke_tests.sh
```diff
@@ -264,6 +264,11 @@ run_selected_test 8 "Qwen3-Omni Thinker GSPO LoRA e2e" \
     env CUDA_VISIBLE_DEVICES="0,1" NUM_GPUS=2 \
     bash tests/special_e2e/run_gspo_qwen3_omni_thinker_lora_smoke.sh
 
+# Test 9 is optional because its tiny checkpoint is not part of the CI image.
+run_selected_test 9 "Qwen-Image-Edit I2I FlowGRPO e2e" \
+    env CUDA_VISIBLE_DEVICES="${CUDA_DEVICE_LIST}" NUM_GPUS="${NUM_GPUS}" \
+    bash tests/special_e2e/run_flowgrpo_qwen_image_edit.sh
+
```
zhtmike: consider drop this for now.

tests/special_e2e/build_qwen_image_edit_plus_tiny_random.py
```diff
@@ -0,0 +1,310 @@
...
+
+No tiny-random Qwen-Image-Edit checkpoint is published on the Hub, so the
+FlowGRPO image-edit smoke test builds one locally. Following
+``build_sd3_tiny_random.py`` and ``build_qwen3_omni_tiny_random.py``, the
+diffusion components (transformer, VAE, Qwen2.5-VL text encoder) are constructed
+in memory with random weights and shrunk to a couple of layers. The tokenizer,
+Qwen2VLProcessor and scheduler are re-serialized from a cached source checkpoint
+(``Qwen/Qwen-Image-Edit-2511`` by default) because the ``<image>`` placeholder
+expansion depends on the exact special-token ids and image-processor geometry;
+tokenization quality is irrelevant for a plumbing smoke test.
```
zhtmike: make it clean and short

tests/special_e2e/run_flowgrpo_qwen_image_edit.sh
```diff
@@ -0,0 +1,122 @@
...
+# Ulysses sequence parallelism needs the SP-capable native attention on both
+# actor and rollout; FA3 (_flash_3_varlen_hub) does not support SP. Keep FA3 as
+# the default for SP=1 (backward compatible).
+if [[ "${ACTOR_SP}" -gt 1 ]]; then
+    ATTN_BACKEND=${ATTN_BACKEND:-native}
+    ROLLOUT_ATTN_BACKEND=${ROLLOUT_ATTN_BACKEND:-TORCH_SDPA}
+else
+    ATTN_BACKEND=${ATTN_BACKEND:-_flash_3_varlen_hub}
+    ROLLOUT_ATTN_BACKEND=${ROLLOUT_ATTN_BACKEND:-FLASH_ATTN}
+fi
```
zhtmike, line 35: SP is verified?

verl_omni/pipelines/qwen_image_edit_flow_grpo/vllm_omni_rollout_adapter.py
```diff
@@ -0,0 +1,625 @@
...
+        # Parse the condition images via the shared ImageGenerationRequest interface.
+        request_payload = custom_prompt
+        if (
+            isinstance(custom_prompt, dict)
+            and prompt_embeds is not None
+            and custom_prompt.get("prompt") is None
+            and custom_prompt.get("prompt_token_ids") is None
+        ):
+            request_payload = {**custom_prompt, "prompt": ""}
+        gen_request = ImageGenerationRequest.from_request_payload(request_payload) if request_payload else None
```
zhtmike, line 360: should we use ImageGenerationRequest for the current T2I pipelines? If yes, can we do it in separate PRs?

verl_omni/pipelines/qwen_image_flow_grpo/common.py
NancyFyong: With few inference steps + bf16, noise_pred and negative_noise_pred can be close enough that the CFG extrapolation neg + scale * (pos - neg) cancels comb_pred to ~0, making noise_norm 0 → inf/NaN that silently NaNs the FSDP grad via old_log_prob. The clamp_min(1e-6) guards that. Perhaps we can prevent this from happening？

verl_omni/pipelines/qwen_image_edit_flow_grpo/diffusers_training_adapter.py
NancyFyong: _processor_cache_key avoids re-copying the ~16 MB processor directory on every worker when config.json is missing. The key is based on the resolved path + file stat metadata so workers sharing the same source checkpoint hit the same cache entry.

verl_omni/pipelines/qwen_image_edit_flow_grpo/diffusers_training_adapter.py
NancyFyong: The copytree → write config.json → rename pattern is an atomic publish: each worker builds in a private temp dir, then atomically renames to the shared cache location so concurrent workers never see a half-written copy. The rmtree in finally cleans up if the rename failed (e.g. another worker won the race).

tests/special_e2e/run_flowgrpo_qwen_image_edit.sh
NancyFyong: Yes, SP>1 has been verified locally with ACTOR_SP=2 — the native attention backend path works correctly for Qwen-Image-Edit e2e gpu smoke test. However, I haven't officially used and verified it in training yet.

verl_omni/pipelines/qwen_image_edit_flow_grpo/diffusers_training_adapter.py
zhtmike: possible to move this to driver side?

verl_omni/pipelines/qwen_image_edit_flow_grpo/diffusers_training_adapter.py
zhtmike: possible to move this to driver side, so we don't need copy or cache

verl_omni/pipelines/qwen_image_flow_grpo/common.py
zhtmike: I am not sure if it will change the behavior for all T2I scripts. Possibly to guard this for T2I only.
cc @SamitHuang

verl_omni/pipelines/qwen_image_edit_flow_grpo/diffusers_training_adapter.py
NancyFyong: Removed. The processor config is now created once by the driver directly in the shared module directory, so VERL_OMNI_PROCESSOR_CACHE and all copy/cache logic are no longer needed.

verl_omni/pipelines/qwen_image_edit_flow_grpo/vllm_omni_rollout_adapter.py
NancyFyong: Yes. This PR only uses ImageGenerationRequest for the new image-edit pipeline. Minor pipelines can be handled in a separate PR to keep this change focused.But are the changes to other T2i models slightly more significant?

NancyFyong:
@zhtmike all done，In addition to claim 1e-6，Perhaps we can remove this first, since I only used it for guarding.

verl_omni/pipelines/qwen_image_edit_flow_grpo/vllm_omni_rollout_adapter.py
zhtmike: Sure, we may keep a note for this.

verl_omni/workers/config/diffusion/model.py
```diff
@@ -158,14 +158,6 @@ def __post_init__(self):
             )
             processor_path = os.path.join(self.local_path, "processor")
             if os.path.exists(processor_path):
-                from verl_omni.pipelines.model_base import DiffusionModelBase
-
-                prepared_processor_path = DiffusionModelBase.get_class_by_name(
-                    self.architecture,
-                    self.algorithm,
-                ).prepare_processor_files(self.local_path)
```
zhtmike, line 166: I didn't see where is prepare_processor_files being called?

zhtmike:
> @zhtmike all done，In addition to claim 1e-6，Perhaps we can remove this first, since I only used it for guarding.

yeah, pls check if removing it is safe

zhtmike:
there is no epsilon in official implementation https://github.com/huggingface/diffusers/blob/main/src/diffusers/pipelines/qwenimage/pipeline_qwenimage_edit.py#L799-L801

I think it should be fine if the trainer is properly settled

verl_omni/workers/config/diffusion/model.py
NancyFyong: It is called once on the driver in verl_omni/trainer/main_diffusion.py before trainer.init().

Qwen hook creates <local_path>/processor/config.json in the shared model directory, and workers subsequently load that same processor path directly. I removed only the redundant worker-side init.

zhtmike: oh right

NancyFyong:
> there is no epsilon in official implementation https://github.com/huggingface/diffusers/blob/main/src/diffusers/pipelines/qwenimage/pipeline_qwenimage_edit.py#L799-L801
> 
> I think it should be fine if the trainer is properly settled

sure，i dropped the clamp so the shared helper is byte-for-byte identical to upstream/main.

To confirm this doesn't silently change the edit pipeline, I instrumented `apply_true_cfg` to compute the clamped and unclamped outputs on the **same** inputs across a full Qwen-Image-Edit e2e rollout (`bf16`, `num_inference_steps=4`, `true_cfg_scale=4.0`):

- min `||comb_pred||` over the run: ~2.9–3.4 (guard threshold is `1e-6`)
- rows with `||comb_pred|| < 1e-6`: `0 / 1024` on every call
- max `|clamped - unclamped|` output diff: `0.0` on every call
- the training step completes with finite `actor/loss` and `actor/grad_norm`

So the clamp never fires in this regime, which is exactly why removing it changes nothing.

tests/gpu_smoke/run_gpu_smoke_tests.sh
```diff
@@ -183,7 +183,7 @@ run_selected_test() {
 
 # ── Determine which tests to run ───────────────────────────────────────────────
 declare -A RUN_TEST=(
-    [0]=1 [1]=1 [2]=1 [3]=1 [4]=1 [5]=1 [6]=1 [7]=1 [8]=1
+    [0]=1 [1]=1 [2]=1 [3]=1 [4]=1 [5]=1 [6]=1 [7]=1 [8]=1 [9]=0
```
zhtmike: revert [9]=0

verl_omni/pipelines/qwen_image_edit_flow_grpo/diffusers_training_adapter.py
```diff
@@ -0,0 +1,101 @@
...
+    @classmethod
+    def prepare_processor_files(cls, model_path: str) -> Optional[str]:
+        """Create the processor config omitted by Qwen-Image-Edit checkpoints."""
+        processor_dir = Path(model_path) / "processor"
+        if not processor_dir.is_dir():
+            raise FileNotFoundError(f"Qwen-Image-Edit processor directory not found: {processor_dir}")
+        config_path = processor_dir / "config.json"
+        if not config_path.is_file():
+            config_path.write_text(json.dumps({"model_type": "qwen2_vl"}), encoding="utf-8")
+        return str(processor_dir)
```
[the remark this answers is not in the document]
zhtmike: just ignore this

tests/gpu_smoke/run_gpu_smoke_tests.sh
NancyFyong: done

zhtmike:
Looks good, pls add a DOC PR how of how to train with qwen-image edit and also how to add new image edit model, thanks :)

zhtmike (approved):
Thanks for the PR!

PR #253, opened 2026-07-10
[1/N][omni] feat: Omni model base refactor

zhtmike:
### What does this PR do?

Phase 1 of [#232]: core interfaces + Qwen3-Omni adapters for formal omni-model support.
- `OmniModelBase` — pluggable training adapter base class, registered by `(architecture, stage)`.
- `OmniRolloutPipelineBase` — registry for vLLM-Omni pipeline topologies, keyed by `model_type`.
- `OmniModelConfig` — dataclass for omni model training, mirrors `HFModelConfig` + omni-specific fields.
- `omni_trainer.yaml` — minimal Hydra config inheriting verl's `ppo_trainer`.
- `Qwen3OmniThinkerAdapter` — strips talker/code2wav, redirects forward to thinker, configures processor and tokenizer.
- `Qwen3OmniRolloutAdapter` — generates per-stage model-topology defaults for thinker_only, thinker_talker, full modes.
- Regression tests for all monkey-patches dropped from the adapter.
- `tie_word_embeddings=False` patch dropped — already the default in transformers >= 5.0.
- `_no_split_modules` override dropped — thinker subclass already correct in v5.
- MoE expert unfusing hook dropped — PEFT's v4→v5 conversion handles fused `gate_up_proj` natively.
- `AutoModelForCausalLM.register()` dropped — deferred to Phase 2, will route through `AutoModelForMultimodalLM`.

Phase 2: `main_omni.py` thin Hydra wrapper, `OmniPPOTrainerSync` V1 trainer, `OmniAgentLoopWorkerTQ` V1 agent loop.
Phase 2: update GSPO recipe to use `main_omni` instead of `main_ppo` + `external_lib`.
### Checklist Before Starting

- [ ] Search for similar PRs. Paste at least one query link here: ...
- [ ] Format the PR title as `[{modules}] {type}: {description}` (This will be checked by the CI)
  - `{modules}` include `fsdp`, `vllm_omni`, `rollout`, `trainer`, `ci`, `training_utils`, `recipe`, `ray`, `worker`, `single_controller`, `misc`, `perf`, `model`, `algo`, `env`, `tool`, `ckpt`, `doc`, `data`, `cfg`, `reward`, `diffusion`, `omni`, `tests`, `docker`
  - If this PR involves multiple modules, separate them with `,` like `[diffusion, doc]`
  - `{type}` is in `feat`, `fix`, `refactor`, `chore`, `test`
[...]

PR #281, opened 2026-07-17
[doc, cfg] chore: document reward-server profiling in reward docs

cr-gao:
Doc follow-up to #256.

## What

- `docs/api/reward.rst`: add `OmniRewardLoopManager` to the API reference; the intro previously pointed at verl's `RewardLoopManager`, which #256 replaced.
- `docs/algo/async_reward.md`: new "Profiling the reward servers" section — under async reward the profiling window is the generation phase (near-zero `reward` timer, populated reward trace); recipe details stay in the profiler guide.
- `docs/start/http_scorer.md`: note that HTTP scorers run outside VeRL-Omni, so the reward-server profiler does not cover them.

Also regenerates `_generated_omni_trainer.yaml`: it was added in #253 in parallel with #256's `reward.yaml` change, so on main it is missing the `reward_model.rollout.profiler` block and fails the autogen pre-commit check for any PR.

## Testing

`pre-commit run` on the changed files passes (docs checks, autogen-config verification, license, compile).

### DIFF
--- a/docs/algo/async_reward.md
+++ b/docs/algo/async_reward.md
@@ -1,7 +1,7 @@
 (async_reward)=
 # Async Reward for Diffusion Training
 
-Last updated: 06/09/2026
+Last updated: 07/17/2026
 
 Async reward lets VeRL-Omni score completed rollout samples through reward-loop
 workers while other samples are still being generated. It is useful when reward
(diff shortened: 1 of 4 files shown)


AndyZhou952 (approved):
LGTM

SamitHuang (approved):
LGTM

PR #290, opened 2026-07-21
[1/N][omni, rollout] feat: support Qwen3-Omni text-only full-model fully-async rollout

hbhflw2000:
## Summary

Add the verl-omni integration required to run Qwen3-Omni Thinker full-model fully-async RL for the text-only path: text prompts in, text-token rollout out, with a Megatron actor and standalone vLLM-Omni AR rollout replicas.

Here, **full-model** refers to full actor-weight synchronization rather than LoRA. It does not mean that all Qwen3-Omni modalities are supported by this PR.

## Scope

- Wire Qwen3-Omni Thinker into the vLLM-Omni AR async-server path for text-token rollout.
- Enforce node-local rollout replicas with `data_parallel_size=1`; scale through verl-owned outer replicas.
- Generate and validate actor-local stage placement before engine startup.
- Reuse verl's bounded bucket IPC weight-update path for full-model synchronization.
- Add request abort, cache reset, and generation resume lifecycle hooks.
- Preserve optional multimodal payload fields at the adapter boundary for future integration without claiming end-to-end multimodal support.
- Allow AR-only jobs to skip diffusion-only component registration.

## Validation

- Targeted fully-async rollout tests: **18 passed**.
- Targeted pre-commit hooks: **all passed**.
- Text-only GSM8K 32-GPU full-model fully-async RL gate:
  - 4 nodes, 16 training GPUs and 16 rollout GPUs.
  - Four node-local TP4 rollout replicas.
  - **2/2 training steps completed**, `python_rc=0`.
  - Weight versions advanced `0 -> 1 -> 2` on all rollout replicas.
  - Rollout/actor Pearson correlation: `0.98896`.
  - Rollout KL: `0.00737`; actor grad norm: `0.2747`.
  - No aborted responses, processed stale trajectories, OOMs, or distributed failures.
- Rebased onto latest `main` (`34445be`); the rebased patch is range-diff identical to the E2E-validated version.

## Related component PRs

The end-to-end stack composes repository-owned lower-layer changes:

- NVIDIA NeMo Megatron-Bridge [#4988](https://github.com/NVIDIA-NeMo/Megatron-Bridge/pull/4988): Qwen3-Omni Megatron/HF parity.
- vLLM-Omni [#5273](https://github.com/vllm-project/vllm-omni/pull/5273): AR sampled-token logprob handling.
- vLLM-Omni [#5277](https://github.com/vllm-project/vllm-omni/pull/5277): multi-node lifecycle and deployment-scoped port ownership.
- verl [#7106](https://github.com/verl-project/verl/pull/7106): optional R2 router replay correctness for THD-packed Megatron batches.

The default fully-async path does not require R2; #7106 is only used when validating the optional R2 configuration.

## Non-goals

- Image-, video-, or audio-conditioned rollout.
- Audio or vision generation rollout.
- A single rollout replica spanning multiple nodes.
- Inner vLLM-Omni data parallelism greater than one.
- Fully-async LoRA weight synchronization.

Multimodal understanding and audio/vision rollout are being developed and validated separately and are intentionally outside this PR.


SamitHuang:
@AndyZhou952  PTAL

hbhflw2000:
Updated on 2026-07-26:

- Refreshed this stacked PR onto the latest head of dependency [#258](https://github.com/verl-project/verl-omni/pull/258), commit `b5317be2`.
- The two dependency commits added after the full-model run only remove a redundant LoRA hook and adjust LoRA recipe PPO steps; the validated full-model path is unchanged.
- Fully-async rollout targeted tests — **18 passed**.
- Targeted pre-commit hooks — **all passed**.
- 32-GPU full-model async gate — **2/2 steps completed**, `python_rc=0`, rollout/actor Pearson correlation `0.98955`, rollout KL `0.00844`, grad norm `0.2923`, no aborted/stale samples, and weight versions advanced `0 -> 1 -> 2` across all rollout ranks.

This PR remains stacked on #258 and should stay draft until #258 is rebased/merged; the current GitHub conflict status comes from that open dependency chain rather than this integration commit.

SamitHuang:
@hbhflw2000 the results seem good. pls resolve the conflicts

zhtmike:
can you do a rebase with latest main first? Hard to review and git log is not clear, thx

hbhflw2000:
@zhtmike Done, thanks. I rebased the PR onto the latest main (34445be) and replaced the previous stacked/merge history with a single DCO-signed commit (3431222). GitHub now reports the PR as mergeable.The rebased patch is range-diff identical to the E2E-validated version. After the rebase, all 18 targeted tests and all targeted pre-commit hooks passed.

zhtmike:
We should review it now or after you merge your upstream fix? the current code does not work for current upstream dependencies, right?

zhtmike:
And can you post the reward curve and other relevant metrics?

verl_omni/workers/rollout/vllm_rollout/vllm_omni_async_server.py
```diff
@@ -60,8 +63,21 @@
...
+    return value
+
+
+def _select_ar_model_config_type(model_config) -> type[HFModelConfig] | type[OmniModelConfig]:
+    """Keep legacy language-model backends compatible with the omni rollout server."""
+    target = model_config.get("_target_", "") if hasattr(model_config, "get") else ""
+    model_type = model_config.get("model_type", "language_model") if hasattr(model_config, "get") else "language_model"
+    if str(target).endswith("OmniModelConfig") or model_type == "omni_model":
+        return OmniModelConfig
+    return HFModelConfig
```
zhtmike: instead of doing this, pls using OmniModelConfig directly. You may modify OmniModelConfig or inherited it based on your wish. Pls Don't fallback to HFModelConfig

verl_omni/workers/rollout/vllm_rollout/vllm_omni_async_server.py
zhtmike, line 1: you may also need a yaml for megatron trainer for this

hbhflw2000:
Hi @zhtmike ,
Sorry for replying so late, yes, better when upstream fixes merged. I'm changing to OmniModelConfig and luanching a new e2e 50step test to show metrics. I'll post reward curves and other metrics soon.

examples/gspo_trainer/qwen3_omni/config/qwen3_omni_thinker_gspo_megatron_fully_async.yaml
zhtmike, line 1: is it better to create a default yaml, like https://github.com/verl-project/verl/blob/main/verl/trainer/config/_generated_ppo_megatron_trainer.yaml. Using a custom yaml create a lot of burden

zhtmike: Using a default yaml is beneficial, we can just use a single script for the training task, and it is much clearer. you can see the practice for v1 script in  https://github.com/verl-project/verl-omni/blob/main/examples/gspo_trainer/README.md#run-training

zhtmike:
> Hi @zhtmike , Sorry for replying so late, yes, better when upstream fixes merged. I'm changing to OmniModelConfig and luanching a new e2e 50step test to show metrics. I'll post reward curves and other metrics soon.

got it

hbhflw2000:
here are 50 step curves:
Qwen3-Omni fully-async RL 50-step core metrics

zhtmike:
1. can you show  `rollout_corr/training_log_ppl` and `rollout_corr/rollout_log_ppl`, to see if the model is actually learning?
2. can you show the validation reward curve?

you can refer to my exp for fsdp training. https://wandb.ai/mikecheung/gspo/runs/j5mro1tn

hbhflw2000:
D-Chat20260730145000
@zhtmike Thx for the information, I've checked the fsdp recipe and aligned configs. Here are the updated metrics, I've changed mini batch size from 16 to 32, which may be possible reason for instability of reward curve.

zhtmike:
> D-Chat20260730145000 @zhtmike Thx for the information, I've checked the fsdp recipe and aligned configs. Here are the updated metrics, I've changed mini batch size from 16 to 32, which may be possible reason for instability of reward curve.

seems good! maybe train a bit longer.

hbhflw2000:
D-Chat20260731102238
Hi @zhtmike, thanks again for the detailed review and for sharing the FSDP LoRA reference. Here is a consolidated update covering the review comments and the longer validation run.

### Review comments addressed

- Removed the `HFModelConfig` fallback and now use `OmniModelConfig` directly.
- Replaced the example-specific custom YAML with repository-level defaults:
  - `fully_async_omni_megatron_trainer.yaml`
  - `_generated_fully_async_omni_megatron_trainer.yaml`
- The training script now composes the default trainer config and only overrides recipe-specific values, following the existing verl configuration pattern.
- The PR has been rebased and cleaned up, and GitHub currently reports it as cleanly mergeable.

### Dependency status

- vLLM-Omni [#5273](https://github.com/vllm-project/vllm-omni/pull/5273), the sampled-token logprob fix, has been merged.
- vLLM-Omni [#5277](https://github.com/vllm-project/vllm-omni/pull/5277), the multi-node lifecycle fix, is cleanly mergeable and close to landing.
- verl [#7106](https://github.com/verl-project/verl/pull/7106) is an optional R2 router-replay fix. The default fully-async path keeps R2 disabled, so it is independent and does not block this PR.
- Megatron-Bridge [#4988](https://github.com/NVIDIA-NeMo/Megatron-Bridge/pull/4988) remains the main required lower-layer dependency for the Qwen3-Omni Megatron full-model path.

Therefore, I think this PR is ready for review now. For final merge ordering, the conservative path is to land #5277 and #4988 first; once #5277 lands, #4988 should be the only remaining required dependency.

### 100-step E2E validation

I completed a 100/100-step Qwen3-Omni text-only full-model fully-async RL run:

- 32 GPUs across 4 nodes
- 16 Megatron training GPUs and 16 standalone vLLM-Omni rollout GPUs
- Four node-local TP4 rollout replicas
- 128 responses per actor update
- Full GSM8K validation every 10 steps
- Full-model weight synchronization after every update

Key results:

- Validation reward: `0.8173 -> 0.9598`
- Training reward:
  - first 10-step mean: `0.8953`
  - last 10-step mean: `0.9695`
- Rollout/actor Pearson correlation:
  - overall mean: `0.9909`
  - last 10-step mean: `0.9922`
- Rollout KL mean: `0.0100`
- Actor grad norm: `0.126` mean, `0.336` maximum
- No aborted responses, dropped stale samples, OOMs, or distributed failures

The latest figure posted above includes the requested training/rollout log-PPL curves, training reward, and full validation reward. The log-PPL curves remain aligned without divergence, while validation reward shows clear task-level improvement.

The implementation and E2E validation are ready, so review can proceed while the remaining lower-layer dependency is being finalized. Thanks!

verl_omni/trainer/config/omni_megatron_trainer.yaml
zhtmike, line 1: perhaps for generality, we should use `omni_megatron_trainer.yaml` instead, with coperated with sync mode later

verl_omni/__init__.py
```diff
@@ -28,11 +28,19 @@
...
 import verl_omni.models  # noqa: E402, F401
 import verl_omni.pipelines  # noqa: E402, F401
 import verl_omni.reward_loop  # noqa: E402, F401
-import verl_omni.trainer  # noqa: E402, F401
-import verl_omni.workers.engine  # noqa: E402, F401
+
+if not _skip_diffusion_components:
+    import verl_omni.experimental  # noqa: E402, F401
+    import verl_omni.trainer  # noqa: E402, F401
+    import verl_omni.workers.engine  # noqa: E402, F401
```
zhtmike: pls revert change

verl_omni/workers/rollout/vllm_rollout/vllm_omni_async_server.py
```diff
@@ -514,7 +624,7 @@ async def _run_generation(self, prompt, params, request_id: str, lora_request, p
         if self._ar_mode:
             generator = self.engine.generate(
                 prompt=prompt,
-                sampling_params_list=params,
+                sampling_params=params,
```
zhtmike: why change endpoint?

verl_omni/workers/rollout/vllm_rollout/vllm_omni_async_server.py
```diff
@@ -321,6 +389,48 @@ async def _sleep_hybrid(self):
...
+            await self.engine.reset_prefix_cache(
+                reset_running_requests=True,
+                reset_connector=True,
+            )
+        return {"aborted": True, "request_id": request_id}
+
+    async def resume_generation(self):
+        if self.node_rank == 0:
+            await self.engine.resume_generation()
+
```
zhtmike: @knlnguyen1802 expert, pls take a look

verl_omni/workers/rollout/vllm_rollout/vllm_omni_async_server.py
```diff
@@ -272,9 +339,10 @@ async def run_server(self, args: argparse.Namespace):
         self._server_port, self._server_task = await run_uvicorn(app, args, self._server_address)
 
     async def run_headless(self, args: argparse.Namespace):
-        """Run headless server in a separate thread."""
-        # TODO (mike): support multi node
-        raise NotImplementedError("vLLM-Omni headless mode is not implemented yet.")
```
zhtmike, line 278: revert my code

verl_omni/workers/rollout/vllm_rollout/vllm_omni_async_server.py
```diff
@@ -218,6 +258,27 @@ def _write_deploy_config(self, engine_kwargs: dict, pipeline_name: str, adapter_
...
+                "Use data_parallel_size=1 and scale with verl-owned outer replicas."
+            )
+        config_path = getattr(args, "deploy_config", None) or getattr(args, "stage_configs_path", None)
+        placement = validate_vllm_omni_rollout_placement(
+            config_path=config_path,
+            visible_device_count=self._visible_device_count(),
+        )
+        logger.info("vLLM-Omni AR rollout placement: %s", placement)
+        return placement
+
```
zhtmike: what happens here, cc @knlnguyen1802

verl_omni/workers/rollout/vllm_rollout/vllm_omni_async_server.py
```diff
@@ -149,27 +169,45 @@ def _preprocess_engine_kwargs(self, engine_kwargs: dict) -> None:
...
+                if adapter_cls is None:
+                    registered = sorted(OmniRolloutPipelineBase._registry)
+                    raise ValueError(
+                        f"No Omni rollout adapter registered for pipeline_name={pipeline_name!r}. "
+                        f"Registered: {registered}"
+                    )
                 self._write_deploy_config(engine_kwargs, pipeline_name, adapter_cls, pipeline_mode)
-                # Store per-stage rollout flags for downstream use.
                 self._rollout_flags = adapter_cls.rollout_flags(pipeline_mode=pipeline_mode)
-                # Merge pipeline-specific HF config overrides.
```
zhtmike, line 164: revert delete

verl_omni/workers/rollout/vllm_rollout/vllm_omni_async_server.py
```diff
@@ -149,27 +169,45 @@ def _preprocess_engine_kwargs(self, engine_kwargs: dict) -> None:
         engine_kwargs.pop("output_mode", None)
         if self._ar_mode:
             engine_kwargs.pop("custom_pipeline", None)
-            # TODO (mike): drop this later
-            # It should be automatically inferred from the model config
```
zhtmike, line 154: revert my todo

verl_omni/workers/rollout/vllm_rollout/vllm_omni_async_server.py
```diff
@@ -97,6 +104,19 @@ def _validate_configs(self) -> None:
...
+                )
+            logprobs_mode = getattr(self.config, "logprobs_mode", "processed_logprobs")
+            if getattr(self.config, "calculate_log_probs", False) and logprobs_mode not in {
+                "raw_logprobs",
+                "processed_logprobs",
+            }:
+                raise ValueError(
+                    "vLLM-Omni AR rollout requires raw_logprobs or processed_logprobs "
+                    f"when calculate_log_probs is enabled, got {logprobs_mode!r}"
+                )
```
zhtmike: seems it is over defensive.

verl_omni/workers/rollout/vllm_rollout/vllm_omni_async_server.py
```diff
@@ -149,27 +169,45 @@ def _preprocess_engine_kwargs(self, engine_kwargs: dict) -> None:
...
-            if adapter_cls is not None:
-                # Generate deploy config using the adapter's stage topology.
+            if pipeline_name:
+                adapter_cls = OmniRolloutPipelineBase.get_class(pipeline_name)
+                if adapter_cls is None:
+                    registered = sorted(OmniRolloutPipelineBase._registry)
+                    raise ValueError(
+                        f"No Omni rollout adapter registered for pipeline_name={pipeline_name!r}. "
+                        f"Registered: {registered}"
+                    )
```
zhtmike: over defensive, again

verl_omni/workers/rollout/vllm_rollout/vllm_omni_async_server.py
```diff
@@ -180,12 +218,14 @@ def _write_deploy_config(self, engine_kwargs: dict, pipeline_name: str, adapter_
...
-        devices = os.environ.get(device_control_env, "")
+        visible_devices = os.environ.get(device_control_env, "")
         tp_size = self.config.tensor_model_parallel_size
 
         deploy_dict: dict[str, object] = {"pipeline": pipeline_id}
 
-        if devices:
+        if visible_devices:
+            device_count = len([device for device in visible_devices.split(",") if device.strip()])
+            devices = ",".join(str(device_id) for device_id in range(device_count))
```
zhtmike, line 208: why change this

verl_omni/workers/rollout/vllm_rollout/utils.py
zhtmike, line 161: add a test to guard this

verl_omni/workers/rollout/vllm_rollout/utils.py
```diff
@@ -25,7 +30,7 @@
 logger.setLevel(os.getenv("VERL_LOGGING_LEVEL", "WARN"))
 
 
-class vLLMOmniColocateWorkerExtension(CustomPipelineWorkerExtension):
+class vLLMOmniColocateWorkerExtension(vLLMColocateWorkerExtension, CustomPipelineWorkerExtension):
```
zhtmike: @knlnguyen1802 , it is necessary?

verl_omni/workers/rollout/vllm_rollout/placement_guard.py
zhtmike, line 1: what is a usage for this file, 
add a header, but keep it simple

verl_omni/pipelines/__init__.py
zhtmike, line 1: why change this file?

verl_omni/trainer/config/_generated_fully_async_omni_megatron_trainer.yaml
```diff
@@ -0,0 +1,1054 @@
+# This reference configration yaml is automatically generated via 'scripts/generate_trainer_config.sh'
+# in which it invokes 'python3 scripts/print_cfg.py --cfg job --config-name=fully_async_omni_megatron_trainer.yaml' to flatten the 'verl_omni/trainer/config/fully_async_omni_megatron_trainer.yaml' config fields into a single file.
```
knlnguyen1802: Why this file is so long ?

verl_omni/workers/rollout/vllm_rollout/utils.py
knlnguyen1802: I dont think it is necessary, please explain what function you need from `vLLMColocateWorkerExtension`

verl_omni/workers/rollout/vllm_rollout/utils.py
```diff
@@ -102,28 +121,13 @@ def update_weights_from_ipc(self, peft_config: dict = None, base_sync_done=False
...
-                    lora_tensors=accumulated_weights,
-                )
-            else:
-                lora_request = OmniTensorLoRARequest(
-                    lora_name=VLLM_LORA_NAME,
-                    lora_int_id=VLLM_LORA_INT_ID,
-                    lora_path=VLLM_LORA_PATH,
-                    peft_config=peft_config,
-                    lora_tensors=accumulated_weights,
-                )
```
knlnguyen1802, line 157: Recover these change if inheritance is not necessary

verl_omni/workers/rollout/vllm_rollout/vllm_omni_async_server.py
knlnguyen1802: agree that might not need these change, it should be guard in vllm-omni

verl_omni/workers/rollout/vllm_rollout/vllm_omni_async_server.py
```diff
@@ -218,6 +258,27 @@ def _write_deploy_config(self, engine_kwargs: dict, pipeline_name: str, adapter_
...
+                "A vLLM-Omni AR rollout replica must fit on one node. "
+                "Use data_parallel_size=1 and scale with verl-owned outer replicas."
+            )
+        config_path = getattr(args, "deploy_config", None) or getattr(args, "stage_configs_path", None)
+        placement = validate_vllm_omni_rollout_placement(
+            config_path=config_path,
+            visible_device_count=self._visible_device_count(),
+        )
+        logger.info("vLLM-Omni AR rollout placement: %s", placement)
+        return placement
```
knlnguyen1802: Why you need this guard, what happen it cannot pass the guard , then how to config the gpu for each stage ?

verl_omni/workers/rollout/vllm_rollout/vllm_omni_async_server.py
knlnguyen1802: These function is necessary for fully async

verl_omni/trainer/config/omni_megatron_trainer.yaml
hbhflw2000: Agreed. I’ll rename it to omni_megatron_trainer.yaml and keep the base config backend-oriented so it can be shared by sync and async recipes. Async-only behavior will remain configurable through recipe overrides.

verl_omni/__init__.py
hbhflw2000: Agreed. This was a local import-isolation workaround and should not be part of this PR.

verl_omni/workers/rollout/vllm_rollout/vllm_omni_async_server.py
hbhflw2000: This changes the AsyncOmni generate argument rather than the HTTP endpoint. However, the existing sampling_params_list path is valid for a single AR stage, so I’ll revert this to minimize the PR scope.

verl_omni/workers/rollout/vllm_rollout/vllm_omni_async_server.py
hbhflw2000: This override is needed because verl’s base lifecycle accesses AsyncLLM.output_processor, which AsyncOmni does not expose. Fully-async weight sync must pause, abort stale requests, clear caches, and resume. I’ll narrow the implementation to AsyncOmni’s public abort API and add lifecycle tests if needed. Expert comment is welcome.

verl_omni/workers/rollout/vllm_rollout/vllm_omni_async_server.py
hbhflw2000: This was a fail-fast guard for the supported topology: each vLLM-Omni AR replica is node-local DP1, while verl owns the outer replicas. I’ll simplify it to a concise launch requirement and rely on vLLM-Omni for deploy-config parsing if needed. Waiting for expert's comment.

verl_omni/workers/rollout/vllm_rollout/vllm_omni_async_server.py
hbhflw2000: Agreed that the logprobs validation and duplicated placement checks are unnecessary. I’ll remove them and keep only one explicit launch-time NotImplemented check for the currently supported DP1 topology.

verl_omni/workers/rollout/vllm_rollout/vllm_omni_async_server.py
hbhflw2000: Agreed. I’ll restore the upstream registry flow and let the canonical adapter/engine validation report invalid configurations.

verl_omni/workers/rollout/vllm_rollout/vllm_omni_async_server.py
hbhflw2000: This is intentional. vLLM-Omni deploy YAML treats stage devices as logical indices relative to CUDA_VISIBLE_DEVICES. Ray may expose physical or non-contiguous IDs, so copying the raw values can remap devices incorrectly. I’ll add a concise comment and retain the offset-device test.

verl_omni/workers/rollout/vllm_rollout/placement_guard.py
hbhflw2000: After simplifying the placement contract, a separate parser module is no longer necessary. I plan to remove this file and keep the small launch guard next to the replica code.

verl_omni/pipelines/__init__.py
hbhflw2000: This was part of the same local dependency workaround and is unrelated to fully-async rollout. I’ll revert the file completely.

verl_omni/workers/rollout/vllm_rollout/vllm_omni_async_server.py
hbhflw2000: Addressed in the current revision: the AR path now constructs `OmniModelConfig` directly, with no `HFModelConfig` fallback.

verl_omni/workers/rollout/vllm_rollout/vllm_omni_async_server.py
hbhflw2000: Addressed with a default Megatron trainer config and its generated reference. Per the latest review, I will rename the shared source config to `omni_megatron_trainer.yaml` so both sync and async recipes can build on it.

examples/gspo_trainer/qwen3_omni/config/qwen3_omni_thinker_gspo_megatron_fully_async.yaml
hbhflw2000: Addressed in the current revision: the recipe-specific YAML was replaced by a short default trainer config plus the generated flattened reference, following the existing config-generation workflow. I will also apply the requested generic `omni_megatron_trainer.yaml` naming.

verl_omni/workers/rollout/vllm_rollout/vllm_omni_async_server.py
hbhflw2000: Agreed. I will restore the original `run_headless` implementation and TODO verbatim; changing the error text is unrelated to this PR.

verl_omni/workers/rollout/vllm_rollout/vllm_omni_async_server.py
hbhflw2000: Agreed. I will restore the original comments; their deletion was unrelated churn.

verl_omni/workers/rollout/vllm_rollout/vllm_omni_async_server.py
hbhflw2000: Agreed. I will restore the original TODO and attribution verbatim.

verl_omni/workers/rollout/vllm_rollout/utils.py
hbhflw2000: Agreed. I will add a two-bucket test that verifies all LoRA tensors are accumulated and passed to exactly one atomic `add_lora` call.

verl_omni/workers/rollout/vllm_rollout/utils.py
hbhflw2000: The AR path needs `vLLMColocateWorkerExtension.update_weights_from_ipc` for verl's HF-to-vLLM weight-name normalization, MoE weight-loader patching, bounded bucket loading, and one-shot post-load processing. However, I agree that mixing it into the diffusion extension is unnecessary. I will introduce a dedicated AR extension inheriting `vLLMColocateWorkerExtension` and leave the existing diffusion extension unchanged.

verl_omni/trainer/config/_generated_fully_async_omni_megatron_trainer.yaml
hbhflw2000: This is the flattened reference generated by `scripts/generate_trainer_config.sh`, following the existing `_generated_*` config convention and the earlier request for a default Megatron trainer config. It expands the inherited verl config, hence the size; users edit the short source config, not this generated file. I will rename and regenerate it together with `omni_megatron_trainer.yaml`.

verl_omni/workers/rollout/vllm_rollout/utils.py
hbhflw2000: Agreed. After separating the AR extension, I will restore the diffusion worker's existing inheritance, `__new__`, and Omni-specific LoRA path. Only the dedicated AR extension will reuse verl's standard vLLM weight-update implementation.

verl_omni/workers/rollout/vllm_rollout/vllm_omni_async_server.py
hbhflw2000: Agreed. On a second pass, these DP/logprob checks duplicate backend validation. I will remove them from verl-omni and let vLLM-Omni own stage/config validation. The reference recipe will still use the currently validated DP1 topology.

verl_omni/workers/rollout/vllm_rollout/vllm_omni_async_server.py
hbhflw2000: This guard was added after multi-node bring-up failures to enforce the topology we validated: each verl-owned outer rollout replica contains one node-local vLLM-Omni stage using actor-local logical device IDs. If it failed, users had to change the outer replica/TP allocation or stage devices, which is too much policy for this layer. I will remove this parser/guard and rely on vLLM-Omni's stage placement validation; the recipe will continue to emit logical devices and default to DP1.

verl_omni/workers/rollout/vllm_rollout/vllm_omni_async_server.py
hbhflw2000: Update after the expert feedback below: I will remove the custom placement preflight entirely rather than keep a second launch guard in verl-omni. vLLM-Omni will own stage placement validation, while the reference recipe remains on the validated node-local DP1 topology.

verl_omni/workers/rollout/vllm_rollout/vllm_omni_async_server.py
hbhflw2000: Update after the expert review: I will remove the extra DP and logprob guards from verl-omni instead of retaining a duplicate launch-time check. The supported recipe still defaults to DP1, and backend validation remains in vLLM-Omni.

verl_omni/workers/rollout/vllm_rollout/placement_guard.py
hbhflw2000: Update after the placement discussion: I will remove both this parser module and its custom launch guard. The recipe will retain the validated logical-device/DP1 defaults, while vLLM-Omni remains responsible for stage placement validation.

hbhflw2000:
Hi @zhtmike @knlnguyen1802 ，thanks for the detailed review, I've addressed the feedbacks, could you take a look when convinient?

tests/workers/rollout/rollout_vllm/test_vllm_omni_fully_async_on_cpu.py
zhtmike, line 1: is it running on cpu or gpu?

if it is cpu, you need to suffix.
if it is gpu, you need to update yaml or shell script for gpu

verl_omni/workers/rollout/vllm_rollout/utils.py
```diff
@@ -25,6 +31,14 @@
...
 
 
+class vLLMOmniARColocateWorkerExtension(vLLMColocateWorkerExtension):
+    """Use verl's standard vLLM weight update path for autoregressive stages."""
+
+    def _maybe_reload_standard_weights_from_ipc(self, receiver) -> bool:
+        """Keep Qwen3-Omni MoE weight loading bounded by IPC bucket size."""
+        return False
+
+
```
zhtmike: @knlnguyen1802 is it fine to bind vllm-omni with vllm... ?

zhtmike: kindly ping @knlnguyen1802 for this~

knlnguyen1802: I suggest not to bind them like this even if for AR it is very closed to vLLM, we might need a lot of patching in the future when extend to more model (you can see the list of support model in vllm-omni)

hbhflw2000: @knlnguyen1802 Thanks, I agree with your concern. I rechecked the latest main and confirmed that the existing `vLLMOmniColocateWorkerExtension` already supports standard AR model detection, bucketed full-weight loading, MoE weight-loader patching, and one-shot post-load processing.

I will remove the subclass of verl's `vLLMColocateWorkerExtension` and restore the native vLLM-Omni extension for AR mode. I will also add a CPU unit test covering the AR full-weight update path. This keeps model-specific patches and future model support owned by vLLM-Omni rather than coupling the AR path directly to verl's vLLM extension.

tests/workers/rollout/rollout_vllm/test_vllm_omni_fully_async_on_cpu.py
hbhflw2000: Thanks, you are right. These are CPU-only unit tests using mocked workers and engines; they do not require a GPU. I will rename the file to `test_vllm_omni_fully_async_on_cpu.py` so it is collected by the existing CPU CI pattern. No GPU YAML or GPU test entry is needed.

hbhflw2000:
Hi, @zhtmike @knlnguyen1802 , I've addressed problems noticed in previous comments, passed local/e2e tests, could you take a look when convinient? Thx

verl_omni/workers/rollout/vllm_rollout/utils.py
zhtmike, line 1: since the code change is staled, pls revert the change of thie file

zhtmike (commented):
looks good, one comment

verl_omni/workers/rollout/vllm_rollout/utils.py
```diff
@@ -42,11 +46,7 @@ class vLLMOmniColocateWorkerExtension(CustomPipelineWorkerExtension):
     _pending_lora_peft_config: dict | None = None
 
     def __new__(cls, **kwargs):
-        set_death_signal()
-
-        # 1. patch for Lora
```
knlnguyen1802, line 48: Need to revert this after remove inherit from verl

verl_omni/workers/rollout/vllm_rollout/utils.py
hbhflw2000: Done. I merged the latest `main` and fully reverted the stale changes in this file, so `utils.py` no longer appears in the PR diff. The CPU test now follows the current per-update ZMQ contract from `main`.

verl_omni/workers/rollout/vllm_rollout/utils.py
hbhflw2000: Done. The inheritance-related changes have been fully reverted. The PR now uses the native vLLM-Omni worker extension from current `main`, and `utils.py` is no longer part of the PR diff.

zhtmike (approved):

hbhflw2000:
The CPU CI failure occurs because the pinned vLLM-Omni uses Python 3.11-only StrEnum, while the job runs on Python 3.10. It is unrelated to this PR’s implementation, and all other checks have passed.

zhtmike:
> The CPU CI failure occurs because the pinned vLLM-Omni uses Python 3.11-only StrEnum, while the job runs on Python 3.10. It is unrelated to this PR’s implementation, and all other checks have passed.

pls fix the error introduced in the your new test script~

zhtmike:
and since it is a part of a large PR, pls update title with [1/N]

hbhflw2000:
Thanks. I updated the title and fixed the CPU collection failure without monkeypatching: only the three tests requiring the real vLLM-Omni diffusion worker are skipped on Python 3.10, while they still exercise the real import and weight-update path on Python 3.11+. This is a temporary test-side workaround for the pinned vLLM-Omni revision’s `StrEnum` incompatibility and can be removed once the pin advances to the current fixed version.

hbhflw2000:
Hi, @zhtmike Mike, seems the same problem occurred. I checked the upstream fix further: vLLM-Omni #5366 landed only after the v0.25 rebase, while release/v0.24.1 still uses StrEnum. Updating the pin directly would therefore also require upgrading the vLLM stack to 0.25. Would you prefer a temporary module-level skip for this Python 3.10 test, or should I submit a minimal #5366 backport to release/v0.24.1 and update the pin afterward?

zhtmike:
> Hi, @zhtmike Mike, seems the same problem occurred. I checked the upstream fix further: vLLM-Omni #5366 landed only after the v0.25 rebase, while release/v0.24.1 still uses StrEnum. Updating the pin directly would therefore also require upgrading the vLLM stack to 0.25. Would you prefer a temporary module-level skip for this Python 3.10 test, or should I submit a minimal #5366 backport to release/v0.24.1 and update the pin afterward?

simply drop the test case, may add it back after vllm upgraded.

zhtmike:
let us wait for v0.2.0 release, then merge this

Issue #280, opened 2026-07-17
[RFC] Rollout-train log-prob consistency: measurements + default monitoring


cr-gao:
Follow-up to the #97 roadmap item on trainer/rollout consistency. I measured the mismatch across three pipelines before proposing changes. Where the existing consistency engineering applies, it holds — matched backends sit at the bf16 noise floor. But BAGEL shipped a schedule regression that breaks replay entirely, found and fixed below. The proposal is observability + docs + that bugfix; no kernel work is needed.

## Measurement

verl-omni `01c87ee` + vllm-omni `fe478a95`. `d = old_log_probs (trainer replay) − rollout_log_probs (vLLM-Omni)`, dumped after the `old_log_probs` recompute and before any optimizer update. 3 training steps × 64 samples per arm, seed 42. Scheduler math is shared fp32 code and the sigma tables match bit-exactly, so `d` isolates the bf16 transformer forward pair. Mean `d` is slightly negative everywhere; expected, since the replay scores the rollout's own sample.

**SD3.5-Medium LoRA OCR** (`examples/flowgrpo_trainer/sd35/`, 3×RTX 4090):

| arm | variation | mean\|d\| | p50 | p99 |
|---|---|---|---|---|
| a0 | baseline (native/TORCH_SDPA, bf16, cps) | 2.24e-5 | 1.44e-5 | 1.30e-4 |
| a3 | sde_type=sde | 3.60e-5 | 1.42e-5 | 2.60e-4 |
| a4 | late window `[5,10]`, t ∈ [9, 707] | 6.13e-5 | 5.84e-5 | 2.06e-4 |
| a5 | CFG `guidance_scale=4.5` | 4.71e-4 | 2.90e-4 | 2.69e-3 |

Three more arms with FA3 hub attention, fp32 trainer forward, replay micro-batch 4 are all within single-seed noise of a0.

**Second and third implementation pairs** (4×RTX PRO 6000 96GB):

| pair | recipe | mean\|d\| | p50 | p99 |
|---|---|---|---|---|
| Qwen-Image 20B | #188's FlowDPPO OCR: true CFG 4.0 + sde | 5.89e-5 | 3.55e-5 | 2.95e-4 |
| BAGEL-7B-MoT as-found | FlowGRPO OCR: 3-branch CFG + renorm, sde | **5.03e-3** | 5.00e-3 | 5.56e-3 |
[...]

PR #291, opened 2026-07-22
[trainer, doc, tests] feat: log rollout-train log-prob consistency metrics by default

cr-gao:
### What

Implements the observability part of RFC #280: rollout-train log-prob consistency is measurable today only with throwaway code — `compute_rollout_corr_metrics_from_logprobs` has no callers and nothing reports the gap that let the #279 regression ship silently.

- **Default monitoring**: when the rollout returns log-probs (`calculate_log_probs=true`) and bypass mode is off, the trainer logs `rollout_corr/*` diagnostics per global batch — mean/max `|old_log_probs − rollout_log_probs|` plus the existing KL/PPL family. No new config, no behavior change; inactive when log-probs are absent (the default).
- **Per-timestep breakdown**: `|Δlogp|` grouped by denoising **timestep value** as `rollout_corr/logprob_abs_diff/ts_{t}` — the SDE window is per-sample random, so column index would mix timesteps; values come from the fixed schedule table, so key cardinality is bounded by `num_inference_steps`. Emitted only when `all_timesteps` aligns with the log-prob columns.
- **Docs**: consistency-monitoring section in `docs/algo/rollout_correction.md` — what the metrics mean, what a healthy magnitude looks like, and the caveat that recipes shipping `clip_ratio`/`kl_mask_threshold` at 1e-5 sit below the noise floor.
- **VeOmni**: `validate_attention_consistency` warns instead of silently skipping `strategy=veomni`.

### Duplicate check

No open PR touches consistency monitoring; follows up my own RFC #280. Omni-model pearson correlation (review point 2) is deferred until the omni trainer refactor lands, as discussed in the RFC thread.

### Test

- `pytest tests/trainer/diffusion/test_rollout_correction_on_cpu.py tests/utils/test_diffusion_attention_on_cpu.py` — 25 passed, including 5 new ones: exact aggregate/group values, no-timesteps backward compat, shape-mismatch guard, veomni warning.
- E2e on 3×RTX 4090, SD3.5-Medium LoRA OCR recipe, `calculate_log_probs=True`, 2 steps: metrics appear once per global batch and reproduce the offline analysis behind #280 on the same recipe/seed — step 1 `logprob_abs_diff_mean` 2.0970e-5 / `_max` 1.3846e-4 vs 2.097e-5 / 1.385e-4 measured offline. The breakdown emits the 5 timestep values the recipe's window actually covers (`ts_1000` 6.1e-6 → `ts_790.4` 3.4e-5), matching the trainer's scheduler table exactly.
- `ruff check` / `ruff format` clean.

### DIFF
--- a/docs/algo/rollout_correction.md
+++ b/docs/algo/rollout_correction.md
@@ -1,6 +1,6 @@
 # Rollout Correction for Diffusion Training (Experimental)
 
-Last updated: 05/19/2026
+Last updated: 07/22/2026
 
 > **Status:** Experimental. The API, default thresholds and recommended preset may change.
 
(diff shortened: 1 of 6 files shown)


zhtmike:
good

zhtmike (approved):
LGTM

docs/algo/rollout_correction.md
```diff
@@ -85,6 +85,29 @@ rollout backend is drifting too far — tighten the RS band or fall back to
...
+guidance scale and later denoising steps amplify it. Values orders of magnitude above that
+indicate a schedule or backend mismatch — enable `calculate_log_probs=true` and check these
+metrics first when debugging convergence.
+
+> **Threshold caveat:** some recipes ship `diffusion_loss.clip_ratio` or `kl_mask_threshold`
+> at 1e-5, below the measured noise floor. Bypass mode is the worst case — the PPO ratio
+> `exp(current − rollout)` then carries the full rollout-train gap, so most samples fall
+> outside the clip band / inside the mask from numerical noise alone. Decoupled mode compares
+> against the recompute instead, but a noticeable `actor/pg_clipfrac` can still be pure noise
+> there. Keep thresholds above the |Δlogp| level these metrics report (the default is `1e-4`).
```
zhtmike, line 109: This is not True btw, @cr-gao 

clip_ratio = 1e-5 is a correct hypar-parameter value, don't let AI mess this up

cr-gao: Thanks for catching it, I double-checked and find that log-probs are mean-pooled, and the `kl_mask_threshold` half was also misinterpreted. Fixed in PR #292 .

PR #289, opened 2026-07-21
[reward, rollout, diffusion] feat: add DiNa-LRM support for SD3.5 FlowGRPO

GongyeLiu:
### What does this PR do?

> Add **concise** overview of what this PR aims to achieve or accomplish. Reference related GitHub issues and PRs that help with the review.

- Add [DiNa-LRM](https://github.com/HKUST-C4G/diffusion-rm) scoring for SD3.5 FlowGRPO.
- Add `image`, `latent`, and `both` rollout output modes.
- Add a PickScore-SFW data preprocessing script, an 8-GPU training recipe, and documentation.
- Build on the SD3.5 FlowGRPO support introduced in #178.

DiNa-LRM is a latent-space diffusion reward model. Unlike image reward models like HPS-v3, DiNa-LRM scores the clean diffusion latent directly. This results in faster rollout and lower GPU memory consumption.

The training rollout uses `output_type=latent` to skip VAE decoding, while validation uses `output_type=both` to preserve decoded images for logging.

The experiment uses PickScore-SFW prompts with:
- Base model: SD3.5-Medium
- Reward model: DiNa-LRM-SD35M-12layers
- 7 GPUs for actor and rollout workers
- 1 GPU for the DRM server

The hyperparameters used in the experiment are:  `clip_ratio=3e-6`, `lr=5e-5`, `rollout.n=18`, `kl_loss_coef=0.01`,  `train_batch_size=14`, `ppo_mini_batch_size=7`, `ppo_micro_batch_size_per_gpu=9`.
Namely, each training
step generates 252 rollout samples. The prompts are divided into two PPO mini-batches of 7 prompts each, resulting in two PPO update rounds per training step.

image

~100s to run one step (252 trajectories):
image

Peak VRAMs: 40G for 512x512 images:

image

Validation: ~80 steps is enough for improving quality:

image

image

### Checklist Before Starting

- [x] Search for similar PRs. Paste at least one query link here: ...
- [x] Format the PR title as `[{modules}] {type}: {description}` (This will be checked by the CI)
  - `{modules}` include `fsdp`, `vllm_omni`, `rollout`, `trainer`, `ci`, `training_utils`, `recipe`, `ray`, `worker`, `single_controller`, `misc`, `perf`, `model`, `algo`, `env`, `tool`, `ckpt`, `doc`, `data`, `cfg`, `reward`, `diffusion`, `omni`, `tests`, `docker`
  - If this PR involves multiple modules, separate them with `,` like `[diffusion, doc]`
[...]

PR #288, opened 2026-07-21
[rollout] refactor: multi-tokenizer prompt path in agent loop for SD3.5

AndyZhou952:
### What does this PR do?

> Add **concise** overview of what this PR aims to achieve or accomplish. Reference related GitHub issues and PRs that help with the review.

- Add `actor_rollout_ref.model.extra_tokenizers` so multi-encoder diffusion models (SD3.5: CLIP + T5) tokenize once in the agent loop and pass `extra_prompt_ids` to rollout — no decode/re-encode in the pipeline.
- Refactor SD3.5 rollout to encode prompts from token ids via `SD3TokenIdPromptMixin`, with vLLM-Omni server forwarding `extra_prompt_ids` / `negative_extra_prompt_ids`.
- Document the agent-loop → rollout tokenization contract and wire the SD3.5 OCR LoRA example; add CPU parity tests for token-id vs text-path encoding.

### Checklist Before Starting

- [x] Search for similar PRs. Paste at least one query link here: ...
- [x] Format the PR title as `[{modules}] {type}: {description}` (This will be checked by the CI)
  - `{modules}` include `fsdp`, `vllm_omni`, `rollout`, `trainer`, `ci`, `training_utils`, `recipe`, `ray`, `worker`, `single_controller`, `misc`, `perf`, `model`, `algo`, `env`, `tool`, `ckpt`, `doc`, `data`, `cfg`, `reward`, `diffusion`, `omni`, `tests`, `docker`
  - If this PR involves multiple modules, separate them with `,` like `[diffusion, doc]`
  - `{type}` is in `feat`, `fix`, `refactor`, `chore`, `test`
  - If this PR breaks any API (CLI arguments, config, function signature, etc.), add `[BREAKING]` to the beginning of the title.
  - Example: `[BREAKING][diffusion, fsdp] feat: new rollout scheduler`

### Test

> For changes that can not be tested by CI (e.g., algorithm implementation, new model support), validate by experiment(s) and show results like training curve plots, evaluation results, etc.

The result of ten separate runs look normal.
image

### API and Usage Example

> Demonstrate how the API changes if any, and provide usage example(s) if possible.

```python
# Add code snippet or script demonstrating how to use this
```

### Design & Code Changes

> Demonstrate the high-level design if this PR is complex, and list the specific changes.

### Checklist Before Submitting

> [!IMPORTANT]
> Please check all the following items before requesting a review, otherwise the reviewer might deprioritize this PR for review.

[...]

PR #282, opened 2026-07-18
[model, rollout, reward, algo, worker, trainer] feat: Qwen3-TTS talker GSPO + online DPO

williammtan:
### What does this PR do?

Adds native Qwen3-TTS talker RL to verl-omni as two recipes that share one talker actor and one
vLLM-Omni AR rollout:

- **GSPO** (group RL): a multi-metric audio reward (intelligibility, speaker similarity, emotion, stability) through verl's stock GSPO policy loss and GDPO advantage estimator.
- **Online DPO**: two candidates per prompt, judged pairwise by a remote audio-capable LLM (Gemini by default, any endpoint), trained with a pairwise DPO loss on the talker's codec-0 sequence log-prob against a frozen reference.

Addresses RFC #90. The talker is not a plain causal LM (2-channel input, speaker slot, 16 codebooks), so the actor computes real teacher-forced codec-0 log-probs and the rollout stays on-policy while the codec2wav decoding is reward-side.

### Checklist Before Starting

- [x] Searched for similar open PRs: [`is:pr is:open tts`](https://github.com/verl-project/verl-omni/pulls?q=is%3Apr+is%3Aopen+tts), plus `qwen3-tts`, `speechjudge`, `audio judge dpo`. No open PR contains Qwen3-TTS or TTS RL (the single `tts` hit, #233, is a CI refactor). RFC #90 requests this recipe and has no implementation PR.
- [x] Title formatted as `[{modules}] {type}: {description}`: `[model, rollout, reward, algo, worker, trainer] feat: Qwen3-TTS talker GSPO + online DPO`.
- [x] No breaking API change: everything is additive (a new `[tts]` extra, new `reward.*` / `actor.policy_loss` config keys, a new trainer entry). Existing T2I/LLM/diffusion behavior is untouched.

#### CPU tests

```bash
pytest -q --asyncio-mode=auto tests/     # python_files = *_on_cpu.py
```

```text
299 passed   # 45 in the 7 TTS-specific files
```

| Area | File |
| --- | --- |
| talker forward (layout; verl-gather == talker-gather) | `tests/models/test_qwen3_tts_forward_on_cpu.py` |
| reward functions (asr/sim/emo, CER, stability, composition) | `tests/utils/reward_score/test_tts_rewards_on_cpu.py` |
| audio reward managers (decode, multi weighted-sum) | `tests/reward_loop/test_audio_reward_manager_on_cpu.py` |
[...]

PR #149, opened 2026-06-07
[diffusion] feat: add FLUX FlowGRPO adapters

SH111-H:
### What does this PR do?

This PR adds first-class FLUX FlowGRPO support to verl-omni.

It introduces a new `flux_flow_grpo` pipeline package with both sides of the verl-omni diffusion integration:

- a training-side `DiffusionModelBase` adapter for FLUX FlowGRPO log-prob recomputation
- a vLLM-Omni custom rollout pipeline that returns the trajectory tensors and conditioning metadata required by FlowGRPO

Compared with existing Qwen-Image / Wan / SD3 integrations, FLUX needs its own adapter because it uses packed latent sequences, FLUX-specific text/image position ids, pooled CLIP projections, T5 prompt embeddings, and guidance behavior that does not map cleanly to the existing pipelines.

### Checklist Before Starting

- [x] Search for similar PRs. Query link: https://github.com/verl-project/verl-omni/pulls?q=is%3Apr+FLUX+FlowGRPO
- [x] Format the PR title as `[{modules}] {type}: {description}`. Title: `[diffusion] feat: add FLUX FlowGRPO adapters`

### Test

Validated in a project environment with vLLM-Omni dependencies:

```bash
python -m pytest tests/pipelines/test_flux_flow_grpo_adapter_on_cpu.py -q
```

Result:

```text
11 passed
```

Also validated:

```bash
python -m ruff check verl_omni/pipelines/flux_flow_grpo tests/pipelines/test_flux_flow_grpo_adapter_on_cpu.py
python -m ruff format --check verl_omni/pipelines/flux_flow_grpo tests/pipelines/test_flux_flow_grpo_adapter_on_cpu.py
python -m compileall -q verl_omni/pipelines/flux_flow_grpo tests/pipelines/test_flux_flow_grpo_adapter_on_cpu.py
```

A real FLUX checkpoint smoke check confirmed that the custom pipeline constructor receives the checkpoint transformer config, including `guidance_embeds=False`.

Note: a full Ray e2e run is environment-sensitive. The included special e2e script assumes a local FLUX checkpoint in diffusers layout and a vLLM-Omni installation with FLUX support.

### API and Usage Example

FLUX FlowGRPO is registered through the existing model/pipeline registry mechanism:

```python
@DiffusionModelBase.register("FluxPipeline", algorithm="flow_grpo")
class Flux(DiffusionModelBase):
    ...

@VllmOmniPipelineBase.register("FluxPipeline", algorithm="flow_grpo")
[...]

PR #297, opened 2026-07-23
[reward, diffusion, tests] fix: align reward_extra_info keys across the batch

JeremyJC67:
# [reward, diffusion, tests] fix: align reward_extra_info keys across the batch

## Problem

`MultiVisualRewardManager` publishes a **sample-dependent** key set. For a sub-reward
that returns a dict it emits `reward/{key}` *plus* `reward/{key}/{subkey}` for every
extra field ([`multi.py#L159-L164`](https://github.com/verl-project/verl-omni/blob/main/verl_omni/reward_loop/reward_manager/multi.py#L159-L164)),
but when the sub-reward raises, the exception path emits **only** `reward/{key}`
([`multi.py#L168-L172`](https://github.com/verl-project/verl-omni/blob/main/verl_omni/reward_loop/reward_manager/multi.py#L168-L172)).

`DiffusionAgentLoopWorker._postprocess` took the key list from sample 0 alone and then
indexed every other sample with `info[key]`:

```python
reward_extra_keys = list(reward_extra_infos[0].keys())
for key in reward_extra_keys:
    non_tensor_batch[key] = np.array([info[key] for info in reward_extra_infos])
```

So a batch whose samples disagree on keys either crashes or silently loses data:

1. **Crash** — a sub-reward that fails on any sample *other than the first* raises
   `KeyError: 'reward/{key}/{subkey}'` and kills the whole training step.
2. **Silent data loss** — a sub-reward that fails on *sample 0* makes the subkey
   disappear from the batch for every other sample, with no warning.

This is reachable today: `genrm_ocr.compute_score` returns `{"score", "genrm_response"}`
and talks to a GenRM over HTTP, so a single timeout in a batch is enough. It is used by
`examples/flowgrpo_trainer/qwen_image/run_qwen_image_ocr_multi_reward.sh` and
`tests/special_e2e/run_flowgrpo_qwen_image.sh`.

## Fix

Collect the union of keys over the batch and fill gaps with `None`, mirroring how
`extra_fields` is already aggregated a few lines below in the same function
(`np.empty(..., dtype=object)` + `.get(key)`).

When every sample carries the key — the common case — the original `np.array` path is
kept unchanged, so numeric columns keep their `float64` dtype and downstream metric
reduction is unaffected. The object-dtype path is only taken for keys that are genuinely
missing somewhere.

## Not a duplicate

Checked before opening, per `AGENTS.md`:

```
gh pr list --repo verl-project/verl-omni --state open --search "reward_extra_info"
gh pr list --repo verl-project/verl-omni --state open --search "reward extra keys"
gh pr list --repo verl-project/verl-omni --state open --search "diffusion_agent_loop"
gh issue list --repo verl-project/verl-omni --state open --search "reward_extra_info"
```

No open PR or issue touches this path. The keyword hits (#282, #278, #115, #290, #288,
#149) are feature PRs that do not modify `_postprocess`'s reward-extra aggregation.
Closest neighbour is **#289**, which adds a `required` flag so a sub-reward can fail-fast
instead of zero-filling — that changes *policy* on failure, while this PR fixes the
*batch aggregation* that breaks once any failure path is taken. They are complementary
and do not touch the same lines; happy to rebase if #289 lands first.

## Tests

New CPU test file `tests/agent_loop/test_diffusion_agent_loop_postprocess_on_cpu.py`
(5 cases): later-sample failure, first-sample failure, dtype-preservation regression
guard, all-samples-failed, and disjoint key sets across two sub-rewards.

Verified the tests actually catch the bug — with the source change reverted and only the
tests applied, 3 of the 5 fail at the exact line being fixed:

```
$ git checkout HEAD~1 -- verl_omni/agent_loop/diffusion_agent_loop.py
$ pytest -q --asyncio-mode=auto tests/agent_loop/test_diffusion_agent_loop_postprocess_on_cpu.py
E   KeyError: 'reward/ocr/genrm_response'
verl_omni/agent_loop/diffusion_agent_loop.py:347: KeyError
E   KeyError: 'reward/a'
verl_omni/agent_loop/diffusion_agent_loop.py:347: KeyError
3 failed, 2 passed, 16 warnings in 6.86s
```

With the fix applied:

```
$ pytest -q --asyncio-mode=auto tests/agent_loop/test_diffusion_agent_loop_postprocess_on_cpu.py
5 passed, 16 warnings in 6.78s
```

Full CPU suite, no regressions:

```
$ pytest -q --asyncio-mode=auto tests/          # pytest.ini: python_files = *_on_cpu.py
264 passed, 18 warnings in 10.42s
```

Lint:

```
$ ruff check verl_omni/agent_loop/diffusion_agent_loop.py tests/agent_loop/test_diffusion_agent_loop_postprocess_on_cpu.py
All checks passed!
$ ruff format --diff <same files>
2 files already formatted
```

Environment: Linux x86_64, Python 3.10, vllm 0.24.0+cpu, torch 2.11.0+cpu, verl @ pinned
commit — i.e. the `cpu_unit_tests.yml` recipe.

## AI assistance

AI assistance (Claude) was used to locate the defect, draft the patch and write the
tests. Every command and result shown above was actually executed on the environment
described, and is reproducible from this branch.

Opening as a **draft** while I complete my own line-by-line review of the change; I will
mark it ready for review once that is done. Please do not spend review time on it until
then.

## Follow-up

The same sample-0-key-set pattern exists upstream in
`verl/experimental/agent_loop/agent_loop.py`. Happy to send the matching fix to `verl`
once this lands, if maintainers agree the approach is right.

### DIFF
--- a/tests/agent_loop/test_diffusion_agent_loop_postprocess_on_cpu.py
+++ b/tests/agent_loop/test_diffusion_agent_loop_postprocess_on_cpu.py
@@ -1,0 +1,93 @@
+# Copyright 2026 Bytedance Ltd. and/or its affiliates
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+"""CPU tests for reward_extra_info aggregation in DiffusionAgentLoopWorker._postprocess.
+
+Sub-rewards do not publish a fixed key set per sample. ``MultiVisualRewardManager``
+emits ``reward/{key}/{subkey}`` for every field of a dict result, but on an exception
+it emits only ``reward/{key}``. ``_postprocess`` must therefore align a batch whose
+samples carry different reward-extra keys.
+"""
+
+import numpy as np
+import torch
+from verl.experimental.agent_loop.agent_loop import AgentLoopMetrics
+
+from verl_omni.agent_loop.diffusion_agent_loop import (
+    DiffusionAgentLoopWorker,
+    _InternalDiffusionAgentLoopOutput,
+)
+
+# Mirrors MultiVisualRewardManager: a dict-returning sub-reward publishes the extra
+# field, the exception path publishes the bare score only.
+_OK = {"reward/ocr": 1.0, "reward/ocr/genrm_response": "matched", "reward/combined": 1.0}
+_FAILED = {"reward/ocr": 0.0, "reward/combined": 0.0}
+
+
+def _make_output(reward_extra_info: dict) -> _InternalDiffusionAgentLoopOutput:
+    return _InternalDiffusionAgentLoopOutput(
+        prompt_ids=torch.zeros(1, 4, dtype=torch.long),
+        response_diffusion_output=torch.zeros(1, 3, 8, 8),
+        reward_score=float(reward_extra_info["reward/combined"]),
+        num_turns=1,
+        metrics=AgentLoopMetrics(),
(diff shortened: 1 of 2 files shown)


zhtmike:
what is the issue exactly, can you open an issue first?

JeremyJC67:
Sure! I would update current issue asap~

Issue #321, opened 2026-07-30
[RFC] DigenRL: A Disaggregated RL framework for Diffusion-Model on Heterogeneous Resources


maochiyu1111:
# [RFC] DigenRL: A Disaggregated Execution Framework for Diffusion-Model RL on Heterogeneous Resources

## 1. Summary

This RFC proposes DigenRL, a disaggregated execution framework for reinforcement learning post-training of diffusion models on heterogeneous resources. Its core design organizes the Generator, Trainer, and reward/scoring components into independently configurable resource pools while preserving the algorithmic semantics of existing on-policy GRPO training, then reduces execution bubbles through Generation-Axis Pipeline (GAP), Timestep Parallelism (TSP), Elastic Trainer-Assisted Generation (TAG), and Recursive Bubble-Triggered TAG Search (BTS). Trajectory-Consistent Stale Synchronization (TCSS) is an explicitly enabled bounded-staleness extension; when TCSS is disabled, the framework preserves strict on-policy train-batch boundaries and trajectory-level policy consistency.

### 2.1 Current Execution Path

Diffusion-model RL systems commonly use the following execution path:

Rollout in diffusion-model RL must proceed sequentially through the denoising timesteps, whereas training on a fixed trajectory can process multiple samples and timesteps in parallel. The two phases have different scaling characteristics, memory requirements, and preferred device topologies. In today's colocated execution model, they occupy the same resource pool in sequence, making it difficult to use heterogeneous devices efficiently. The colocated execution path is shown below:

```
TaskRunner
  → global_pool: Role.ActorRollout / Role.ActorRolloutRef
  → BaseRayDiffusionTrainer._init_online_rollout_stack()
  → DiffusionAgentLoopWorker.generate_sequences(complete DataProto)
  → sleep_replicas()
  → reward / old log-prob / ref log-prob / advantage
  → actor_rollout_wg.update_actor()
  → CheckpointEngineManager.update_weights()
```

### 2.2 Resource Disaggregation Alone Is Not Enough

Simply assigning the two phases to separate resource pools is not enough. If training must wait for the entire rollout batch, the Generator and Trainer will still alternate between idle periods on opposite sides of the batch barrier.

```
Generator: [ generate the whole batch ][ idle during training ][ wait for sync ]
Trainer:   [          idle           ][ train whole batch   ][ sync          ]
```

### 2.3 On-Policy Disaggregated Execution Framework

[...]

PR #331, opened 2026-08-02
[diffusion, trainer, ray, cfg, tests] feat: add separate diffusion training baseline

maochiyu1111:
### What does this PR do?

This PR adds an opt-in resource-separated execution mode for the current synchronous diffusion policy-gradient
trainer. With `actor_rollout_ref.separate=true`, training workers remain pure Actor workers while rollout runs through
the existing standalone `LLMServerManager` and `CheckpointEngineManager` on independent Ray resources.

The default colocated path remains unchanged. In separate mode, the Trainer synchronizes updated weights to rollout
workers after each logical-batch update, with one initial synchronization before training. Unsupported trainer,
sample-source, rollout or checkpoint configurations fail early.

Separate mode currently supports **full finetuning only**. `actor_rollout_ref.model.lora.rank`, the compatibility
field `actor_rollout_ref.model.lora_rank`, and `actor_rollout_ref.model.lora_adapter_path` fail closed during
configuration validation, before Ray initialization.

Related issue: [#321](https://github.com/verl-project/verl-omni/issues/321).

Main changes:

- add the documented, default-off `actor_rollout_ref.separate` configuration;
- add pure-Actor resource mapping, standalone rollout initialization and guarded colocated lifecycle calls;
- add fail-fast validation and focused CPU control-flow/config tests;
- reject all LoRA entry points before Ray initialization;
- document separate mode and add a pinned-stack GPU smoke.

### Motivation and Scope

This is the synchronous foundation for running diffusion Trainer and Generator workers on distinct Ray resources,
without introducing asynchronous or stale-sample semantics.

This PR includes:

- current `trainer.use_v1=false`, online policy-gradient execution;
- one full rollout/reward/update barrier per logical batch;
- synchronous initial and post-update weight publication from Trainer to rollout workers;
- full-finetuning-only synchronous separate mode; LoRA configurations are rejected before Ray initialization.

This PR does not include:

- Verl v1 async trainers, queues, partial rollouts, pipeline overlap or stale samples;
- reward-worker placement changes;
- changes to reward, advantage, loss, checkpoint or data formats;
- a Verl dependency bump or the validation-only backport of an upstream compatibility fix;
- local A/B launch wrappers, machine-specific runtime environment files or experiment logs.

### Checklist Before Starting

- [x] Read-only duplicate search was run for issue #321 and relevant keywords.
- [x] The PR title follows `[{modules}] {type}: {description}`.
- [x] The material difference from open PR #315 has been explained in issue #321.

Search links:

- [`321 in:body`](https://github.com/verl-project/verl-omni/pulls?q=is%3Apr+is%3Aopen+321+in%3Abody)
- [`separate diffusion trainer rollout`](https://github.com/verl-project/verl-omni/pulls?q=is%3Apr+is%3Aopen+separate+diffusion+trainer+rollout)

The issue-number query returned no open PR. The closest relevant result from the keyword query is
[#315](https://github.com/verl-project/verl-omni/pull/315), which adds Verl v1 separate-async
trainer/queue/checkpoint behavior. This PR instead targets `trainer.use_v1=false`, preserves the synchronous
full-batch barrier, adds no queue, and performs exactly one update and one publish per logical batch.

### Design & Code Changes

```text
prompt batch
  -> standalone Generator replicas on rollout pools
  -> unchanged reward / log-prob / advantage path
  -> pure Actor worker group on the existing global_pool
  -> exactly one actor update
  -> exactly one synchronous weight publish
```

| Area / File | Change | Why |
| --- | --- | --- |
| diffusion trainer config and generated YAML | add documented `separate: false` | make the behavior opt-in and preserve defaults |
| `main_diffusion.py` | validate early; select pure Actor on the existing resource pool | prevent hidden rollout allocation on Trainer devices |
| `ray_diffusion_trainer.py` | initialize the existing standalone rollout/checkpoint stack; reject LoRA; guard the original colocated sleep call sites | separate resources without changing batch or publish semantics |
| CPU tests | cover invalid configs, all three LoRA entry points, absent/default field compatibility, roles/pools, manager arguments and initial sleep behavior | lock down the new control-flow boundary |
| `docs/examples/config.md` | document scope, constraints and a valid example | make the user-facing flag discoverable and accurate |
| GPU smoke | exercise pure Actor, standalone rollout and NCCL weight publication for two steps | provide a repository-pinned integration gate |

### Compatibility

- Default behavior: unchanged because `actor_rollout_ref.separate=false`.
- Configuration/API: one optional boolean field; unsupported combinations fail before model/worker initialization.
- Supported separate scope: online policy-gradient, `trainer.use_v1=false`, `hybrid_engine=false` and full finetuning.
- LoRA: unsupported in separate mode and rejected before Ray initialization.
- Resources/checkpoint engine: positive rollout resources and a non-`naive` backend are required.
- Checkpoint/data formats: unchanged.
- Breaking change: No.

### API and Usage Example

```bash
bash examples/dancegrpo_trainer/wan22/run_wan22_5b_t2v_hpsv3_auto.sh \
  actor_rollout_ref.separate=true \
  actor_rollout_ref.hybrid_engine=false \
  actor_rollout_ref.rollout.nnodes=1 \
  actor_rollout_ref.rollout.n_gpus_per_node=8 \
  actor_rollout_ref.rollout.checkpoint_engine.backend=nccl \
  trainer.nnodes=1 \
  trainer.n_gpus_per_node=8
```

The existing script forwards trailing Hydra overrides. On a Ray cluster, the values above request one 8-device pure
Actor Trainer group and one 8-device standalone Generator group; model, reward-model and dataset values remain the
script's existing inputs. `actor_rollout_ref.rollout.agent.num_workers` controls CPU request concurrency, not the
number of rollout GPUs.

#### Automated tests

| Command | Result | Coverage |
| --- | --- | --- |
| `python -m pytest -q tests/trainer/diffusion/test_separate_trainer_on_cpu.py tests/workers/config/test_diffusion_config_on_cpu.py tests/special_sanity/test_config_docs.py` | PASS (45 tests) | validation, all three LoRA entry points, roles/pools, manager arguments, initial sleep guard, Hydra config and docs |
| generated-config pre-commit hook in the exact final scope | PASS | generated config freshness |
| `pre-commit run --all-files` | PASS | full repository hooks |
| final-scope patch checks | PASS | patch whitespace/integrity |

#### Pinned-stack GPU smoke

This update adds `tests/special_e2e/run_flowgrpo_qwen_image_separate.sh` and registers it in
`tests/gpu_smoke/run_gpu_smoke_diffusion_e2e.sh`. The smoke exercises pure Actor training, standalone vLLM-Omni
rollout, full-finetuning weight publication through the NCCL checkpoint engine, and two training steps on the
repository-pinned stack.

Status: **Pending GPU CI after this push.**

#### Supplemental NPU validation (not a pinned-stack gate)

These Ascend NPU runs used the repository Verl pin plus a validation-only compatibility backport from upstream Verl
commit `7dc5559a`. They are retained only as supplemental evidence for numerical agreement, policy-version
progression and Trainer/Generator placement. They are not evidence that NPU separate runs on the unmodified
repository-pinned stack.

Environment:

- Hardware: two nodes, 8 Ascend 910B1 NPUs per node;
- placement: 8 Trainer ranks on Node 1 and 8 Generator replicas on Node 2;
- model: Wan2.2 TI2V 5B with HPSv3 reward;
- key config: batch `16`, rollout `n=8`, `128` samples, seed `42`, TP `1`, `10` training steps;
- comparison: two colocated baseline runs and one separate run. All three used the same training and sampling
  parameters, seed, prompt order, and no-resume setting.

| Metric / Invariant | Baseline (run 1) | Baseline (run 2) | Separate | Supplemental observation |
| --- | ---: | ---: | ---: | --- |
| step-1 reward mean | 0.742856741 | 0.742856741 | 0.742856741 | bitwise-identical across all three runs |
| 10-step reward MAE vs. baseline run 1 | — | 0.001318508 | 0.001762307 | same order of magnitude |
| 10-step max absolute reward delta vs. baseline run 1 | — | 0.002978563 | 0.004747331 | both within `±0.005` |
| 10-step Pearson correlation vs. baseline run 1 | — | 0.999618514 | 0.998835137 | highly correlated |
| samples / group size per step | 128 / 8 | 128 / 8 | 128 / 8 | matched |
| policy versions | colocated path | colocated path | `0..10`, once each | progressed as expected |
| Ray Job terminal state | `SUCCEEDED` | `SUCCEEDED` | `SUCCEEDED` | historical run completed |

We ran two baselines and one separate run, using the first baseline as the reference for 10-step reward agreement.
The second baseline has an MAE of `0.001318508`, a maximum absolute delta of `0.002978563`, and a Pearson
correlation of `0.999618514`; the corresponding separate results are `0.001762307`, `0.004747331`, and
`0.998835137`. The two error levels are of the same order of magnitude, and both maximum absolute deltas are below
`0.005`.

image

All three runs have an identical step-1 reward. Their subsequent curves closely overlap without matching point for
point. This shows that repeating the same colocated configuration itself produces small differences, and that the
current gap between separate and baseline is within normal training variation.

Validation dependency note:

- The Verl revision pinned by this repository calls `StatelessProcessGroup(..., socket=listen_socket)`, but the
  installed vLLM API no longer accepts `socket`. For NPU validation only, we backported
  [upstream Verl commit `7dc5559a`](https://github.com/verl-project/verl/commit/7dc5559a05cc1373ca0269b8f3da6290ad61540a),
  which replaces that call with `StatelessProcessGroup.create(..., listen_socket=listen_socket)`. This compatibility
  issue is already fixed upstream; the local backport changed no product code and is not included in this PR.

These NPU runs are supplemental only. The pinned-stack acceptance gate for this update is the GPU smoke above,
which remains pending.

#### Static checks

- `pre-commit run --all-files`: PASS.
- `git diff --check upstream/main...HEAD`: PASS.

### Reviewer Guide

Suggested review order:

1. `verl_omni/trainer/config/diffusion_trainer.yaml` and `validate_separate_config()`;
2. role/resource mapping in `verl_omni/trainer/main_diffusion.py`;
3. standalone rollout initialization and update/publish boundaries in
   `verl_omni/trainer/diffusion/ray_diffusion_trainer.py`;
4. `docs/examples/config.md`;
5. focused control-flow/config tests and the synchronous separate GPU smoke.

- [x] Read the current `CONTRIBUTING.md` and `AGENTS.md`.
- [x] Review the exact 9-file PR scope line by line.
- [x] Independently verify the recorded tests and supplemental NPU evidence.
- [x] Run and confirm the generated-config, shell, selector and pre-commit checks.
- [x] Document the Hydra override usage example and separate-mode constraints.
- [x] Add the pinned-stack GPU smoke and keep its result Pending until current-head CI succeeds.
- [x] Be prepared to explain and maintain the change end to end.


zhtmike:
I am not sure about this. Ideally we follow verl's roadmap for this. Is it better to put custom trainer under recipe? @knlnguyen1802

knlnguyen1802:
Does this PR belong to a bigger PR, currently I don't see any problem that this PR fix or resolve, please explain more

maochiyu1111:
@knlnguyen1802  

Yes, this PR is the first implementation step of RFC #321, rather than a standalone feature or part of one large implementation PR.

The concrete problem addressed here is that the current legacy synchronous diffusion policy-gradient trainer does not provide a supported opt-in path for placing pure Actor workers and standalone rollout/checkpoint workers on separate Ray resources. This makes it impossible to establish a synchronous disaggregated baseline while preserving the existing full-batch training semantics.

Keeping this as a separate first PR lets us validate resource separation and numerical equivalence before introducing the later GAP and other optimizations proposed in RFC #321. It is also different from #315, which targets the Verl v1 asynchronous trainer and queue-based execution.

Regarding placement, I also saw the suggestion about putting this under recipe/. Guidance on whether this synchronous disaggregated baseline should live in the core legacy trainer or as a recipe-specific trainer would be very helpful, but there is no recipe/ yet.

knlnguyen1802 (changes requested):
## Review

**Request changes.** The default colocated path is unchanged, and separate mode correctly reuses standalone `LLMServerManager` (`worker_group=None`) plus `Role.Actor` instead of forking a verl trainer. That is a reasonable RFC #321 PR1 shape, but it is not merge-ready.

### Blocking

1. **LoRA + separate will crash on the first NCCL weight publish.** Separate mode maps online training to `Role.Actor`. `ActorRolloutRefWorker` only sets `self.peft_merge` when `"rollout" in self.role`. Non-naive `update_weights` then does `if actor_has_lora and not self.peft_merge`. Full-finetune (the Wan2.2 run) is fine; `lora_rank > 0` raises `AttributeError`. This PR also uses vanilla `CheckpointEngineManager`; v1 separate-async uses `OmniCheckpointEngineManager` to push LoRA `peft_config` onto standalone replicas. Please fail-closed in `validate_separate_config()` for LoRA, **or** initialize those fields on actor-only workers and use `OmniCheckpointEngineManager`, with a CPU test.

2. **The NPU evidence is not valid against this repo's pin.** The PR body says the 8+8 910B1 run used a local backport of verl `7dc5559a` (`StatelessProcessGroup(..., socket=listen_socket)`). That commit is not in `.github/verl_pin.txt` (`8a694930`); pinned `verl/utils/distributed.py` still passes `socket=listen_socket`. CUDA NCCL does not take that path, but this PR has no GPU smoke on the pin (v1 already has `tests/special_e2e/run_flowgrpo_qwen_image_v1_separate_async.sh`). Please bump the pin, **or** add a pinned-stack GPU smoke and drop the NPU usage claim until the pin includes the fix.

3. **User-facing knob is undocumented, and the usage example 404s.** `actor_rollout_ref.separate` is not in `docs/examples/config.md`. The example launches `examples/dancegrpo_trainer/wan22/run_wan22_5b_t2v_hpsv3_npu.sh`, which is not in the tree (`run_wan22_5b_t2v_hpsv3_auto.sh` is). Please document the flag and point the example at a script that exists.

### Placement

Following up on the earlier thread: this is RFC #321 PR1, and it is not a duplicate of #315. The remaining question is whether a new v0 flag (`actor_rollout_ref.separate`) should land in the legacy trainer vs a v1 hook vs `recipe/`. v1 `trainer.v1.trainer_mode=separate_async` already exists and still builds colocated hybrid replicas on trainer GPUs; this PR's unique topology is pure Actor with no colocated rollout. A thin core flag is fine **if** that distinction is the intent — but later GAP/TSP PRs will keep growing whatever we land here, so we should call it explicitly.

### CI

No `ci` label, so L1 never ran. I can apply it after the next push.

Thanks for the numerical table and for keeping this opt-in / default-off.

knlnguyen1802:
Please check CI @maochiyu1111

maochiyu1111:
> Please check CI @maochiyu1111

@knlnguyen1802 I checked the [GPU CI run](https://github.com/verl-project/verl-omni/actions/runs/32467717583). All three groups fail during Python import, before reaching any test or separate-mode code, with `ValueError: Either a revision or a version must be specified` from `transformers.integrations.hub_kernels` / `kernels.layer.LayerRepository`.

The same failure is present in the [`main` GPU smoke run](https://github.com/verl-project/verl-omni/actions/runs/32354886076) immediately after [#417](https://github.com/verl-project/verl-omni/pull/417) upgraded `kernels` from `0.14.1` to `0.16.0`. The preceding [`main` run](https://github.com/verl-project/verl-omni/actions/runs/32258202700), using `kernels==0.14.1`, passed all three groups. This PR does not modify dependency or GPU preparation files.

Therefore, the new synchronous separate smoke has not reached its test logic yet. I will wait for the baseline dependency compatibility to be fixed on `main`, then update the branch and request a fresh GPU CI run. Please let me know if you would prefer a different next step.

maochiyu1111:
> ## Review
> **Request changes.** The default colocated path is unchanged, and separate mode correctly reuses standalone `LLMServerManager` (`worker_group=None`) plus `Role.Actor` instead of forking a verl trainer. That is a reasonable RFC #321 PR1 shape, but it is not merge-ready.
> 
> ### Blocking
> 1. **LoRA + separate will crash on the first NCCL weight publish.** Separate mode maps online training to `Role.Actor`. `ActorRolloutRefWorker` only sets `self.peft_merge` when `"rollout" in self.role`. Non-naive `update_weights` then does `if actor_has_lora and not self.peft_merge`. Full-finetune (the Wan2.2 run) is fine; `lora_rank > 0` raises `AttributeError`. This PR also uses vanilla `CheckpointEngineManager`; v1 separate-async uses `OmniCheckpointEngineManager` to push LoRA `peft_config` onto standalone replicas. Please fail-closed in `validate_separate_config()` for LoRA, **or** initialize those fields on actor-only workers and use `OmniCheckpointEngineManager`, with a CPU test.
> 2. **The NPU evidence is not valid against this repo's pin.** The PR body says the 8+8 910B1 run used a local backport of verl `7dc5559a` (`StatelessProcessGroup(..., socket=listen_socket)`). That commit is not in `.github/verl_pin.txt` (`8a694930`); pinned `verl/utils/distributed.py` still passes `socket=listen_socket`. CUDA NCCL does not take that path, but this PR has no GPU smoke on the pin (v1 already has `tests/special_e2e/run_flowgrpo_qwen_image_v1_separate_async.sh`). Please bump the pin, **or** add a pinned-stack GPU smoke and drop the NPU usage claim until the pin includes the fix.
> 3. **User-facing knob is undocumented, and the usage example 404s.** `actor_rollout_ref.separate` is not in `docs/examples/config.md`. The example launches `examples/dancegrpo_trainer/wan22/run_wan22_5b_t2v_hpsv3_npu.sh`, which is not in the tree (`run_wan22_5b_t2v_hpsv3_auto.sh` is). Please document the flag and point the example at a script that exists.
> 
> ### Placement
> Following up on the earlier thread: this is RFC #321 PR1, and it is not a duplicate of #315. The remaining question is whether a new v0 flag (`actor_rollout_ref.separate`) should land in the legacy trainer vs a v1 hook vs `recipe/`. v1 `trainer.v1.trainer_mode=separate_async` already exists and still builds colocated hybrid replicas on trainer GPUs; this PR's unique topology is pure Actor with no colocated rollout. A thin core flag is fine **if** that distinction is the intent — but later GAP/TSP PRs will keep growing whatever we land here, so we should call it explicitly.
> 
> ### CI
> No `ci` label, so L1 never ran. I can apply it after the next push.
> 
> Thanks for the numerical table and for keeping this opt-in / default-off.


@knlnguyen1802 Thanks for the detailed review. I addressed the three blocking items in this update:

1. **LoRA + separate:** `validate_separate_config()` now fails closed for `model.lora.rank`, the compatibility `model.lora_rank`, and `model.lora_adapter_path` before Ray initialization. The synchronous separate baseline is explicitly full-finetuning only, with focused CPU coverage for all three LoRA entry points.

2. **Pinned-stack evidence:** I removed the claim that the existing NPU runs validate unmodified repository-pin support. Those runs used a validation-only compatibility backport from Verl `7dc5559a` and are now labeled supplemental numerical/topology evidence only. This update adds a synchronous separate GPU smoke and registers it in `ci-e2e-diffusion` to provide the pinned-stack gate.

3. **Documentation/example:** `actor_rollout_ref.separate` is now documented in `docs/examples/config.md`, including its default, supported scope, full-finetuning restriction, resource/checkpoint constraints, and the fact that `rollout.agent.num_workers` controls CPU request concurrency. The example now points to the existing `run_wan22_5b_t2v_hpsv3_auto.sh` recipe.

The branch is rebased onto the current `main`. Local focused CPU result: `PASS (45 tests)`. `pre-commit run --all-files` also passed.

The GPU smoke is committed and wired into the diffusion GPU group. The requested GPU CI was triggered, but it did not reach any test logic: all three groups failed during Python import with the same `kernels==0.16.0` compatibility regression currently present on `main`.

### Placement

Looking beyond this PR, GAP will introduce a pipelined fit lifecycle with fragment-level scheduling and atomic updates, while TAG will temporarily use Trainer resources for rollout. Although their implementations can live in separate modules, these stages will materially change Trainer behavior.

Given that scope, my preference would be to keep only the reusable pure-Actor disaggregation primitives in core and place the DigenRL-specific Trainer and pipeline under `recipe/`, if that is the intended home for custom trainers. Would you recommend keeping PR1 as the thin core foundation and placing the later GAP/TSP/TAG stages under a new recipe, or would you prefer a v1-based extension instead? The current repository does not yet have a top-level `recipe/`, so guidance on the expected structure would be helpful. I would like to align on that direction before publishing the follow-up PR.

knlnguyen1802:
@maochiyu1111 Please check CI failure and resolve conflict

maochiyu1111:
> @maochiyu1111 Please check CI failure and resolve conflict

@knlnguyen1802  get it, I will fix it ASAP

knlnguyen1802 (approved):

maochiyu1111:
@knlnguyen1802 Thanks for reviewing and merging PR1. I’d like to follow up on the placement of the upcoming DigenRL work before preparing the next PRs.

GAP/TSP/TAG will substantially change the Trainer’s execution model and per-step dataflow, rather than merely adding isolated hooks. For example, the current GAP implementation introduces an alternative training loop with generation-axis pipelining, lane-affine rollout/reward execution, fragment-ready reward/log-prob/advantage/backward processing, and cross-fragment gradient accumulation followed by a single optimizer/scheduler step per logical batch.

Should these features live in a DigenRL-specific Trainer and pipeline under a new `recipe/` directory, rather than continuing to extend the legacy core Trainer? If `recipe/` is preferred, could you also suggest the expected directory structure?

I’d like to align on this before implementing the follow-up PRs. Thanks!

PR #334, opened 2026-08-03
[rollout,tests] fix: preserve request-batched diffusion logprobs

wtomin:
## Summary
- Add `request_batch_needs_logprobs` to decide whether a packed diffusion batch must compute log-probs when any request asks for them.
- Use the helper in Qwen-Image and SD3 FlowGRPO request-batch adapters so packed request order no longer drops `log_probs`.
- Add CPU regression coverage for mixed `logprobs=False/True` packed requests.

## Bug

Bug in `ci-core` ci tests:
```bash
=================================== FAILURES ===================================
________________________________ test_generate _________________________________

init_server = Actor(vLLMOmniHttpServer, a9b65986840129ecc7f1667401000000)

    def test_generate(init_server):
        """Concurrent generate() covering basic output, logprobs, and multi-request correctness."""
        results = _generate_concurrent(init_server, _PROMPTS, logprobs_first_only=True)
    
        for i, output in enumerate(results):
>           _assert_valid_diffusion_output(output, index=i, expect_logprobs=(i == 0))

tests/workers/rollout/rollout_vllm/test_vllm_omni_generate.py:355: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

output = DiffusionOutput(diffusion_output=tensor([[[0.6000, 0.5765, 0.5765,  ..., 0.5765, 0.5765, 0.6039],
         [0.5961, 0....tensor([1, 1, 1, 1, 0, 0]), 'negative_prompt_embeds': None, 'negative_prompt_embeds_mask': None, 'global_steps': None})
index = 0, expect_logprobs = True

    def _assert_valid_diffusion_output(output: DiffusionOutput, *, index: int, expect_logprobs: bool = False) -> None:
        assert isinstance(output, DiffusionOutput), f"Request {index}: expected DiffusionOutput"
        assert len(output.diffusion_output) == 3, f"Request {index}: expected 3 channels (CHW)"
        h, w = len(output.diffusion_output[0]), len(output.diffusion_output[0][0])
        assert h > 0 and w > 0, f"Request {index}: image dimensions must be positive"
        assert output.stop_reason in ("completed", "aborted", None), f"Request {index}: unexpected stop_reason"
        assert 0.0 <= output.diffusion_output[0][0][0] <= 1.0, f"Request {index}: pixel values must be in [0, 1]"
        if expect_logprobs:
            lp = output.log_probs
>           assert lp is not None, f"Request {index}: log_probs should be present when logprobs=True"
E           AssertionError: Request 0: log_probs should be present when logprobs=True
E           assert None is not None

tests/workers/rollout/rollout_vllm/test_vllm_omni_generate.py:343: AssertionError
```

## Root Cause

The failing test submits concurrent diffusion requests with mixed logprobs settings: **only the first logical request has logprobs=True, while the others use False**. Under request-level batching, the rollout adapter previously used `sampling_params_list[0]` from the packed batch to decide whether to compute log-probs for the whole batch. That packed-batch first request is determined by scheduler/batching order, not by the original test order. If a `logprobs=False` request becomes the first packed request, log-prob computation is disabled for the entire batch, so `all_log_probs` is missing even for the request that asked for `logprobs=True`. The server then returns `DiffusionOutput.log_probs=None`, causing the assertion failure.

## Test plan

- [x] `NUM_GPUS=2 bash tests/gpu_smoke/run_gpu_smoke_core.sh` local run passed.
- [x] `python -m pytest tests/pipelines/test_request_batch_sde_windows_on_cpu.py -q`

### DIFF
--- a/tests/pipelines/test_request_batch_sde_windows_on_cpu.py
+++ b/tests/pipelines/test_request_batch_sde_windows_on_cpu.py
@@ -14,11 +14,9 @@
 
 """CPU tests for per-sample SDE window sampling under request packing."""
 
-from types import SimpleNamespace
-
 import torch
 
-from verl_omni.pipelines.request_batch import request_batch_needs_logprobs, sample_per_sample_sde_windows
+from verl_omni.pipelines.request_batch import sample_per_sample_sde_windows
 
 
 def _advanced_generator(seed: int, latent_numel: int = 8) -> torch.Generator:
@@ -51,22 +49,3 @@ def test_per_sample_sde_windows_match_serial_and_ignore_pack_order():
         device="cpu",
     )
     assert packed == [serial[i] for i in packed_order]
-
-
-def test_request_batch_needs_logprobs_ignores_pack_order():
-    sampling_params = [
-        SimpleNamespace(extra_args={"logprobs": False}),
-        SimpleNamespace(extra_args={"logprobs": True}),
-        SimpleNamespace(extra_args={"logprobs": False}),
-    ]
-
-    assert request_batch_needs_logprobs(sampling_params, default=True)
-
-
-def test_request_batch_needs_logprobs_disabled_only_when_all_requests_disable_it():
-    sampling_params = [
-        SimpleNamespace(extra_args={"logprobs": False}),
-        SimpleNamespace(extra_args={"logprobs": False}),
-    ]
-
-    assert not request_batch_needs_logprobs(sampling_params, default=True)
--- a/tests/pipelines/test_request_batch_sde_windows_on_cpu.py
+++ b/tests/pipelines/test_request_batch_sde_windows_on_cpu.py
@@ -14,9 +14,11 @@
 
 """CPU tests for per-sample SDE window sampling under request packing."""
 
+from types import SimpleNamespace
+
 import torch
 
-from verl_omni.pipelines.request_batch import sample_per_sample_sde_windows
+from verl_omni.pipelines.request_batch import request_batch_needs_logprobs, sample_per_sample_sde_windows
 
 
 def _advanced_generator(seed: int, latent_numel: int = 8) -> torch.Generator:
@@ -49,3 +51,22 @@ def test_per_sample_sde_windows_match_serial_and_ignore_pack_order():
         device="cpu",
     )
     assert packed == [serial[i] for i in packed_order]
+
+
+def test_request_batch_needs_logprobs_ignores_pack_order():
+    sampling_params = [
+        SimpleNamespace(extra_args={"logprobs": False}),
+        SimpleNamespace(extra_args={"logprobs": True}),
+        SimpleNamespace(extra_args={"logprobs": False}),
+    ]
+
+    assert request_batch_needs_logprobs(sampling_params, default=True)
+
+
+def test_request_batch_needs_logprobs_disabled_only_when_all_requests_disable_it():
+    sampling_params = [
+        SimpleNamespace(extra_args={"logprobs": False}),
+        SimpleNamespace(extra_args={"logprobs": False}),
+    ]
+
+    assert not request_batch_needs_logprobs(sampling_params, default=True)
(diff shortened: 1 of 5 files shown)


knlnguyen1802 (approved):
LGTM, I think it is fine to follow this rule for batching. One minor thing is that we can add one more test when sampling_params is empty

wtomin:
> LGTM, I think it is fine to follow this rule for batching. One minor thing is that we can add one more test when sampling_params is empty

When sampling_params is empty, there should raise an error, right? I don't know if we should tolerate empty sampling_params. @knlnguyen1802

zhtmike:
> only the first logical request has logprobs=True, while the others use False.

If we enable `logprobs=True`, we expect all requests should be `logprobs=True`. Is it better to fix in the front-end instead of just bypassing it to model core?

wtomin:
> > only the first logical request has logprobs=True, while the others use False.
> 
> If we enable `logprobs=True`, we expect all requests should be `logprobs=True`. Is it better to fix in the front-end instead of just bypassing it to model core?

I think in rollout of RL training, it's like that all samples in the batch have  `logprobs=True`. But for this simple test case in  `tests/workers/rollout/rollout_vllm/test_vllm_omni_generate.py`:

```python

def _generate_concurrent(
    server,
    prompts: list[str],
    *,
    logprobs_first_only: bool = True,
    sampling_overrides: dict[str, Any] | None = None,
) -> list[DiffusionOutput]:
    sampling_overrides = sampling_overrides or {}
    refs = []
    for i, prompt in enumerate(prompts):
        rid = f"test_{i}_{uuid4().hex[:8]}"
        sampling_params = {
            "num_inference_steps": _SMOKE_STEPS,
            "true_cfg_scale": 1.0,
            "height": _SMOKE_SIZE,
            "width": _SMOKE_SIZE,
            "logprobs": (i == 0) if logprobs_first_only else True,
            **sampling_overrides,
        }
        refs.append(
            server.generate.remote(
                prompt_ids=_tokenize_prompt(prompt),
                sampling_params=sampling_params,
                request_id=rid,
            )
        )
    return ray.get(refs, timeout=600)
```

It only sets the first sample to have  "logprobs=True". 

I guess when these samples reach the vllm-omni server, the order of these samples are messed up. They are somehow re-ordered and re-packed, so that the first sample may have "logprobs=False", and that's why the ci-core test failed.

zhtmike:
> > > only the first logical request has logprobs=True, while the others use False.
> > 
> > 
> > If we enable `logprobs=True`, we expect all requests should be `logprobs=True`. Is it better to fix in the front-end instead of just bypassing it to model core?
> 
> I think in rollout of RL training, it's like that all samples in the batch have `logprobs=True`. But for this simple test case in `tests/workers/rollout/rollout_vllm/test_vllm_omni_generate.py`:
> 
> ```python
> def _generate_concurrent(
>     server,
>     prompts: list[str],
>     *,
>     logprobs_first_only: bool = True,
>     sampling_overrides: dict[str, Any] | None = None,
> ) -> list[DiffusionOutput]:
>     sampling_overrides = sampling_overrides or {}
>     refs = []
>     for i, prompt in enumerate(prompts):
>         rid = f"test_{i}_{uuid4().hex[:8]}"
>         sampling_params = {
>             "num_inference_steps": _SMOKE_STEPS,
>             "true_cfg_scale": 1.0,
>             "height": _SMOKE_SIZE,
>             "width": _SMOKE_SIZE,
>             "logprobs": (i == 0) if logprobs_first_only else True,
>             **sampling_overrides,
>         }
>         refs.append(
>             server.generate.remote(
>                 prompt_ids=_tokenize_prompt(prompt),
>                 sampling_params=sampling_params,
>                 request_id=rid,
>             )
>         )
>     return ray.get(refs, timeout=600)
> ```
> 
> It only sets the first sample to have "logprobs=True".
> 
> I guess when these samples reach the vllm-omni server, the order of these samples are messed up. They are somehow re-ordered and re-packed, so that the first sample may have "logprobs=False", and that's why the ci-core test failed.

in this case I prefer to fix the test case directly

wtomin:
I will temporally remove this test case where there are mixed `logprobs=True` and  `logprobs=False` prompts in the same batch. 

If the mixed case is still important, it's always possible to add it on. @SamitHuang

zhtmike:
use a pytest("skip", reason="xxx") or something instead of just commenting out

zhtmike (approved):

Issue #249, opened 2026-07-09
[RFC] Support trainer v1 for verl omni Diffusion model


knlnguyen1802:
### Feature request

This RFC proposes a verl v1-style trainer path for verl-omni diffusion policy-gradient training. The design follows the three upstream verl v1 modes, `sync`, `colocate_async`, and `separate_async`, while keeping diffusion-specific math and data layout inside verl-omni.

The main idea is to reuse generic verl v1 infrastructure where the contract already fits, especially TransferQueue, ReplayBuffer, LLMServerManager, load balancing, checkpoint synchronization, and trainer-mode hook semantics. verl-omni should add thin diffusion adapters around those pieces instead of copying or subclassing the upstream PPO trainer wholesale.

### Motivation

The current diffusion trainer already owns the correct diffusion training semantics: image/video rollout tensors, dense SDE-step log probabilities, image-level rewards, diffusion advantages, actor updates, validation dumping, and diffusion metrics.

Upstream verl v1 adds a useful trainer architecture around asynchronous rollout production and TransferQueue-backed sampling. The useful part for verl-omni is the control-plane design, not the token-level PPO compute path. Directly subclassing upstream `PPOTrainer` would pull in assumptions that do not hold for diffusion, such as token responses, response masks, token rewards, critic values, and text-specific metrics.

## Goals

- Add a v1 trainer architecture for diffusion policy-gradient training.
- Support the three verl v1 trainer modes: `sync`, `colocate_async`, and `separate_async`.
- Reuse upstream verl infrastructure whenever it is generic enough.
- Preserve the existing diffusion `DataProto` compute contract.
- Keep the existing blocking diffusion agent loop path working.
- Add a TransferQueue agent-loop adapter beside the existing path.
- Keep diffusion KV-cache operations as no-op hooks, because pure diffusion models do not have KV cache.

## Non-Goals

- Do not implement `direct_preference` in this RFC.
- Do not add partial denoising-state continuation after abort.
- Do not add token-value or critic paths from upstream PPO v1.
- Do not rewrite upstream verl v1 components that can be reused through their public contracts.

## Architecture

```mermaid
flowchart TB
    subgraph Legend[Legend]
        LR[Reused from verl]
        LO[Existing verl-omni]
        LN[New verl-omni adapter]
    end

    subgraph Entry[Entrypoint and orchestration]
        CLI[main_diffusion_v1.py]
        Runner[DiffusionTaskRunnerV1]
[...]

Issue #221, opened 2026-07-02
# RFC: Fully Async Policy Trainer for VeRL-Omni


knlnguyen1802:
### Feature request

Support fully async policy trainer, mainly benefit for omni model that have autoregressive part

### 1.1 verl `fully_async_policy` (upstream)

Location: `verl/experimental/fully_async_policy/`

| File | Role |
|------|------|
| `fully_async_main.py` | Hydra entry point + `FullyAsyncTaskRunner` (Ray actor) that wires Rollouter, Trainer, MessageQueue |
| `fully_async_rollouter.py` | `FullyAsyncRollouter(SeparateRayPPOTrainer)` — streaming sample-by-sample generation, staleness control, partial rollout |
| `fully_async_trainer.py` | `FullyAsyncTrainer(SeparateRayPPOTrainer)` — consumes samples from MessageQueue, runs PPO fit_step, triggers param sync |
| `message_queue.py` | `MessageQueue` (Ray actor) + `MessageQueueClient` (async wrapper) |
| `detach_utils.py` | `RolloutSample` dataclass, `assemble_batch_from_rollout_samples`, `MetricsAggregator`, `safe_create_task` |
| `config/*.yaml` | Hydra configs extending `ppo_trainer` / `ppo_megatron_trainer` |

Key architecture:
1. **Rollouter** generates sequences **sample by sample** and pushes them into
   a `MessageQueue`.
2. **MessageQueue** is a bounded Ray actor with async producer/consumer
   semantics.
3. **Trainer** pulls `require_batches * ppo_mini_batch_size` samples, runs one
   local fit_step, and after `trigger_parameter_sync_step` local updates
   triggers a **parameter synchronisation** with the Rollouter.
4. **ParameterSynchronizer** = `CheckpointEngineManager` with NCCL backend
   (`verl.checkpoint_engine`).

Both `FullyAsyncRollouter` and `FullyAsyncTrainer` inherit
`SeparateRayPPOTrainer` (→ `RayPPOTrainer`) and reuse its `_fit_*` template
methods, worker init, validation, and checkpoint logic.  They override
`__init__` (do **not** call `super().__init__()`), `init_workers`, and `fit`.

### 1.2 verl-omni diffusion trainer (current)

Location: `verl_omni/trainer/diffusion/`

| File | Role |
|------|------|
| `main_diffusion.py` | Hydra entry point + `TaskRunner` that wires the colocate trainer |
[...]

PR #339, opened 2026-08-03
[rollout, reward, tests] fix: preserve diffusion async semantics

### LINKED ISSUE #343: [Bug] Diffusion async agent loop can publish premature terminal status and lose validation context
## System Info

- Repository: `verl-project/verl-omni`
- Affected revision: current `main` at `8ffe7f2`
- Affected paths: `verl_omni/agent_loop/diffusion_agent_loop.py` and `diffusion_agent_loop_tq.py`
- Local focused-test environment: macOS arm64, Python 3.12.10
- Hardware: not required; this is deterministic async control-plane behavior

## Information

- [x] The problem is in the official diffusion agent-loop implementation.
- [x] It is reproducible with CPU-only fakes; no model weights or dataset are required.

## Tasks

- [x] Diffusion policy-gradient rollout through the asynchronous agent loop and TransferQueue path.

## Reproduction

There are two related lifecycle gaps.

### 1. A prompt can publish `failure` before its sibling sessions settle

Configure `rollout.n = 2` for one prompt. Make session 0 raise immediately and let session 1 finish after a short delay.

The current TransferQueue worker awaits the sessions with the default `asyncio.gather(*tasks)`. When session 0 raises, `gather` propagates that exception immediately but does not wait for the still-running sibling. The exception handler then publishes the prompt-level `failure` status while session 1 is still able to finish and write a trajectory.

A minimal deterministic reproduction is:

```python
events = []

async def fail():
    raise RuntimeError("session failed")

async def finish_later():
    await asyncio.sleep(0.01)
    events.append("sibling settled and wrote output")

try:
    await asyncio.gather(fail(), finish_later())
except RuntimeError:
    events.append("prompt failure published")

await asyncio.sleep(0.02)
assert events == [
    "prompt failure published",
    "sibling settled and wrote output",
]
```

In the real path, ReplayBuffer may observe the terminal failure and clear the prompt group before the late sibling write arrives. This violates the per-prompt lifecycle and can leave a late/orphan trajectory.

### 2. Asynchronous reward requests lose train/validation intent

`generate_sequences()` reads `batch.meta_info["validate"]`, but that value is not carried through `_run_agent_loop()` / `_agent_loop_postprocess()` into `_compute_score()`. The asynchronous reward `DataProto` is therefore created with empty `meta_info` for both training and validation.

With a fake reward worker that records its input, a validation rollout observes:

```python
assert received_data.meta_info == {}
```

so a validation-aware reward implementation cannot distinguish validation from training. The same gap exists when the TransferQueue worker delegates to the base diffusion agent-loop postprocessing path.

## Expected behavior

1. All per-prompt session tasks should settle before exactly one terminal `finished` or `failure` status is published.
2. If any session fails, the prompt should publish `failure` only after successful siblings can no longer write after that terminal state.
3. Asynchronous reward requests should preserve `DataProto.meta_info["validate"]` for both the normal and TransferQueue diffusion paths.

The fix should not change model math, tensors, precision, accelerator kernels, public configuration, or API behavior. A proposed minimal fix and focused CPU regression coverage are in #339. This is related to the broader trainer-v1 lifecycle in #249, but it is a narrower correctness bug rather than a duplicate of that RFC.

WenzheWang:
### What does this PR do?

Fix two related diffusion agent-loop correctness gaps:

- Wait for every per-prompt TransferQueue session to settle before publishing a
  single terminal `finished` or `failure` status. This prevents a successful
  sibling from writing trajectories after ReplayBuffer observes and clears a
  failed group.
- Preserve the train/validation intent in asynchronous reward `DataProto`
  requests, including the TransferQueue path, so validation-aware reward
  implementations can distinguish the two flows.

The fixes share the same diffusion agent-loop lifecycle and are covered by one
focused CPU regression suite, so they form one coherent correctness PR.

Fixes #343. Related context: #249. The current head is synchronized with `main`
after #315; this remains a narrower complementary correctness fix.

AI assistance was used for implementation, tests, and PR preparation. The human
submitter reviewed every changed line, understands the change end-to-end, and
reran the relevant tests.

### Checklist Before Starting

- [x] Search for similar PRs. Queries:
  [open diffusion agent-loop PRs](https://github.com/verl-project/verl-omni/pulls?q=is%3Apr+is%3Aopen+diffusion+agent+loop),
  [exact head](https://github.com/verl-project/verl-omni/pulls?q=is%3Apr+head%3AWenZheWang%3Awenzhe%2Ffix-diffusion-agent-loop-semantics).
  Also inspected #315, #297, and the broader #249 / #221 RFCs; none implements
  this narrow behavior.

### Test

Focused CPU regressions on head `18ae610`, synchronized with `main` at `8ffe7f2`:

```bash
TORCH_COMPILE_DISABLE=1 TORCHINDUCTOR_DISABLE=1 \
  tests/agent_loop/test_diffusion_agent_loop_correctness_on_cpu.py
```

Result: `3 passed, 1 warning in 3.84s`. The warning is Ray's existing
deprecated import warning.

The local macOS environment does not install the Linux-only optional
`vllm_omni` runtime used by the repository CPU workflow. For the focused rerun,
the repository-pinned verl revision (`8a694930`) was used and only unrelated
top-level pipeline auto-registration was isolated; the real agent-loop modules
and all three tests were loaded from the PR tree.

A controlled CPU harness also exercised the real `_run_prompt` and
`_compute_score` implementations three times on exact `main` and commit
`404058f`:

| Revision | `_compute_score` accepts `validate` | Async reward `meta_info` | Validation intent preserved |
|---|---:|---|---:|
| `main` (`8ffe7f2`) | no | `{}` | 0/3 |
| PR #339 (`404058f`) | yes | `{"validate": true}` | 3/3 |

The same attempts reproduced the sibling-lifecycle difference: `main` recorded
a post-terminal trajectory write in 3/3 attempts, while the PR waited for the
successful sibling before publishing `failure` and recorded no late write in
3/3 attempts. This is deterministic CPU control-plane evidence, not a full
model-training or accelerator experiment.

Additional checks on the final tree:

- Relevant-file pre-commit passed: Ruff, Ruff format, mypy, generated trainer
  configs, docs-time, docstring, license, device API, DataProto, test-structure,
  naming, and Python compile hooks.
- `git diff --check` passed.
- The history-only identity cleanup preserved the exact final tree; range-diff
  showed no code change from the reviewed pre-rewrite head.
- Independent review of the final diff returned `safe`.

The lightweight local environment does not contain optional `vllm_omni`, so
repository-wide CPU collection is unavailable there. The focused suite above
directly covers all changed behavior. The GitHub Linux CPU workflow installs
`vllm_omni` and discovers the new `_on_cpu.py` test when a maintainer applies
the `ci` label.

### API and Usage Example

No public API or configuration changes. The exported
`DiffusionAgentLoopWorkerTQ` remains the repository-style decorated Ray
ActorClass; the new `validate` argument is internal, defaulted plumbing.

### Design & Code Changes

- `diffusion_agent_loop_tq.py`: use direct
  `asyncio.gather(..., return_exceptions=True)` session settlement and publish
  one terminal status only after all siblings settle; retain a single
  `@ray.remote DiffusionAgentLoopWorkerTQ` class.
- `diffusion_agent_loop.py`: carry `validate` from batch/trajectory context into
  asynchronous reward `DataProto.meta_info` without forwarding it to concrete
  agent-loop implementations.
- `test_diffusion_agent_loop_correctness_on_cpu.py`: cover sibling settlement,
  terminal-status ordering, and both training and validation reward metadata.
  The direct CPU test uses the same Ray backing-class seam already used in the
  repository's Qwen3-Omni adapter path.

This is hardware-independent control-plane behavior. It does not change model
math, tensor values, precision, accelerator kernels, or numerical-alignment
requirements.

- [x] Apply pre-commit checks. All relevant hooks passed as described above.
- [x] Documentation assessed. No update is required because this changes no
  public API, configuration, or user workflow.
- [x] Add unit or end-to-end tests to CI. The new
  `test_diffusion_agent_loop_correctness_on_cpu.py` file follows the existing
  CPU CI discovery convention, so no workflow change is required.


zhtmike:
What is the bug are you trying to fix? Can you post a issue first?

WenzheWang:
Thanks for the review. The branch is synchronized with `main` at `8ffe7f2f63dfdb5857881fd6b636176e4d12594f`; the tested PR head is `404058f03bb354b3b85180344998cf46759a0827`.

I ran the same controlled CPU probe three times on each exact revision. The requested validation comparison is:

| Revision | `_compute_score` accepts `validate` | Async reward `meta_info` | Validation intent preserved |
|---|---:|---|---:|
| `main` (`8ffe7f2`) | no | `{}` | 0/3 |
| PR #339 (`404058f`) | yes | `{"validate": true}` | 3/3 |

The lifecycle probe also recorded a trajectory write after terminal failure/clear in 3/3 `main` attempts and 0/3 PR attempts. On the PR head the injected exception remains visible, but all sessions settle before the prompt-level `failure` is published.

The focused CPU regression suite on the exact PR head is `3 passed, 1 warning`; full pre-commit, the PR-title validator, a public Ray ActorClass API smoke, and independent review also passed. The exported actor now uses the repository's `@ray.remote` class style, while tests instantiate the undecorated implementation without Ray-private metadata.

Detailed controlled evidence remains in #343: https://github.com/verl-project/verl-omni/issues/343#issuecomment-5175476890

This evidence is limited to deterministic hardware-independent control-plane semantics; it does not claim full-scale GPU/NPU training, production failure frequency, performance, or numerical/precision validation. All review threads are addressed and resolved. The remaining external gates are an approving maintainer review and applying the `ci` label to run the repository Linux CPU workflow.

verl_omni/agent_loop/diffusion_agent_loop_tq.py
```diff
@@ -143,10 +148,22 @@ async def _run_prompt(
...
+                        exc_info=(type(error), error, error.__traceback__),
+                    )
+                status = "failure"
+            else:
+                status = "finished"
+            await tq.async_kv_put(key=uid, partition_id=partition_id, tag={"status": status})
+        except Exception:
+            logger.exception(f"Error in _run_prompt for uid={uid}")
+            if tasks:
+                await _settle_session_tasks(tasks)
```
knlnguyen1802: I think this can simply fix by just add `return_exceptions=True` to the `asyncio.gather`. This modification doesnt need this long

knlnguyen1802 (commented):
Thank you for identify 2 errors and fix it, but left some comment

verl_omni/agent_loop/diffusion_agent_loop_tq.py
WenzheWang: Thanks — simplified in `7a66216`.

The helper is removed. `_run_prompt` now directly uses `asyncio.gather(*tasks, return_exceptions=True)`, derives `failure` from the returned exceptions, and publishes one terminal status after all sessions settle. The outer exception path uses the same direct gather only to finish any siblings that were already created before an unexpected setup error.

Validation on the synchronized head:

- focused CPU suite: `3 passed, 1 warning`;
- controlled failure injection: 3/3 runs had no post-terminal sibling write and preserved validation metadata;
- full pre-commit and PR-title validation passed;
- independent final review: `safe`.

tests/agent_loop/test_diffusion_agent_loop_correctness_on_cpu.py
```diff
@@ -0,0 +1,122 @@
...
+    assert sibling_settled.is_set()
+    assert all(task.done() for task in tasks)
+    assert len(errors) == 1
+    assert isinstance(errors[0], RuntimeError)
+
+
+@pytest.mark.asyncio
+async def test_run_prompt_publishes_failure_after_siblings_settle(monkeypatch):
+    worker_cls = DiffusionAgentLoopWorkerTQ.__ray_metadata__.modified_class
+    worker = object.__new__(worker_cls)
```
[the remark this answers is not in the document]
WenzheWang: Addressed in `7a66216`.

The undecorated implementation is now `DiffusionAgentLoopWorkerTQImpl`, while the existing exported actor entry point remains `DiffusionAgentLoopWorkerTQ = ray.remote(DiffusionAgentLoopWorkerTQImpl)`. The regression test instantiates the implementation class directly and no longer accesses `__ray_metadata__` or another Ray-private attribute.

The focused CPU suite passes (`3 passed, 1 warning`), the public remote wrapper import smoke passed, and full pre-commit passed.

verl_omni/agent_loop/diffusion_agent_loop_tq.py
```diff
@@ -255,6 +263,9 @@ async def _write_trajectory_to_tq(
         )
 
 
+DiffusionAgentLoopWorkerTQ = ray.remote(DiffusionAgentLoopWorkerTQImpl)
```
zhtmike: not a good practice to declare global variable here.

either revert to the previous decorator or put it under __init__ or somewhere locally

knlnguyen1802:
Could you provide some compare between main branch on validate and this branch ?

WenzheWang:
Thanks for asking. I ran the same controlled CPU validation probe three times on exact current `main` and the current PR head:

| Revision | `_compute_score` accepts `validate` | Async reward `DataProto.meta_info` | Validation intent preserved |
|---|---:|---|---:|
| `main` (`8ffe7f2`) | no | `{}` | 0/3 |
| PR #339 (`404058f`) | yes | `{"validate": true}` | 3/3 |

The PR threads the validation flag from the rollout context into `_compute_score` and places it in the asynchronous reward request. The focused CPU suite passes (`3 passed, 1 warning`), and full pre-commit passes on `404058f`.

This comparison proves the hardware-independent validation metadata route only. It is not a full model-training, accelerator, performance, or numerical-alignment comparison.

verl_omni/agent_loop/diffusion_agent_loop_tq.py
WenzheWang: Addressed in `404058f` by restoring the repository's decorator-style actor declaration:

```python
@ray.remote
class DiffusionAgentLoopWorkerTQ(DiffusionAgentLoopWorkerTQImpl):
    ...
```

The exported manager entry point remains `DiffusionAgentLoopWorkerTQ`. The undecorated implementation is retained only as the direct unit-test seam, so the regression test does not depend on Ray-private `__ray_metadata__`. I also swept the submitted production path and confirmed that the module-level `DiffusionAgentLoopWorkerTQ = ray.remote(...)` assignment is gone.

Validation on the exact head: focused CPU suite `3 passed, 1 warning`; public Ray ActorClass API smoke passed; full pre-commit and PR-title validation passed; independent review returned `safe`.

verl_omni/agent_loop/diffusion_agent_loop_tq.py
```diff
@@ -255,6 +263,11 @@ async def _write_trajectory_to_tq(
         )
 
 
+@ray.remote
+class DiffusionAgentLoopWorkerTQ(DiffusionAgentLoopWorkerTQImpl):
+    """Ray actor wrapper for the TransferQueue-backed diffusion worker."""
```
knlnguyen1802: I think it's not necessary to make a separate `DiffusionAgentLoopWorkerTQImpl` you can just keep `DiffusionAgentLoopWorkerTQ`

knlnguyen1802 (approved):
LGTM now, thanks for the fix

verl_omni/agent_loop/diffusion_agent_loop_tq.py
WenzheWang: Updated in `18ae610`. The separate `DiffusionAgentLoopWorkerTQImpl` has been removed; the implementation now stays in the single decorator-style `@ray.remote DiffusionAgentLoopWorkerTQ` class, preserving the exported Ray ActorClass API and manager behavior. The CPU regression uses the backing-class seam already present in the repository's Qwen3-Omni adapter path. Focused result: `3 passed, 1 warning`; all relevant pre-commit hooks passed, and the final independent review returned `safe`.

zhtmike (approved):

PR #311, opened 2026-07-27
[diffusion, cfg, tests] feat: save rollout and validation videos as mp4

NancyFyong:
### What does this PR do?

Make the diffusion FlowGRPO trainer save **video** rollouts, not just images.
`_dump_generations` and `_maybe_log_val_generations` assumed image tensors
`[N, C, H, W]` and permuted them directly, which raised on 5-D video batches
`[N, T, C, H, W]`. This branches on tensor rank so video runs write one `.mp4`
per sample (via `diffusers.export_to_video`) while the image path stays
byte-for-byte, and wandb receives an encoded mp4 file path instead of raw
frames (avoiding the heavy `moviepy` dependency).

It also fixes a color-inversion bug: `diffusers.export_to_video` rescales NumPy
input by 255 even when it is already `uint8 [0, 255]`, overflowing to a
photographic negative (`255 - x`). Frames now go through
`verl_omni.utils.media.video_tensor_to_pil_frames`, which hands `export_to_video`
PIL images (not rescaled) and sanitizes non-finite decoder output before
quantization.

### Checklist Before Starting

- [x] Search for similar PRs:
  - https://github.com/verl-project/verl-omni/pulls?q=is%3Apr+dump_generations (0 results)
  - https://github.com/verl-project/verl-omni/pulls?q=is%3Apr+rollout+video
  - https://github.com/verl-project/verl-omni/pulls?q=is%3Apr+diffusion+video+mp4 (0 results)

  No existing PR touches `_dump_generations` / `_maybe_log_val_generations` or the
  video-vs-image dump branching. Nearby diffusion PRs (#278 LTX2.3 t2av pipeline,
  #300 distillation losses, #172 rollout server routing) address unrelated scope.
- [x] PR title formatted as `[{modules}] {type}: {description}`.

### Test

All new tests are pure-CPU and run in seconds (`diffusers` + `imageio`/ffmpeg only).

```bash
python -m pytest -q \
  tests/utils/test_media_on_cpu.py \
  tests/trainer/diffusion/test_dump_generations_video_on_cpu.py \
  tests/trainer/diffusion/test_val_video_logging_on_cpu.py
# -> 10 passed
```

- `test_media_on_cpu.py` — exact uint8 pixel values, NaN/±inf/clip handling, shape rejection.
- `test_dump_generations_video_on_cpu.py` — one `.mp4` per video sample, one `.jpg` per image
  sample (unchanged), `max_samples` bound, and a decode-and-check that warm colors are **not**
[...]

PR #98, opened 2026-05-20
[diffusion, model, reward] feat: support Wan22 DanceGRPO text-to-video training

ZihaoW123:
### What does this PR do?

support Wan22 DanceGRPO text-to-video training using HPSv3 reward model

### Checklist Before Starting

- [x] Search for similar PRs. Paste at least one query link here: ...
- [x] Format the PR title as `[{modules}] {type}: {description}` (This will be checked by the CI)
  - `{modules}` include `fsdp`, `vllm_omni`, `rollout`, `trainer`, `ci`, `training_utils`, `recipe`, `ray`, `worker`, `single_controller`, `misc`, `perf`, `model`, `algo`, `env`, `tool`, `ckpt`, `doc`, `data`, `cfg`, `reward`, `diffusion`, `omni`, `tests`, `docker`
  - If this PR involves multiple modules, separate them with `,` like `[diffusion, doc]`
  - `{type}` is in `feat`, `fix`, `refactor`, `chore`, `test`
  - If this PR breaks any API (CLI arguments, config, function signature, etc.), add `[BREAKING]` to the beginning of the title.
  - Example: `[BREAKING][diffusion, fsdp] feat: new rollout scheduler`
  
### Changes
- **Data**: `wan22_hpsv3.py` converts a prompt file (one prompt per line) into train/test.parquet.
- **Reward**: `hpsv3_reward` provides the computational model for HPSv3 (Towards Wide-Spectrum Human Preference Score).
- **Wan22 Adapter**: Wan22DanceGRPOPipelineWithLogProb and Wan22DanceGRPO
- **sde_type**: Add the SDE type `dance_sde` to FlowMatchSDEDiscreteScheduler in `flow_match_sde.py`.

#### Examples & docs

- `examples/dancegrpo_trainer/README.md`, `wan22_hpsv3.py`, `run_wan22_5b_t2v_hpsv3_npu.sh`

### Usage
Generate datasets:

```bash
python3 examples/dancegrpo_trainer/data_process/wan22_hpsv3.py \
  --input_path $WORKSPACE/data/hpsv3/prompt.txt \
  --output_dir $WORKSPACE/data/hpsv3
```

Train:

```bash
bash examples/dancegrpo_trainer/run_wan22_5b_t2v_hpsv3_npu.sh
```

### Test

After training on a set of dataset (~ 1200 prompts), the rewrds/mean and val-core/reward/mean@1 went as expected:

wan22_dance_grpo_hpsv3_npu_rewards_mean
[...]

PR #340, opened 2026-08-04
[trainer, tests] fix: normalize Wan22 channels-last 6-D video before …

MichaelSou1:
…dumping

BaseRayDiffusionTrainer._dump_generations only handled 4-D image batches [N, C, H, W] and 5-D channels-first video [N, T, C, H, W] (added in #311). Wan22 DanceGRPO rollout (#98) returns channels-last video with a singleton frame dim, [N, 1, F, H, W, C], which fell through to the image branch and raised (the path is only reachable with trainer.validation_data_dir set, which the official NPU recipe never sets, so it was not caught earlier).

Squeeze the singleton dim and permute to channels-first before the existing rank check, so both the video and image branches stay unchanged for their original input shapes. Add a CPU regression test covering frame count and channel order for the 6-D input.

Test: pytest -q tests/trainer/diffusion/test_dump_generations_video_on_cpu.py -> 5 passed, 16 warnings in 16.71s (run on a GPU host with the full verl-omni dependency stack, since vllm_omni has no macOS/arm64 wheel). Also verified with ruff check / ruff format --check on both changed files.

AI assistance (Claude, via CatPaw) was used for this change.

### What does this PR do?

> Add **concise** overview of what this PR aims to achieve or accomplish. Reference related GitHub issues and PRs that help with the review.

### Checklist Before Starting

### Test

### API and Usage Example

```python
```

### Design & Code Changes

> [!IMPORTANT]

### DIFF
--- a/tests/trainer/diffusion/test_dump_generations_video_on_cpu.py
+++ b/tests/trainer/diffusion/test_dump_generations_video_on_cpu.py
@@ -126,3 +126,25 @@ class TestDumpGenerations:
         rows = _read_jsonl(tmp_path)
         assert len(rows) == 1
         assert rows[0]["reward"] == 0.0
+
+    def test_wan22_channels_last_singleton_video_batch(self, tmp_path):
+        """Wan22 rollout returns ``[N, 1, F, H, W, C]``; the dump must normalize it
+        to channels-first video (one mp4 per sample, frames and channel order intact)."""
+        imageio = pytest.importorskip("imageio")
+
+        outputs = torch.zeros(2, 1, 8, 16, 16, 3)  # [N, 1, F, H, W, C], solid warm frame
+        outputs[..., 0] = 0.75  # R
+        outputs[..., 1] = 0.35  # G
+        outputs[..., 2] = 0.15  # B
+        _dump(tmp_path, outputs)
+
+        visual = os.path.join(str(tmp_path), "0")
+        mp4s = sorted(f for f in os.listdir(visual) if f.endswith(".mp4"))
+        assert mp4s == ["0.mp4", "1.mp4"]
+
+        reader = imageio.get_reader(os.path.join(visual, "0.mp4"))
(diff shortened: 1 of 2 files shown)


zhtmike:
What bug are you want to solve? Kindly post the issue first

SamitHuang:
@NancyFyong  Can you PTAL?

NancyFyong:
Hi @MichaelSou1
Thanks for the fix. Could you clarify where the 6-D shape `[N, 1, F, H, W, C]` is
  introduced?


  Is the Wan2.2 rollout single-sample output already `[1, F, H, W, C]` before the agent
  loop adds the trainer batch dimension, meaning the extra `1` is `num_outputs_per_prompt`?
  If so, should this be normalized at the rollout/agent-loop boundary rather than only
  inside `_dump_generations`, since other consumers of `batch["responses"]` may also see
  the 6-D channels-last tensor?


  Also, should the new `squeeze(1)` be guarded with `outputs.shape[1] == 1` so that
  non-singleton 6-D outputs fail with a clearer error?

zhtmike:
I think we also need to cc @ZihaoW123

wtomin:
Why don't you edit `Wan22DanceGRPOPipelineWithLogProb` so that it squeezes `[N, 1, F, H, W, C]` to `[N, F, H, W, C]`? I don't think it is suitable to make an exception for wan22 in `ray_diffusion_trainer.py` as it may affect other pipelines too.

NancyFyong:
> Why don't you edit `Wan22DanceGRPOPipelineWithLogProb` so that it squeezes `[N, 1, F, H, W, C]` to `[N, F, H, W, C]`? I don't think it is suitable to make an exception for wan22 in `ray_diffusion_trainer.py` as it may affect other pipelines too.

Regarding the issue of dimensions, I believe it will be resolved through the RFC https://github.com/verl-project/verl-omni/issues/403 process.

PR #351, opened 2026-08-06
[rollout] fix: fall back attention on unsupported GPUs

jiefangxuanyan:
Description to be added.


jiefangxuanyan:
Closing because the original SM120 smoke-test failure is now handled by the attention-backend selection logic on upstream main; the current test passes there without this server-side fallback.

Issue #342, opened 2026-08-04
[RFC] Making Nsight Systems Profiling Feasible in VeRL-Omni


jiefangxuanyan:
# RFC: Making Nsight Systems Profiling Feasible in VeRL-Omni

I previously presented my profiling findings on the `old_log_prob` and `update_actor` phases at the weekly meeting. The community suggested that I organize the profiling approach more systematically.

The existing [`docs/perf/profiler.md`](docs/perf/profiler.md) already provides some guidance on profiling with Nsight Systems. However, for a large and composite project such as VeRL-Omni, oversized profiling results are a common problem. An oversized result may fail during report finalization with an error about non-monotonic timestamps. It may also produce a `.nsys-rep` file so large that the Nsight Systems GUI runs out of memory or becomes unresponsive.

To prevent profiling results from becoming too large, profiling needs to be restricted along three dimensions: processes, collection features, and time.

## Limiting Processes

VeRL-Omni uses Ray to manage processes. After Nsight Systems profiling is enabled for the controller and workers, child processes inherit the parent's Nsight configuration by default.

The information needed for profiling usually exists in only a small number of processes, such as the controller process that carries NVTX ranges and the worker processes that execute CUDA calls. Nsight collection in other Ray workers not only creates additional reports that obscure the intended profiling targets, but can also slow down process execution. This overhead may distort the behavior shown in the profiling result.

During an earlier experiment, I monkey-patched Ray to prevent child processes from carrying their parent's Nsight options. This method was effective, but I encountered substantial difficulties when trying to turn the same principle into an engineering solution.

Without the monkey patch, the default behavior has to be changed by adding restrictions at Ray actor and task creation sites throughout the framework. However, one of Ray's advantages is that new actors and tasks can be created flexibly from almost anywhere in the program. This also makes these creation sites difficult to control. I tried using AI tools to modify them, but found that creation sites were frequently missed or modified incorrectly. Some sites also sit deep in the call hierarchy, requiring a parameter to be passed through several layers before the behavior can be changed.

Fortunately, Nsight Systems provides other control mechanisms, as discussed in the section on limiting time. An `nsys` process that never starts an actual capture is effectively a no-op: it does not generate an additional report, and its performance impact is limited.

[...]

PR #352, opened 2026-08-06
[trainer, doc] feat: add controller Nsight profiling controls

jiefangxuanyan:
### What does this PR do?

While profiling Qwen-Image FlowGRPO, I observed GPU-idle intervals at the boundaries of each `old_log_prob` and `update_actor` phase. To understand what the CPU and Python control flow were doing during those intervals, I needed an Nsight Systems configuration that could collect Python stacks without producing oversized reports.

This PR implements part of the profiling approach discussed in [#342](https://github.com/verl-project/verl-omni/issues/342):

- limit the collected Nsight Systems features to the signals needed for this investigation; and
- limit the controller process capture window by reusing the same `cudaProfilerApi` start/stop mechanism already used for worker processes.

Python stack collection uses fixed-frequency sampling, so its data volume grows with elapsed time. The example therefore captures only three training steps in one continuous window and profiles ActorRollout rank 0 by default. This bounds both the capture duration and the number of Nsight processes that actually enter a capture range, avoiding an unnecessary number of large reports.

The PR also adds a Qwen-Image FlowGRPO profiling recipe focused on correlating controller and actor Python stacks with the `old_log_prob` and `update_actor` NVTX ranges.

This is intentionally a partial implementation of #342. Preventing child processes from inheriting Nsight options and avoiding GPU Metrics lock contention remain under discussion and are out of scope. As discussed in the issue, those concerns are shared with upstream `verl` and should be addressed there with a general process-scope rule. This recipe does not enable GPU Metrics.

### Checklist Before Starting

- [x] Search for similar PRs. Single-keyword queries: [`profile`](https://github.com/verl-project/verl-omni/pulls?q=is%3Apr+is%3Aopen+profile), [`profiler`](https://github.com/verl-project/verl-omni/pulls?q=is%3Apr+is%3Aopen+profiler), [`profiling`](https://github.com/verl-project/verl-omni/pulls?q=is%3Apr+is%3Aopen+profiling), [`nsys`](https://github.com/verl-project/verl-omni/pulls?q=is%3Apr+is%3Aopen+nsys), and [`nsight`](https://github.com/verl-project/verl-omni/pulls?q=is%3Apr+is%3Aopen+nsight). `profiling` and `nsight` returned only this PR; `profiler` and `nsys` returned no open PRs. The broader `profile` query additionally returned [#329](https://github.com/verl-project/verl-omni/pull/329), an unrelated multi-turn agent trajectory PR whose match comes from profiling timing text in its test output and whose changed files do not overlap this work. No separate open implementation was found.

### Test

Full CPU test suite:

```bash
CUDA_VISIBLE_DEVICES="" \
TORCH_COMPILE_DISABLE=1 \
TORCHINDUCTOR_DISABLE=1 \
pytest -s -x --asyncio-mode=auto -o "python_files=*_on_cpu.py" tests/
```

Single-GPU vLLM-Omni generation and Diffusers FSDP smoke tests:

```bash
CUDA_VISIBLE_DEVICES=0 NUM_GPUS=1 \
pytest -s \
  tests/workers/rollout/rollout_vllm/test_vllm_omni_generate.py::test_generate_request_level_batch \
  tests/workers/test_diffusers_fsdp_engine.py
```

Single-GPU SD3.5 offline-DPO smoke test:

```bash
CUDA_VISIBLE_DEVICES=0 \
NUM_GPUS=1 \
TOTAL_TRAIN_STEPS=1 \
NUM_PAIRS=1 \
bash tests/special_e2e/run_offline_dpo_sd35.sh \
  actor_rollout_ref.model.attn_backend=native \
  actor_rollout_ref.rollout.rollout_attn_backend=TORCH_SDPA
```

All tests above passed.

The new profiling recipe was also validated on one node with 8 NVIDIA H100 GPUs:

```bash
bash examples/flowgrpo_trainer/qwen_image/run_qwen_image_ocr_8x80g_fsdp2_benchmark_nsys.sh
```

The run completed successfully and produced exactly two `.nsys-rep` files: one for the Ray TaskRunner controller and one for ActorRollout rank 0, as intended by the default configuration.

image

Full pre-commit checks passed:

```bash
pre-commit run --all-files --show-diff-on-failure --color=always
```

### API and Usage Example

Existing configurations keep their current behavior. Controller capture-range control is opt-in: it is enabled only when the controller Nsight options explicitly select `capture-range=cudaProfilerApi`.

Run the included single-node Qwen-Image recipe with:

```bash
bash examples/flowgrpo_trainer/qwen_image/run_qwen_image_ocr_8x80g_fsdp2_benchmark_nsys.sh
```

The underlying controller options can also be enabled in another recipe with:

```bash
+global_profiler.global_tool_config.nsys.controller_nsight_options.capture-range=cudaProfilerApi \
+global_profiler.global_tool_config.nsys.controller_nsight_options.capture-range-end=null \
+global_profiler.global_tool_config.nsys.controller_nsight_options.kill=none
```

When `capture-range-end` is null, VeRL-Omni derives the required `repeat-shutdown` count from `global_profiler.steps` and `global_profiler.profile_continuous_steps`. An explicitly configured value is preserved.

### Design & Code Changes

1. Resolve controller Nsight options before launching the Ray TaskRunner. The existing controller options remain unchanged unless `capture-range=cudaProfilerApi` is explicitly configured.
2. Start controller profiling before worker profiling and stop it after worker profiling, so their capture windows cover the same profiled steps. If worker startup fails, the controller capture is stopped before propagating the exception.
3. Add CPU coverage for capture-range counting, default-behavior compatibility, explicit configuration preservation, start/stop ordering, disabled steps, and startup exception safety.
4. Add a single-node Qwen-Image FlowGRPO Nsight recipe. It captures three continuous steps, selects ActorRollout rank 0 by default, samples Python stacks at 20 Hz, and keeps native CPU sampling, context-switch collection, CUDA API/kernel software tracing, Python GIL events, and OS-runtime events disabled. It collects and validates the targeted Ray Nsight reports under a unique run ID.
5. Document controller capture-range configuration, report-volume tradeoffs, steady-state step selection, and multi-report alignment.

AI assistance (OpenAI Codex) was used to implement and document this change. The profiling experiment was validated by the human submitter on 8 NVIDIA H100 GPUs. The human submitter has reviewed every changed line and understands the change end to end.

> [!IMPORTANT]

- [x] Add unit or end-to-end test(s) to [the CI workflow](https://github.com/verl-project/verl-omni/tree/main/.github/workflows) to cover all the code. The new `*_on_cpu.py` test is automatically collected by the existing `cpu_unit_tests` workflow when the PR is labeled `ci`; no workflow-file change is required.


docs/perf/profiler.md
```diff
@@ -145,13 +145,59 @@ actor_rollout_ref.actor.profiler.tool=nsys
...
 Ray). Open them with `nsys-ui`.
 
+### 4a. Continuous `old_log_prob` and `update_actor` timeline
+
+Use the dedicated Qwen-Image recipe to inspect Python control flow while the
+controller waits for old-log-prob inference or an actor update:
+
+```bash
+bash examples/flowgrpo_trainer/qwen_image/run_qwen_image_ocr_8x80g_fsdp2_benchmark_nsys.sh
+```
```
knlnguyen1802, line 182: Better if can add requirement for `nvtx` at this docs

examples/flowgrpo_trainer/qwen_image/run_qwen_image_ocr_fsdp2_benchmark.sh
```diff
@@ -0,0 +1,26 @@
+#!/bin/bash
+# Qwen-Image full-weight RL throughput benchmark for 8 x 80 GB GPUs.
```
knlnguyen1802, line 2: We can make it more flexible, it can run in 4 cards so doesnot really require 8 cards, can fix the filename

examples/flowgrpo_trainer/qwen_image/run_qwen_image_ocr_8x80g_fsdp2_benchmark_nsys.sh
```diff
@@ -0,0 +1,284 @@
...
+    global_profiler.global_tool_config.nsys.controller_nsight_options.trace='"nvtx"' \
+    global_profiler.global_tool_config.nsys.controller_nsight_options.cuda-memory-usage='"false"' \
+    +global_profiler.global_tool_config.nsys.controller_nsight_options.capture-range=cudaProfilerApi \
+    +global_profiler.global_tool_config.nsys.controller_nsight_options.capture-range-end='"repeat:1:async"' \
+    +global_profiler.global_tool_config.nsys.controller_nsight_options.kill=none \
+    +global_profiler.global_tool_config.nsys.controller_nsight_options.sample=none \
+    +global_profiler.global_tool_config.nsys.controller_nsight_options.cpuctxsw=none \
+    +global_profiler.global_tool_config.nsys.controller_nsight_options.python-sampling='"true"' \
+    +global_profiler.global_tool_config.nsys.controller_nsight_options.python-sampling-frequency="$PYTHON_SAMPLE_HZ" \
+    +global_profiler.global_tool_config.nsys.controller_nsight_options.discard-environment='"true"' \
```
knlnguyen1802: Could you add in the docs descript these flag and common pitfall, in my test my cuda version does not support `discard-environment`

jiefangxuanyan: `discard-environment` is there because the `.nsys-rep` file can include environment variables such as `HF_TOKEN`. I’ll remove the flag and add a warning in the docs instead. I’ll also document these flags and the common pitfalls.

examples/flowgrpo_trainer/qwen_image/run_qwen_image_ocr_fsdp2_benchmark.sh
jiefangxuanyan: 4 GPUs only work with 141 GB cards for full fine-tuning. 4 × 80 GB will OOM, and 80 GB cards are what I can access more easily. Should I just use a more generic filename, or add a separate LoRA config that works on 4 × 80 GB?

docs/perf/profiler.md
jiefangxuanyan: Do we need to add `nvtx` to `pyproject.toml`, or is mentioning it in the docs enough?

examples/flowgrpo_trainer/qwen_image/run_qwen_image_ocr_fsdp2_benchmark.sh
knlnguyen1802: I think you can just rename the file name and still put default value is 8 GPU, the comment can keep as now

docs/perf/profiler.md
knlnguyen1802: I think in docs is enough

docs/perf/profiler.md
```diff
@@ -145,13 +145,59 @@ actor_rollout_ref.actor.profiler.tool=nsys
 When `global_profiler.tool=nsys` and `steps` is non-empty, the FlowGRPO
 entrypoint launches the Ray TaskRunner under `nsys` using the
 `controller_nsight_options` from `global_profiler.global_tool_config.nsys`.
-Workers are launched with `worker_nsight_options`, including the required
-`capture-range: cudaProfilerApi` flag.
+Workers use `capture-range: cudaProfilerApi`, and the trainer starts and stops
+their collection around the configured steps. The controller records the full
```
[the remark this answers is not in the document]
jiefangxuanyan: Updated the documentation in `afa4653` to explicitly scope this behavior to the legacy `python -m verl_omni.trainer.main_diffusion` entrypoint. It now also notes that v1 does not yet implement the step-based profiling lifecycle, so controller `capture-range: cudaProfilerApi` is not supported there.

verl_omni/trainer/main_diffusion.py
```diff
@@ -31,6 +31,26 @@
...
+def _resolve_controller_nsight_options(config) -> dict:
+    """Resolve controller Nsight options for the configured profiling steps."""
+    nsight_options = OmegaConf.to_container(config.global_profiler.global_tool_config.nsys.controller_nsight_options)
+    if nsight_options.get("capture-range") == "cudaProfilerApi" and nsight_options.get("capture-range-end") is None:
+        capture_count = _count_controller_capture_ranges(
+            OmegaConf.select(config, "global_profiler.steps"),
+            OmegaConf.select(config, "global_profiler.profile_continuous_steps", default=False),
+        )
+        nsight_options["capture-range-end"] = f"repeat-shutdown:{capture_count}"
+    return nsight_options
```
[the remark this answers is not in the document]
jiefangxuanyan: `main_diffusion_v1` does not currently implement the step-based profiling lifecycle driven by `global_profiler.steps`, including coordinated start/stop control for the controller and workers. Reusing `_resolve_controller_nsight_options()` alone would set `capture-range-end`, but no corresponding capture range would be started. The recipe in this PR uses the legacy `main_diffusion` entrypoint, so I scoped the documentation explicitly instead of adding partial v1 support. Full v1 profiling support should be handled separately.

examples/flowgrpo_trainer/qwen_image/run_qwen_image_ocr_fsdp2_benchmark.sh
jiefangxuanyan: Done in `afcc2d9`.

docs/perf/profiler.md
jiefangxuanyan: I took a closer look and found that this is already documented at the beginning of Section 4:

> Nsight requires `nsys` to be installed on every node and the `nvtx` Python package available in the training environment (`pip install nvtx`).

examples/flowgrpo_trainer/qwen_image/run_qwen_image_ocr_8x80g_fsdp2_benchmark_nsys.sh
jiefangxuanyan: Done in `afcc2d9`.

knlnguyen1802 (approved):
LGTM, cc @SamitHuang

zhtmike (approved):

docs/perf/profiler.md
SamitHuang, line 1: It'd be a helpful reference to attach example profiling results/screenshot in the doc

SamitHuang (approved):
LGTM overall. Nice work. Pls improve the doc @jiefangxuanyan

docs/perf/profiler.md
jiefangxuanyan: Can we reuse the screenshot from the PR description and describe in the text what information we should be able to gather from this profiling run?

SamitHuang: sure

jiefangxuanyan: Done in 9551760.

knlnguyen1802:
@jiefangxuanyan Please fix the CI thanks

jiefangxuanyan:
The failure comes from `transformers==5.15.0` changing Qwen3-Omni's `_no_split_modules`, not from this PR. I capped `transformers<=5.14.1`; the failing test now passes. I will push the fix and rerun CI.

jiefangxuanyan:
The two failures appear to be an unrelated flaky port-allocation race. Both failed while torch.distributed.TCPStore was binding dynamically selected diffusion MASTER_PORTs (33531 and 46399) with EADDRINUSE. The current code releases the reservation socket before AsyncOmni/TCPStore binds, so another concurrently running process can claim the port in between. This code is unchanged by this PR and matches upstream main.

Since this is intermittent, it will likely not reproduce on a rerun. @knlnguyen1802, could you please trigger the CI again?

PR #356, opened 2026-08-06
[rollout] feat: support prompt embedding cache for diffusion rollouts

Sky-Trigger:
### What does this PR do?

This PR integrates vLLM-Omni's prompt embedding cache with verl-omni diffusion rollouts.

vLLM-Omni applies the cache wrapper around `encode_prompt()`. Previously, several
verl-omni adapters converted pre-tokenized prompt IDs and attention masks to tensors
before that boundary. Those tensor inputs bypassed the list-based cache-key path, so
repeated prompts were encoded again even when the cache was enabled.

This PR keeps two execution paths. With prompt caching disabled, the existing
tokenizer-to-tensor path is retained unchanged. With prompt caching enabled,
token IDs and masks remain list-like until `encode_prompt()`, where the existing
tensor conversion occurs after the wrapper has built its cache key. It covers:

- Qwen-Image FlowGRPO and MixGRPO (through the shared FlowGRPO adapter)
- Qwen-Image DiffusionNFT
- Qwen-Image online DPO
- Wan2.2 DanceGRPO

The adapters continue to accept explicit tensor inputs. We verified that both
paths preserve the output produced after `encode_prompt()` relative to the
pre-change implementation: prompt embeddings and masks retain the same values,
device, dtype, and batch/output contract. Precomputed prompt embeddings
continue to bypass token encoding.

The cache is disabled by default, so existing configurations do not enable it
implicitly.

For repeated rollouts, this PR also provides an opt-in routing-affinity mode.
When both the cache and affinity are enabled, all repeated generations derived
from one trainer sample use that sample's stable `uid` as the rollout request
ID. The load balancer can then route them to the same rollout replica, where
they share its worker-local cache. Without affinity, each generation uses a new
UUID and may be distributed across replicas; each replica then needs its own
first cache miss for the same prompt.

Related vLLM-Omni implementation: [vllm-project/vllm-omni#2962](https://github.com/vllm-project/vllm-omni/pull/2962).

### Checklist Before Starting

- [x] Search for similar PRs. Query: [prompt embedding cache PRs](https://github.com/verl-project/verl-omni/pulls?q=is%3Apr+%22prompt+embed%22+cache)

### Test

The new CPU tests cover:

- Default tensor collation and opt-in list-preserving collation.
- Flat and packed/batched token-ID lists.
- List and tensor inputs at each adapter boundary.
- Preservation of tokenizer batch dimensions.
- Exact post-`encode_prompt()` parity with the pre-change output for both the
  cache-disabled tensor path and cache-enabled list path, including embeddings,
  masks, shape, dtype, and device.
- Cache hits for repeated list-valued Qwen-Image and Wan prompts.
- `prompt_ids=None` with precomputed prompt embeddings.
- DPO and DiffusionNFT request preparation without early tensorization.
- FlowGRPO forwarding list inputs through the cache boundary.
- Wan empty-negative-prompt handling for both list and tensor inputs.
- Configuration defaults and rejection of non-positive cache sizes.
- Cache-routing request IDs: random by default, stable per sample only when
  both prompt caching and routing affinity are enabled (covered by the separate
  agent-loop CPU tests).

Focused test command:

```bash
PYTHONPATH=/path/to/verl:/path/to/vllm-omni \
pytest -q \
  tests/pipelines/test_prompt_embed_cache_on_cpu.py \
  tests/workers/config/test_diffusion_config_on_cpu.py
```

Validation results:

- `git diff --check origin/main...HEAD`: passed.
- The focused suite completed with `46 passed` against the compatible local
  vLLM-Omni checkout used while developing the adapter changes.
- A later rerun against a different local vLLM-Omni revision stopped during test
  collection because that checkout no longer exports
  `vllm_omni.diffusion.worker.utils.DiffusionRequestState`. No test assertion was
  reached in that rerun; CI should use the repository's compatible pinned
  vLLM-Omni revision.

#### Performance evidence

The evaluation dataset used for the measurements below has short prompts, with
an average prompt length of only about 20--30 tokens. Text encoding therefore
accounts for a relatively small fraction of end-to-end generation and step time.
Based on the current comparison, this PR provides prompt-cache compatibility,
but the end-to-end gain is limited and it is not recommended as a priority
optimization. Workloads with longer prompts or more expensive text encoders may
benefit more, but that has not been measured here.

##### FlowGRPO

65b4ee316bda2ba3b22d5e42299b9876

##### DiffusionNFT

bb88eb87474cf9cfd3940bd45839db01

This DiffusionNFT plot is from the OCR dataset with each prompt repeated 10
times. In my assessment, the observed end-to-end improvement is small for this
workload. The repeated prompts do not make text encoding the dominant cost, and
replica-local cache routing limits how much of the repeated work can be reused.

These measurements do not by themselves establish the cache hit rate. Cache
entries are private to a rollout replica, so a repeated-prompt workload should
enable routing affinity when it needs repeated generations of one sample to
reuse the same replica-local entry.

### API and Usage Example

Enable the cache in the diffusion rollout configuration:

```yaml
actor_rollout_ref:
  rollout:
    enable_prompt_embed_cache: true
    prompt_embed_cache_size: 32
    enable_prompt_embed_cache_routing_affinity: true
```

Equivalent Hydra overrides:

```bash
actor_rollout_ref.rollout.enable_prompt_embed_cache=true \
actor_rollout_ref.rollout.prompt_embed_cache_size=32 \
actor_rollout_ref.rollout.enable_prompt_embed_cache_routing_affinity=true
```

`prompt_embed_cache_size` must be positive. The default configuration is:

```yaml
enable_prompt_embed_cache: false
prompt_embed_cache_size: 32
enable_prompt_embed_cache_routing_affinity: false
```

Routing affinity is meaningful only together with
`enable_prompt_embed_cache=true`. It does not change prompt values, sampling
parameters, or model weights; it only replaces the per-generation random
request ID with the repeated sample's `uid` for rollout-server routing.

#### Cache configuration

- Adds `enable_prompt_embed_cache` and `prompt_embed_cache_size` to
  `DiffusionRolloutConfig` and the generated trainer configurations.
- Validates that `prompt_embed_cache_size` is positive.
- Passes the cache enable and size options through `vLLMOmniHttpServer` to
  `AsyncOmni`.
- Adds `enable_prompt_embed_cache_routing_affinity`, disabled by default. The
  diffusion agent loop uses a sample's stable `uid` as the request ID only when
  it and prompt caching are both enabled; this is a verl-omni routing change,
  not an `AsyncOmni` cache-engine option. All other requests retain random
  UUIDs.

#### Prompt boundary

- Adds opt-in `preserve_lists=True` support to request-batch prompt-ID and mask
  collation. The default collation behavior remains tensor-based.
- Keeps pre-tokenized IDs and masks as lists through request preparation when they
  must participate in the prompt-cache key.
- Converts list inputs to device tensors only inside each adapter's or shared
  mixin's `encode_prompt()` implementation, after the cache wrapper has
  inspected the arguments.
- Retains the existing tensor path for callers that explicitly provide tensors.

#### Existing validated precedent: SD3.5 FlowGRPO

SD3.5 FlowGRPO already follows the same delayed-tensorization pattern in its
token-ID prompt-encoding boundary. Its `encode_prompt_from_token_ids()` path
accepts Python token-ID lists and converts them to device `torch.long` tensors
inside the CLIP and T5 encoding helpers, rather than before prompt encoding.

The existing `tests/pipelines/test_sd3_token_id_prompt_on_cpu.py` parity tests
compare this path with vLLM-Omni's text-based `encode_prompt()` and require exact
equality (`rtol=0`, `atol=0`). They also cover batch dimensions,
`num_images_per_prompt`, empty prompts, padding, and truncation. This provides an
already validated model-level precedent for moving list-to-tensor conversion into
the prompt-encoding boundary.

#### Adapter compatibility

- Preserves single-prompt and packed/batched prompt shapes.
- Computes batch size from either tensors or nested lists without hard-coding it.
- Preserves Qwen-Image True-CFG and Wan empty-negative-prompt behavior.
- Does not change the final prompt embedding or attention-mask contracts consumed
  by the denoising pipelines.

### DIFF
--- a/docs/contributing/integrating_prompt_embedding_cache.md
+++ b/docs/contributing/integrating_prompt_embedding_cache.md
@@ -1,0 +1,232 @@
+# Integrating the Diffusion Prompt-Embedding Cache
+
+Last updated: 08/13/2026.
+
+This guide defines the contract for making a diffusion rollout pipeline
+compatible with vLLM-Omni's prompt-embedding cache. The cache wraps
+`pipeline.encode_prompt()` and reuses its inference outputs across requests
+with the same cache key. It is disabled by default.
+
+The vLLM-Omni implementation and the authoritative cache-key rules are in
+[vllm-project/vllm-omni#2962](https://github.com/vllm-project/vllm-omni/pull/2962).
+Read that implementation before introducing a new `encode_prompt()` signature
+or forwarding a new argument through an existing one.
+
+## Primary Correctness Requirement
+
+Prompt caching must not change the semantic output of `encode_prompt()`.
+Whether the wrapper is disabled, takes a miss, or takes a hit, the pipeline must
+receive the same inference result that the pre-cache implementation produced.
+For prompt embeddings and masks this includes:
+
+- numerical values;
+- device;
+- dtype;
+- batch shape;
+- maximum-sequence-length truncation behavior;
+- `num_images_per_prompt` or `num_videos_per_prompt` expansion behavior.
+
+The cache stores detached outputs, so downstream rollout code must treat prompt
+embeddings and masks as read-only. Do not add in-place mutations after
+`encode_prompt()`. A miss executes the underlying method and stores its output;
+a hit returns that stored value. The observable inference values and shapes must
+remain identical in both cases.
+
+Explicit Tensor inputs remain supported for compatibility, but they bypass the
+cache. Precomputed prompt embeddings also retain their existing path and must
+not be converted back to token IDs just to make a cache key.
+
+## Cache-Key Contract
+
+vLLM-Omni binds the actual positional and keyword arguments to the
+`encode_prompt()` signature, applies defaults, and builds a key from every
+bound argument. A call is cacheable only when every bound argument is safe to
+serialize as a key and no precomputed embedding argument is supplied.
+
+Supported key values are recursively composed from:
+
+- `None`, `str`, `int`, `float`, `bool`, and `bytes`;
+- `torch.device` and `torch.dtype`;
+- NumPy scalar values;
+- `list` and `tuple` containing only supported values;
+- `dict` whose values contain only supported values.
+
+`torch.Tensor`, PIL images, NumPy arrays, custom objects, and a list, tuple, or
+dictionary containing any of them are not cacheable. If any bound
+`encode_prompt()` argument is not cacheable, vLLM-Omni bypasses the cache for
+the entire call rather than building a partial key. Therefore, a list is not
+the only supported input type, but token IDs and masks should normally be kept
+as lists at the `encode_prompt()` boundary instead of tensors.
+
+Calls with a non-`None` `prompt_embeds`, `negative_prompt_embeds`,
+`pooled_prompt_embeds`, `negative_pooled_prompt_embeds`, `prompt_embeds_mask`,
+or `negative_prompt_embeds_mask` also bypass the cache. They are already
+precomputed-output paths and caching them would retain large tensor values
+without avoiding text-encoder work.
+
+All arguments that can affect text encoding must be included in the
+`encode_prompt()` call and be considered when auditing key eligibility. This
+includes, where the pipeline uses them:
+
+- positive prompt IDs and masks;
+- negative prompt IDs and masks;
+- maximum sequence length;
+- output multiplicity such as `num_images_per_prompt` or
+  `num_videos_per_prompt`;
+- every other non-default argument that changes the text-encoding result.
+
+Do not drop a semantic argument merely to make a call cacheable. Instead,
+represent it with a supported value type or allow that call to bypass the cache.
+
+## Required Pipeline Changes
+
+Do not update only one visible `encode_prompt()` call. Trace every path that
+supplies IDs, masks, or text to the text encoder.
+
+1. **Request batching:** For token-ID adapters, use the list-preserving path in
+   [`request_batch.py`](../../verl_omni/pipelines/request_batch.py) before
+   cacheable calls only. When the cache is enabled,
+   `collate_prompt_rows(..., preserve_lists=True)` and
+   `collate_prompt_mask(..., preserve_lists=True)` batch and pad list values
+   without converting them to tensors before the cache wrapper. When it is
+   disabled or not installed, preserve the original tensor collation path.
+2. **Adapter boundary:** Make `forward()` and, if implemented,
+   `prepare_encode()` accept the forms the cacheable path can produce: flat
+   lists, batched lists, and existing Tensor inputs. Compute batch size without
+   forcing a list through early tensorization.
+3. **`encode_prompt()` implementation:** Accept the supported list forms and
+   convert IDs and masks to tensors on `self.device` inside `encode_prompt()`,
+   after the wrapper has built or looked up the key. The conversion must
+   canonicalize list and Tensor inputs to the same tensors used by the
+   pre-cache implementation.
+4. **Raw-text fallback:** When a step-execution path tokenizes raw text, return
+   token lists while `self._prompt_embed_cache.enabled` is true. Preserve its
+   existing Tensor tokenizer/device path when the cache is disabled.
+5. **Special cases:** Preserve precomputed embedding bypasses. For an empty
+   negative prompt, construct negative IDs and masks in the same compatible
+   representation as the positive prompt rather than introducing an early
+   Tensor only on the cacheable path.
+
+The cache configuration must be exposed through
+`DiffusionRolloutConfig` and forwarded by
+[`vllm_omni_async_server.py`](../../verl_omni/workers/rollout/vllm_rollout/vllm_omni_async_server.py)
+to `AsyncOmni`:
+
+```yaml
+actor_rollout_ref:
+  rollout:
+    enable_prompt_embed_cache: true
+    prompt_embed_cache_size: 32
+```
+
+`prompt_embed_cache_size` is per diffusion worker, must be positive, and uses
+least-recently-used eviction.
+
+## Current Reference Paths
+
+The current integration audits every token-ID rollout path that reaches
+`encode_prompt()`:
+
+| Pipeline | Required cache-compatible paths |
+| --- | --- |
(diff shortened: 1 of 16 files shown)


knlnguyen1802 (commented):
Hi could you provide a convergent curse before review ?

zhtmike:
need to update doc of how to use prompt cache correctly.

Sky-Trigger:
@zhtmike @knlnguyen1802 
This is the FlowGRPO training curve before my PR is applied. Could you take a look and see if it looks right? There must be something wrong before this pr.
09e62166ca52716faefb0ba04d1f2a6b

zhtmike:
gentle ping @knlnguyen1802

knlnguyen1802:
> @zhtmike @knlnguyen1802 This is the FlowGRPO training curve before my PR is applied. Could you take a look and see if it looks right? There must be something wrong before this pr. 09e62166ca52716faefb0ba04d1f2a6b

I think we cant get any clue on this chart

verl_omni/pipelines/request_batch.py
```diff
@@ -173,8 +221,15 @@ def collate_prompt_mask(
...
         field_name=field_name,
         pad_value=0,
+        preserve_lists=preserve_lists,
     )
     if mask is not None:
+        if isinstance(mask, list):
+            mask = [[value != 0 for value in row] for row in mask]
+            if target_seq_len is not None:
+                mask = [row[:target_seq_len] + [False] * max(target_seq_len - len(row), 0) for row in mask]
+            return mask
```
knlnguyen1802, line 232: Need to check carefully whether this padding is correct

verl_omni/pipelines/qwen_image_flow_grpo/vllm_omni_rollout_adapter.py
```diff
@@ -192,8 +207,7 @@ def _tokenize_text_prompt(self, text: str | list[str]):
             max_length=self.tokenizer_max_length + self.prompt_template_encode_start_idx,
             padding=True,
             truncation=True,
-            return_tensors="pt",
-        ).to(self.device)
```
knlnguyen1802, line 196: so this is redundant code until now ?

Sky-Trigger:
the rewards chart does not 

> > @zhtmike @knlnguyen1802 This is the FlowGRPO training curve before my PR is applied. Could you take a look and see if it looks right? There must be something wrong before this pr. 09e62166ca52716faefb0ba04d1f2a6b
> 
> I think we cant get any clue on this chart

The rewards curve does not show the same normal upward trend as it did in PR #287. The curve shown here was obtained from a run on the current main branch without cherry-picking this PR.

knlnguyen1802:
> The rewards curve does not show the same normal upward trend as it did in PR #287. The curve shown here was obtained from a run on the current main branch without cherry-picking this PR.

There is some problem with stepwise batching now, on analyze it

Sky-Trigger:
> > The rewards curve does not show the same normal upward trend as it did in PR #287. The curve shown here was obtained from a run on the current main branch without cherry-picking this PR.
> 
> There is some problem with stepwise batching now, on analyze it

OK, I’ll show the speedup and the reward curve on DiffusionNFT using the OCR dataset with the prompt length manually increased from around 20 tokens to 1,000 tokens.

Sky-Trigger:
@knlnguyen1802 Keeping both paths for now. The non-prompt-cache path is exactly the same as before. The prompt-cache path has been verified, and `encode_prompt` returns exactly the same as before. Updated the description above, feel free to take a look.

Sky-Trigger:
ready for merge now

docs/contributing/integrating_prompt_embedding_cache.md
zhtmike, line 1: I think the doc is deviate from your implementation.

You change all pipelines for this, but the doc never mentions this

zhtmike:
you should state clearly what need to be modified for support prompt caching in model contribution guide

zhtmike:
and kindly ping @ZihaoW123 @AndyZhou952 @wtomin  for review for since they are the pipeline owners

docs/contributing/integrating_prompt_embedding_cache.md
```diff
@@ -0,0 +1,233 @@
...
+- NumPy scalar values;
+- `list` and `tuple` containing only supported values;
+- `dict` whose values contain only supported values.
+
+`torch.Tensor`, PIL images, NumPy arrays, custom objects, and a list, tuple, or
+dictionary containing any of them are not cacheable. If any bound
+`encode_prompt()` argument is not cacheable, vLLM-Omni bypasses the cache for
+the entire call rather than building a partial key. Therefore, a list is not
+the only supported input type, but token IDs and masks should normally be kept
+as lists at the `encode_prompt()` boundary instead of tensors.
```
zhtmike, line 59: clear instructions. looks good

zhtmike: I think we should optimize the prompt cache hash in vllm-omni side, it is just too difficult to satisfy this criteria.
cc @SamitHuang @knlnguyen1802

Sky-Trigger: Yes, and this PR is mainly about making prompt caching compatible. But I don't actually recommend using it, since the performance gain is almost negligible compared to the potential risks.

zhtmike (approved):
looks good for me, waiting other's feedback

ZihaoW123:
After making this change, have you tested whether WAN22 is working properly?

Sky-Trigger:
> After making this change, have you tested whether WAN22 is working properly?

No, I haven’t run an end-to-end Wan2.2 test yet. However, I inspected the entire pipeline, and the adapter follows the same path as the other verified Qwen-Image RL algorithms. If this is a concern, I can remove the Wan2.2 adapter change from this patch and leave it for future contributors to address.

knlnguyen1802 (changes requested):
For this PR, let's me verify it for QwenImage and SD3.5 on GPU first, recently we should take care of precision issue especially for QwenImage

knlnguyen1802:
Update result running on GPU
@Sky-Trigger  Let's wait for verl pin update and move this PR forward
Also please verify on Wan2.2 or remove the adapter change
qwen_image_prompt_embedding_reward

Sky-Trigger:
@knlnguyen1802 PTAL

knlnguyen1802 (approved):
LGTM now, make sure CI pass before merge

Sky-Trigger:
@knlnguyen1802  all ci test passed

PR #358, opened 2026-08-06
[trainer, rollout] feat: refill incomplete diffusion rollout groups

WenzheWang:
### What does this PR do?

Diffusion V1 currently treats a prompt tagged `failure` as sampleable. If one
session in a GRPO group fails after sibling sessions have produced trajectories,
the residual partial group can reach training.

This PR adds an opt-in synchronous-training policy that:

- evicts a selected failed prompt and every residual trajectory in its group;
- retains complete groups selected in the same upstream sample;
- submits the exact number of replacement prompts;
- bounds consecutive replacement rounds;
- preserves existing behavior by default and leaves validation sampling unchanged.

Enable it with:

```yaml
trainer:
  v1:
    sampler:
      drop_incomplete_groups: true
      max_incomplete_group_refill_rounds: 3
```

The effective generation batch size is one while this policy is enabled so an
arbitrary selected failed-group count can be replaced exactly.

### Main vs. this branch: controlled failure-group comparison

Base: `verl-project/verl-omni@cc65f27a3c27edd98f8668f5b2d29c5a1ff6ce7b`

Fixture: `batch_size=2`; group `finished` has two trajectories; group `partial`
is tagged `failure` after retaining one trajectory. The same fake TransferQueue
fixture is exercised against the upstream `ReplayBuffer` directly and the
opt-in trainer sampling path.

| Behavior | main | this branch |
|---|---:|---:|
| Sampled groups | `finished`, `partial` | `finished`, `replacement` |
| Sampled trajectories | 3 | 4 |
| Refill calls | none | `[1]` |
| Failed group reaches training | yes | no |
| Failure metrics | none | 1 group / 1 trajectory evicted; 1 prompt refilled |

### Design and code changes

- Always construct the exact upstream verl `ReplayBuffer`, including when the
  opt-in policy is enabled.
- Keep upstream `ReplayBuffer.sample()` responsible for polling, terminal-group
  selection, ordering, and off-policy handling.
- After an upstream sample, use that same replay-buffer snapshot to identify
  selected failed groups, retain complete groups, clear only failed residuals,
  and request exact replacements.
- Restrict the opt-in policy to `sync`; default `sync`, default
  `separate_async`, and validation retain upstream sampling behavior.
- Sum rollout-failure metrics across `parameter_sync_step` substeps.
- Document the opt-in configuration and regenerate both diffusion reference
  configurations.

### Test

Current commit tested: `e7476ca56b9b88702578b573672d7062cd191f4b`

Local checks on the current tree:

```text
ruff check: passed
ruff format --check: passed
docs timestamp/docstrings/license/device API/DataProto/test structure: passed
compile-all and git diff --check: passed
```

The current tree was also validated with the repository's pinned verl commit
`8a694930275061f52ebd538c906ef8819af56dbd` in the existing cluster runtime:

```text
tests/agent_loop/test_diffusion_agent_loop_correctness_on_cpu.py
tests/trainer/diffusion/test_v1_replay_buffer_on_cpu.py
tests/special_sanity/test_config_docs.py

23 passed, 23 dependency deprecation warnings in 19.52s
Slurm job 265469: COMPLETED, exit 0, 27 seconds wall time
source SHA256: e1f2dba7096b0bed30ccd176c9afcf32bc730fc4007fcf65d1efa11ffc85055a
```

Coverage includes the controlled main/fix comparison, partial and zero-output
failures, multiple and repeated failures, bounded retries, exact refill counts,
unselected-group preservation, validation behavior, default `sync` and
`separate_async` behavior, and exact upstream-buffer construction with the
policy both disabled and enabled.

This is hardware-independent control-plane behavior; it does not change model
kernels, tensor precision, or model outputs, so no precision-alignment claim is
made.

### Checklist Before Starting

- [x] Searched for similar open PRs; no same-scope implementation was found.
- [x] PR title follows `[{modules}] {type}: {description}`.

- [x] Refreshed the branch against the current PR base.
- [x] Updated documentation and generated configuration.
- [x] Added CPU regression coverage for the owned trainer behavior.


verl_omni/agent_loop/diffusion_agent_loop_tq.py
```diff
@@ -158,12 +158,32 @@ async def _run_prompt(
...
+                    terminal_tag["failure_reason"] = "runtime_error"
+                else:
+                    terminal_tag["failure_reason"] = "error"
+            await tq.async_kv_put(key=uid, partition_id=partition_id, tag=terminal_tag)
         except Exception as e:
             logger.exception(f"Error in _run_prompt for uid={uid}: {e}")
             if tasks:
                 await asyncio.gather(*tasks, return_exceptions=True)
-            await tq.async_kv_put(key=uid, partition_id=partition_id, tag={"status": "failure"})
+            if isinstance(e, TimeoutError):
```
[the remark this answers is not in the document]
WenzheWang: Update for the current head (`e7476ca`): this PR no longer changes `diffusion_agent_loop_tq.py`, cancellation reporting, or `failure_reason` semantics; that file is identical to base `cc65f27`, so the branch-specific classification and test details from the earlier reply no longer apply. This thread is historical and outdated with respect to the current diff.

For completeness, the original Python premise is still incorrect for the supported versions: `asyncio.CancelledError` inherits `BaseException`, not `Exception`, so `except Exception` does not catch external task cancellation. No change is required in the current PR.

WenzheWang:
Current status for head `e7476ca` on base `cc65f27`:

- the custom replay-buffer class and module have been removed in response to review feedback;
- the trainer now always constructs the exact upstream verl `ReplayBuffer`;
- the opt-in synchronous path calls upstream `ReplayBuffer.sample()` and only wraps the selected failed groups with exact eviction/refill handling;
- default `sync`, default `separate_async`, and validation keep upstream sampling behavior.

Validation on the exact current tree:

- Ruff check and format check: passed;
- repository docs/license/device API/DataProto/test-structure checks and compile-all: passed;
- focused agent-loop, trainer replay-buffer, and config-doc tests: **23 passed** against the pinned verl revision;
- Slurm job `265469`: `COMPLETED`, exit `0`.

The PR is mergeable, has no unresolved review threads, and is ready for re-review. @knlnguyen1802 @zhtmike @SamitHuang, could you please take another look at the current head when convenient?

verl_omni/trainer/diffusion/v1/replay_buffer.py
```diff
@@ -0,0 +1,165 @@
+# Copyright 2026 Bytedance Ltd. and/or its affiliates
+#
```
knlnguyen1802: When design diffusion V1, I want to inherit `ReplayBuffer` without create a new one because code is very redundant. With this, we might need to update frequently with verl.

knlnguyen1802 (commented):
Will this PR affect separate async of diffusion v1 ?

WenzheWang:
@knlnguyen1802 No. On the current head (`e7476ca`), `separate_async` keeps the upstream replay-buffer behavior:

- with the default `drop_incomplete_groups=false`, both `sync` and `separate_async` construct the exact upstream verl `ReplayBuffer`;
- enabling incomplete-group refill in `sync` mode still constructs that exact upstream buffer and applies refill only in the trainer after an upstream sample;
- enabling the policy with a non-`sync` trainer mode raises a clear `ValueError` during initialization.

There is no custom replay-buffer class or module on the current branch, and the `PolicyGradientDiffusionTrainerV1SeparateAsync` lifecycle is unchanged. Current regression coverage verifies the factory selection and mode boundary; the exact current tree passes 23 focused tests against the pinned verl revision (Slurm `265469`, exit `0`).

verl_omni/trainer/diffusion/v1/replay_buffer.py
WenzheWang: Addressed on the current head (`e7476ca`). The trainer now always constructs the exact upstream verl `ReplayBuffer`, including when incomplete-group refill is enabled; there is no custom replay-buffer class or module. The opt-in trainer path calls `ReplayBuffer.sample()` directly, then uses that same sampled snapshot only to retain selected complete groups, evict selected failed residuals, and request exact replacements. Polling, terminal selection, ordering, and off-policy handling therefore remain upstream-owned. Default `sync`, default `separate_async`, and validation continue to use the upstream path unchanged. The current tree passes Ruff/static checks and 23 focused tests against the pinned verl revision (Slurm `265469`, exit `0`).

knlnguyen1802 (approved):
LGTM now

knlnguyen1802:
Let's wait for v2 release first

zhtmike:
@knlnguyen1802 is this past CI?

knlnguyen1802:
> @knlnguyen1802 is this past CI?

yes I remember trigger it before

zhtmike:
see

> > @knlnguyen1802 is this past CI?
> 
> yes I remember trigger it before

need fix ci for this

knlnguyen1802:
> see
> 
> 
> 
> > > @knlnguyen1802 is this past CI?
> 
> > 
> 
> > yes I remember trigger it before
> 
> 
> 
> need fix ci for this

Ok let's me fix it
Thanks for notify

PR #116, opened 2026-05-28
[reward] feat: add HTTP reward service adapter for external scorer services

ruihanglix:
### What does this PR do?

Adds support for using external HTTP reward services (e.g., PaddleOCR-based scorer) as reward functions in RL training. This enables developers to deploy custom reward models as standalone HTTP services and connect them to verl-omni without modifying any framework code.

Key changes:
1. **`MultiVisualRewardManager` now passes extra config fields to reward functions** — any field beyond `path`/`name`/`weight` in `reward.reward_functions.<key>.*` is forwarded to the corresponding `compute_score` function via kwargs filtering. This is a generic mechanism that enables per-reward custom configuration (e.g., `server_url`, `timeout`).
2. **New generic HTTP scorer client** (`http_scorer_client.py`) — an async `compute_score` adapter that sends images to any HTTP scorer service using the pickle-based protocol.
3. **New data preprocessing script** (`qwenimage_ocr_http_service.py`) — prepares OCR data in a format compatible with the HTTP scorer service (full prompt as ground_truth instead of bare text).
4. **Example training script** (`run_qwen_image_ocr_reward_server.sh`) — demonstrates training with an external OCR reward server, no internal reward model required.

### Motivation

This addresses the RFC discussion in #103 about supporting custom reward models. The implementation follows **Option A (Custom Reward Function via Adapter)** with several modifications purposed by me:
- Users deploy their own HTTP reward server externally
- A `compute_score` adapter calls the endpoint
- The adapter plugs into the existing `MultiVisualRewardManager` path with no framework changes beyond the extra_args forwarding

This approach:
- Requires **zero GPU resources** for reward computation in the training cluster (reward runs on a separate service)
- Supports **any reward model backend** (PaddleOCR, HPSv3, aesthetic models, etc.) as long as it exposes the pickle-based HTTP interface
- Is **fully concurrent** — since `compute_score` is async and `RewardLoopWorker.compute_score_batch` uses `asyncio.gather`, all samples in a batch hit the server concurrently

### Checklist Before Starting

- [x] Search for similar PRs: https://github.com/verl-project/verl-omni/pulls?q=is%3Apr+reward+service
- [x] Format the PR title as `[{modules}] {type}: {description}`

### Test

Validated on 4x NVIDIA B300 GPUs with the Qwen-Image OCR FlowGRPO recipe using an external PaddleOCR reward server (separate GPU).

[...]

PR #366, opened 2026-08-08
[reward, recipe] fix: stop mapping HTTP scorer errors to 0 rewards

ShuoleiWang:
### What does this PR do?

The generic HTTP scorer client previously returned `{"score": 0.0}` in three failure cases:

- the server returned a non-200 HTTP status;
- the response contained an `error`;
- the response contained an empty `scores` list.

A real reward can also be `0.0`. This means the reward manager could not tell the difference between a valid zero reward and a failed scorer request.

The root cause was that the HTTP client returned a normal score instead of raising an exception. Because no exception was raised, the reward manager could not apply its existing `required` policy.

This PR changes the behavior to:

```text
Before:
HTTP/service failure -> score 0 -> training continues

After:
HTTP/service failure -> exception -> reward manager applies required policy
```

The OCR HTTP scorer recipe now sets `required=true`, so these failures stop the training run instead of contributing a zero reward.

A real response such as `{"scores": [0.0]}` remains valid.

The existing behavior for optional scorers is unchanged: when `required=false`, the reward manager may still replace a failed scorer result with zero and continue.

Related work:

- [#116](https://github.com/verl-project/verl-omni/pull/116) introduced the generic HTTP scorer client.
- [#289](https://github.com/verl-project/verl-omni/pull/289) introduced the existing `required` policy and a separate latent HTTP scorer. This PR applies the failure behavior to the older generic HTTP scorer client.

### Checklist Before Starting

- [x] Searched [open PRs for HTTP scorer error handling](https://github.com/verl-project/verl-omni/pulls?q=is%3Apr+is%3Aopen+http+scorer). No matching open PR was found.
- [x] The PR title follows `[{modules}] {type}: {description}`.

### Test

CPU tests:

```bash
pytest -q \
  tests/utils/reward_score/test_http_scorer_client_on_cpu.py \
  tests/utils/reward_score/test_latent_http_scorer_client_on_cpu.py
```

Result:

```text
12 passed
```

Static checks:

```bash
ruff check \
  verl_omni/utils/reward_score/http_scorer_client.py \
  tests/utils/reward_score/test_http_scorer_client_on_cpu.py

ruff format --check \
  verl_omni/utils/reward_score/http_scorer_client.py \
  tests/utils/reward_score/test_http_scorer_client_on_cpu.py

bash -n examples/flowgrpo_trainer/qwen_image/run_qwen_image_ocr_reward_server.sh
git diff --check
```

All checks passed. No GPU is required for this change.

### API and Usage Example

There is no function-signature or configuration-schema change.

Direct callers of `compute_score()` now receive a `RuntimeError` for non-200 responses, service-reported errors, and empty score responses.

The OCR recipe now enables the existing required-reward behavior:

```bash
'+reward.reward_functions.ocr.required=true'
```

### Design & Code Changes

- Update the generic HTTP scorer client to raise exceptions instead of returning zero for the three failure cases.
- Enable `required=true` in the OCR HTTP scorer recipe.
- Document how `required` handles scorer failures.
- Add CPU tests covering:
  - a valid zero reward;
  - a non-200 HTTP response;
  - a service-reported error;
  - an empty score response.

Retries, timeout policy, and non-finite score validation are intentionally outside the scope of this PR.

- [x] Applied the required pre-commit checks.
- [x] Updated the documentation.
- [x] Added CPU unit tests covering the changed behavior.

### AI Assistance

AI assistance was used for repository inspection, implementation drafting, tests, and preparation of this PR description. I reviewed every changed line and ran the checks listed above.

### DIFF
--- a/docs/start/http_scorer.md
+++ b/docs/start/http_scorer.md
@@ -76,6 +76,8 @@ python3 -m verl_omni.trainer.main_diffusion \
     '+reward.reward_functions.my_reward.weight=1.0' \
     '+reward.reward_functions.my_reward.required=true' \
     "+reward.reward_functions.my_reward.server_url=http://<scorer-host>:<port>" \
+    '+reward.reward_functions.my_reward.max_retries=2' \
+    '+reward.reward_functions.my_reward.retry_backoff=0.5' \
     ...
 ```
 
@@ -86,6 +88,10 @@ Key points:
 - **`weight`**: Reward weight when combining multiple reward functions.
 - **`required`**: Controls what happens when this scorer fails. Set it to `true` to stop training. If it is `false` (the default), the reward manager sets the score to `0` and continues.
 - **`server_url`**: Full URL of your scorer service (no trailing slash).
+- **`max_retries`**: Number of retries after the initial attempt (default: `2`, for at most 3 attempts total).
+- **`retry_backoff`**: Initial retry delay in seconds (default: `0.5`). The delay doubles after each failed attempt.
+
+The client retries connection and response-transfer failures, timeouts, and HTTP 408/429/5xx responses. It fails immediately without retrying for other 4xx responses, HTTP 200 responses containing an `error` field, and malformed success responses. Once the retry budget is exhausted, the final error is passed to the reward manager, where `required` determines whether training stops or continues with a zero score.
 
 Any extra key-value pairs added under the same reward function config are forwarded as `**kwargs` to `compute_score`.
 
@@ -115,5 +121,5 @@ OCR_REWARD_SERVER_URL=http://<server-ip>:19082 \
 
 - The HTTP client reuses a single `aiohttp.ClientSession` across calls to avoid per-request connection overhead.
 - Image serialization (tensor to PIL to JPEG) is offloaded to a thread pool via `asyncio.loop.run_in_executor` so it does not block the reward manager's async event loop.
-- The default request timeout is 120 seconds. If your scorer model is slow, consider scaling the service with Gunicorn workers or increasing the timeout in the client code.
+- Each attempt has a 120-second timeout. With the default retry settings, a persistent retryable failure can make three attempts before it is reported.
 - Reward-server profiling (`reward.reward_model.rollout.profiler.*`, see the [profiler guide](../perf/profiler.md)) only covers model-backed rewards served by VeRL-Omni. With an HTTP scorer, reward inference runs inside your external service — profile it there.
--- a/docs/start/http_scorer.md
+++ b/docs/start/http_scorer.md
@@ -1,7 +1,7 @@
 (http_scorer)=
 # Using an External HTTP Scorer Service
 
-Last updated: 07/17/2026
+Last updated: 08/09/2026
 
 VeRL-Omni ships a generic HTTP reward client (`verl_omni.utils.reward_score.http_scorer_client`) that sends generated images to an external scorer service over HTTP and returns the score. This is useful when your reward model is too large to co-locate with training, needs a different runtime (e.g., a separate GPU pool), or is shared across multiple experiments.
 
@@ -74,6 +74,7 @@ python3 -m verl_omni.trainer.main_diffusion \
     "+reward.reward_functions.my_reward.path=pkg://verl_omni.utils.reward_score.http_scorer_client" \
     '+reward.reward_functions.my_reward.name=compute_score' \
     '+reward.reward_functions.my_reward.weight=1.0' \
+    '+reward.reward_functions.my_reward.required=true' \
     "+reward.reward_functions.my_reward.server_url=http://<scorer-host>:<port>" \
     ...
 ```
@@ -83,6 +84,7 @@ Key points:
 - **`path`**: Module path using the `pkg://` prefix.
 - **`name`**: The async function to call (`compute_score`).
 - **`weight`**: Reward weight when combining multiple reward functions.
+- **`required`**: Controls what happens when this scorer fails. Set it to `true` to stop training. If it is `false` (the default), the reward manager sets the score to `0` and continues.
 - **`server_url`**: Full URL of your scorer service (no trailing slash).
 
 Any extra key-value pairs added under the same reward function config are forwarded as `**kwargs` to `compute_score`.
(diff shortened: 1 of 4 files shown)


zhtmike:
I think the training loop should be robust enough to handling the occasionly error from reward server (e.g., occasionaly internet issue, reward inference server is fully busy). So training should be continue with this kind of error.

So is this a really good fix?

ShuoleiWang:
> I think the training loop should be robust enough to handling the occasionly error from reward server (e.g., occasionaly internet issue, reward inference server is fully busy). So training should be continue with this kind of error.
> 
> So is this a really good fix?

@zhtmike You're right, the reward-server failure should not stop training immediately. I ignored the generic HTTP scorer currently has no retry mechanism.

This PR is trying to address: **converting a scorer failure directly into `0` makes it indistinguishable from a valid zero reward, which can distort the training signal.**

Do you think we should do something for this issue?

zhtmike:
> > I think the training loop should be robust enough to handling the occasionly error from reward server (e.g., occasionaly internet issue, reward inference server is fully busy). So training should be continue with this kind of error.
> > So is this a really good fix?
> 
> @zhtmike You're right, the reward-server failure should not stop training immediately. I ignored the generic HTTP scorer currently has no retry mechanism.
> 
> This PR is trying to address: **converting a scorer failure directly into `0` makes it indistinguishable from a valid zero reward, which can distort the training signal.**
> 
> Do you think we should do something for this issue?

as you said, can we implement a retry loop for remote http server? it should be helpful -> Tolerate the occasional failure, but can spot the systematic failure.

And can we run a e2e after implementing this, update some figures to prove that it does work?

ShuoleiWang:
> as you said, can we implement a retry loop for remote http server? it should be helpful -> Tolerate the occasional failure, but can spot the systematic failure.
> 
> And can we run a e2e after implementing this, update some figures to prove that it does work?

OK, very insightful!

ShuoleiWang:
> as you said, can we implement a retry loop for remote http server? it should be helpful -> Tolerate the occasional failure, but can spot the systematic failure.
> 
> And can we run a e2e after implementing this, update some figures to prove that it does work?

@zhtmike Hi, according to your feedback, I have added bounded retries to the generic HTTP scorer client for **transport failures, timeouts, and HTTP 408/429/5xx responses**. The default is 2 retries after the initial request, with 0.5s/1.0s backoff. The client still fails immediately on non-retryable responses.

I reused the **existing Qwen-Image FlowGRPO E2E setup on 4 H20 GPUs**. Its train batch size of 8 and `rollout.n=2` produced 16 scoring inputs for one step. I used the complete Qwen/Qwen-Image checkpoint and real OCR prompts from the Flow-GRPO training dataset. Each scoring input contained one generated JPEG and its OCR prompt, and full body/JPEG hashes confirmed that the request stayed unchanged across retries.

For the transient-failure test, I placed a local fault-injection proxy in front of PaddleOCR. The proxy returned HTTP 503 for the first two attempts of each scoring input, while the HTTP client waited approximately 0.5s/1.0s and retried. On the 3rd attempt, the proxy forwarded the unchanged request body to PaddleOCR. All 16 upstream calls returned finite scores, and training completed global step 1. For the persistent-failure test, the proxy always returned 503 and never called PaddleOCR. No request exceeded 3 attempts, and training stopped before a global step with `Required sub-reward ... failed after 3 attempts`. I also added a CPU integration test with a real local HTTP server, covering both retry recovery and retry exhaustion (1 passed).

zhtmike:
Could you provide a reward curve in jpeg format?

examples/flowgrpo_trainer/qwen_image/run_qwen_image_ocr_reward_server.sh
```diff
@@ -83,6 +83,7 @@ python3 -m verl_omni.trainer.main_diffusion \
     "+reward.reward_functions.ocr.path=pkg://verl_omni.utils.reward_score.http_scorer_client" \
     '+reward.reward_functions.ocr.name=compute_score' \
     '+reward.reward_functions.ocr.weight=1.0' \
+    '+reward.reward_functions.ocr.required=true' \
```
knlnguyen1802: Already have too many script under /flowgpo/qwen_image/ consider mention this in docs only

ShuoleiWang:
> Could you provide a reward curve in jpeg format?
pr366_qwen_image_reward_curve


I re-ran the experiment and here is the reward curve in JPEG format. The upper panel shows the raw mean OCR reward per step; the lower panel shows the number of injected 503 responses per step.
I reused the repository’s existing Qwen-Image FlowGRPO E2E setup and ran 2 independent 20-step traces on each 4 H20 GPUs. In the **sparse-fault run**, a local proxy injected 1 or 2 HTTP 503 responses into a small subset of scoring requests. All requests completed successfully within the configured three-attempt limit, and training reached step 20.
The 2 reward curves are not expected to match exactly the same, because the diffusion rollouts are stochastic and generated different images.

examples/flowgrpo_trainer/qwen_image/run_qwen_image_ocr_reward_server.sh
ShuoleiWang: Done

zhtmike:
> > Could you provide a reward curve in jpeg format?
> 
> pr366_qwen_image_reward_curve
> I re-ran the experiment and here is the reward curve in JPEG format. The upper panel shows the raw mean OCR reward per step; the lower panel shows the number of injected 503 responses per step. I reused the repository’s existing Qwen-Image FlowGRPO E2E setup and ran 2 independent 20-step traces on each 4 H20 GPUs. In the **sparse-fault run**, a local proxy injected 1 or 2 HTTP 503 responses into a small subset of scoring requests. All requests completed successfully within the configured three-attempt limit, and training reached step 20. The 2 reward curves are not expected to match exactly the same, because the diffusion rollouts are stochastic and generated different images.

Thanks!

docs/start/http_scorer.md
```diff
@@ -74,7 +74,10 @@ python3 -m verl_omni.trainer.main_diffusion \
     "+reward.reward_functions.my_reward.path=pkg://verl_omni.utils.reward_score.http_scorer_client" \
     '+reward.reward_functions.my_reward.name=compute_score' \
     '+reward.reward_functions.my_reward.weight=1.0' \
+    '+reward.reward_functions.my_reward.required=true' \
```
zhtmike, line 77: `required` <--- this name is too ambiguous, pls use a better name

ShuoleiWang: The `required` tag is not introduced by this PR, it is an existing MultiVisualRewardManager option introduced in #289 and already used by the SD3.5 DRM recipe. And it means a sub-reward error is propagated instead of being converted to a 0 score. I'm worried renaming it may damage other features or configurations.

zhtmike: I see, let us keep it

ShuoleiWang: It seems the CPU failure in CI was caused by the Qwen3-Omni test incompatibility, and it has already been fixed on main. I have synced the latest main. Could you please retrigger the ci label?

zhtmike (approved):

zhtmike:
thanks for the pr！

PR #341, opened 2026-08-04
[rollout, diffusion,  doc, reward] feat: Ltx2.3 t2av flowgrpo + clap/imagebind reward

ZihaoW123:
### What does this PR do?
Ltx2.3 t2av flowgrpo + clap/imagebind reward

image

> Add **concise** overview of what this PR aims to achieve or accomplish. Reference related GitHub issues and PRs that help with the review.

### Checklist Before Starting

- [x] Search for similar PRs. Paste at least one query link here: ...
- [x] Format the PR title as `[{modules}] {type}: {description}` (This will be checked by the CI)
  - `{modules}` include `fsdp`, `vllm_omni`, `rollout`, `trainer`, `ci`, `training_utils`, `recipe`, `ray`, `worker`, `single_controller`, `misc`, `perf`, `model`, `algo`, `env`, `tool`, `ckpt`, `doc`, `data`, `cfg`, `reward`, `diffusion`, `omni`, `tests`, `docker`
  - If this PR involves multiple modules, separate them with `,` like `[diffusion, doc]`
  - `{type}` is in `feat`, `fix`, `refactor`, `chore`, `test`
  - If this PR breaks any API (CLI arguments, config, function signature, etc.), add `[BREAKING]` to the beginning of the title.
  - Example: `[BREAKING][diffusion, fsdp] feat: new rollout scheduler`

### Test

> For changes that can not be tested by CI (e.g., algorithm implementation, new model support), validate by experiment(s) and show results like training curve plots, evaluation results, etc.

### API and Usage Example

> Demonstrate how the API changes if any, and provide usage example(s) if possible.

```python
# Add code snippet or script demonstrating how to use this
```

### Design & Code Changes

> Demonstrate the high-level design if this PR is complex, and list the specific changes.

### Checklist Before Submitting

> [!IMPORTANT]
> Please check all the following items before requesting a review, otherwise the reviewer might deprioritize this PR for review.

- [x] Read the [Contribute Guide](https://github.com/verl-project/verl-omni/blob/main/CONTRIBUTING.md).
- [x] Apply [pre-commit checks](https://github.com/verl-project/verl-omni/blob/main/CONTRIBUTING.md#code-linting-and-formatting): `pre-commit install && pre-commit run --all-files --show-diff-on-failure --color=always`
- [x] Add / Update the documentation.
- [x] Add unit or end-to-end test(s) to [the CI workflow](https://github.com/verl-project/verl-omni/tree/main/.github/workflows) to cover all the code. If not feasible, explain why: ...

PR #373, opened 2026-08-10
[BREAKING][rollout, diffusion] feat: make pixel responses always uint8

jiefangxuanyan:
### What does this PR do?

This PR changes pixel-valued diffusion rollout `responses` from normalized floating point to an unconditional uint8 representation in `[0, 255]`. This breaking interface change reduces the response tensor payload to one quarter of its original `float32` size. Latent-valued responses remain floating point.

<details>
<summary><strong>Superseded option-3 overview</strong></summary>

This PR adds an optional `uint8` transport representation for pixel-valued diffusion rollout `responses`. The option reduces the response tensor payload to one quarter of its original `float32` size while preserving the historical `responses` and custom-reward contract by default.

</details>

#### Background and motivation

I previously profiled the `update_actor` stage with `examples/flowgrpo_trainer/qwen_image/run_qwen_image_ocr.sh` in an FSDP2 setup. The detailed profiling results are recorded in the [July 28, 2026 weekly report](https://docs.google.com/document/d/1UD8IbrFWIjJGCs1MF2yjPAV5xnXKdmIDsj7UchqhHrc/edit?tab=t.0). A single `dispatch_lazy_compute_data_proto` call took approximately **5.1 seconds**. In a profiling experiment that removed fields unused by `update_actor`, the same call dropped to **1.188 seconds**.

The dominant field was `responses`, which contains all full generated images from rollout. Neither `update_actor` nor `old_log_prob` actually consumes these images, but the current `DataProto`-as-dictionary data bus still transports the field to those workers.

This field-removal result demonstrates the size of the avoidable dispatch overhead, but it is not the expected result of this PR. Eliminating such unused fields cleanly is a broader architectural problem. A robust solution would likely require every worker to declare the fields it consumes and the dispatcher to project each payload accordingly, which is considerably more complex than the current generic dictionary bus. This PR intentionally does **not** attempt to solve that separate problem.

As a simpler mitigation proposed during the profiling discussion, pixel responses can be quantized from normalized `float32` values to `uint8` immediately after rollout. This does not eliminate unnecessary transfers: one quarter of the original response bytes still need to be carried. Therefore, the improvement should be smaller than the measured **5.1 s to 1.188 s** field-removal result. This PR also does not claim a 4x reduction in total IPC time because dispatch has other fields and overheads; the measured end-to-end result is reported below.

The trade-off is that early quantization has a broader impact than filtering `responses` only for `update_actor`. Once the rollout result is represented as `uint8`, every later IPC boundary that carries pixel responses can transfer the compact representation. This includes returning rollout results and passing images into the reward stage, in addition to the `old_log_prob` and `update_actor` paths that motivated the investigation.

> [!WARNING]
> This branch was originally created before [PR #341](https://github.com/verl-project/verl-omni/pull/341) was merged on August 6, 2026. That PR introduced the LTX-2.3 T2AV FlowGRPO recipe together with the ImageBind reward. The recipe explicitly requests `output_type=pt`; under the historical rollout behavior, its decoded `[0, 1]` floating-point video tensor was passed through `responses` to ImageBind without uint8 quantization. ImageBind then performed floating-point resizing and normalization before model inference, so it could consume precision beyond an 8-bit pixel representation. With this PR, that built-in feature instead receives quantized uint8 pixel responses. This is a behavior and reward-numerics change, not only a custom-reward API change. @ZihaoW123, as the author of PR #341, we would appreciate your input on this behavior change.

> [!NOTE]
> In the August 18 weekly meeting, @ZihaoW123 confirmed that the quantization loss is acceptable, provided the video is converted back to normalized floating point for ImageBind. This PR performs the required `float() / 255.0` conversion inside ImageBind preprocessing while keeping the global `responses` and custom-reward interface uint8.

### Checklist Before Starting

- [x] Search for similar PRs. No matching PR was found: [open and closed PRs matching `responses uint8`](https://github.com/verl-project/verl-omni/pulls?q=is%3Apr+responses+uint8)

#### Performance benchmark

I benchmarked `examples/flowgrpo_trainer/qwen_image/run_qwen_image_ocr_fsdp2_benchmark.sh` on the cluster, comparing the historical float32-response baseline with this PR's uint8-response implementation. The baseline ran for 58 steps and the uint8 experiment for 52 steps; the first step was excluded from the statistics for both groups. `Benefit` is the baseline average minus the uint8 average, in seconds per step.

| Stage | float32 avg/step (s) | float32 std/step (s) | uint8 avg/step (s) | uint8 std/step (s) | Benefit (s/step) |
| --- | ---: | ---: | ---: | ---: | ---: |
| `step` | 121.30 | 0.83 | 107.61 | 1.23 | **13.69** |
| `gen` | 33.06 | 0.39 | 29.97 | 0.99 | 3.09 |
| `reward` | 24.79 | 0.50 | 21.12 | 0.55 | 3.67 |
| `old_log_prob` | 15.37 | 0.33 | 12.83 | 0.42 | 2.54 |
| `adv` | 0.01 | 0.00 | 0.01 | 0.00 | 0.00 |
| `update_actor` | 36.37 | 0.26 | 34.01 | 0.29 | 2.36 |
| `update_weights` | 7.36 | 0.32 | 6.93 | 0.14 | 0.43 |
| `dump_rollout_generations` | 4.34 | 0.05 | 2.74 | 0.04 | 1.60 |
| `start_profile` | 0.00 | 0.00 | 0.00 | 0.00 | 0.00 |
| `stop_profile` | 0.00 | 0.00 | 0.00 | 0.00 | 0.00 |
| `other` | 0.00 | 0.00 | 0.00 | 0.00 | 0.00 |

The end-to-end average step time decreased by **13.69 seconds**, from 121.30 seconds to 107.61 seconds per step, an approximately **11.3%** reduction. The per-stage improvements sum to the same 13.69-second reduction.

#### Rollout generation inspection

I compared float32 and uint8 `dump_rollout_generations` outputs at steps 1 and 50. Although different random seeds prevent a pixel-level comparison, their visual quality is similar, with no apparent degradation from uint8 responses.

step 1 dump_rollout_generations, float32 vs uint8

step 50 dump_rollout_generations, float32 vs uint8

#### Validation

Completed validation:

- [x] Run the Qwen-Image OCR FlowGRPO FSDP2 benchmark on the cluster and compare the uint8 implementation with the historical float32-response baseline.
- [x] Run the CPU unit tests.
- [x] Run the single-GPU smoke tests.
- [x] Manually inspect rollout generations produced by the uint8 response path.

### API and Usage Example

There is no new configuration option. Pixel-valued `responses` and the corresponding custom-reward input are always `torch.uint8` tensors in `[0, 255]`. This is a breaking change from the historical normalized floating-point contract. Latent-valued rollout output is not quantized and remains floating point.

<details>
<summary><strong>Superseded option-3 API and usage</strong></summary>

The original draft proposed a rollout option defaulting to `float32`, so existing jobs would retain the historical full-precision custom-reward input:

```yaml
actor_rollout_ref:
  rollout:
    response_transport_dtype: float32
```

Enable compact pixel transport with:

```bash
actor_rollout_ref.rollout.response_transport_dtype=uint8
```

Supported values are:

| Value | Pixel-valued `responses` | Custom reward input | Compatibility |
| --- | --- | --- | --- |
| `float32` | Normalized floating point in `[0, 1]` | Historical full-precision float tensor | Default behavior |
| `uint8` | Integer pixels in `[0, 255]` | The same `uint8` tensor; no conversion back to float | Opt-in contract change |

Latent-valued rollout output is not quantized and remains floating point regardless of this option.

</details>

#### Implementation

- Define pixel-valued rollout `responses` as `torch.uint8` in `[0, 255]` without a configuration switch.
- Quantize pixel output once at the rollout boundary with clamping, multiplication by 255, rounding, and conversion to `torch.uint8`.
- Keep latent output in floating point.
- Pass the uint8 representation directly to custom reward functions without compatibility restoration to the historical float contract.
- Make built-in image, video, reward, and logging paths consume the fixed uint8 contract directly; ImageBind converts pixels back to normalized floating point in its model-specific preprocessing.
- Pass explicit `uint8` images to `wandb.Image(..., normalize=False)` so W&B does not perform another implicit normalization or conversion.

<details>
<summary><strong>Superseded option-3 implementation</strong></summary>

- Add `actor_rollout_ref.rollout.response_transport_dtype` with allowed values `float32` and `uint8`; default to `float32`.
- For pixel output only, quantize at the rollout boundary with clamping, multiplication by 255, rounding, and conversion to `torch.uint8`.
- Keep latent output in floating point.
- Do only forward conversion. No downstream path converts `uint8` responses back to normalized float.
- Allow built-in image/reward serialization paths to consume either historical float pixels or already-quantized `uint8` pixels.
- Pass the configured representation directly to custom reward functions.
- Pass explicit `uint8` images to `wandb.Image(..., normalize=False)` so W&B does not perform another implicit normalization or conversion.

</details>

#### Custom reward contract decision

Quantizing at rollout creates a contract decision for custom reward functions. Three approaches were considered:

1. **Convert `uint8` back to normalized float before invoking custom rewards.** This preserves the historical dtype and range but cannot recover the precision discarded by quantization, and adds a reverse conversion.
2. **Always expose `uint8` to custom rewards.** This gives one compact representation throughout the downstream pipeline but changes the custom reward interface globally.
3. **Make transport configurable and preserve the historical default.** Existing jobs continue to receive full-precision `float32`; jobs that opt into `uint8` accept that custom rewards receive `[0, 255]` integer pixels.

This PR implements **option 2**, as selected in the weekly-meeting discussion. It makes uint8 the pixel-response and custom-reward interface unconditionally and deliberately does not hide the lossy transport by reconstructing a normalized float tensor before invoking custom rewards. This is an intentional breaking change called out in the PR title and API documentation and covered by tests for the new uint8 contract.

<details>
<summary><strong>Superseded option-3 conclusion</strong></summary>

The superseded design implemented option 3. In particular, it deliberately did not hide the lossy transport by reconstructing a float tensor downstream. It requested feedback on the compatibility/performance trade-off and whether the custom reward contract should be documented more formally.

</details>

#### Rounding and W&B behavior

The implementation uses round-to-nearest when converting normalized pixels to `uint8`:

```python
pixels.mul(255).round().to(torch.uint8)
```

Historically, handing normalized float tensors directly to `wandb.Image` made W&B scale them by 255 and cast them to `uint8`, truncating fractional values rather than applying this rounding rule. With a single explicit `uint8` representation and `normalize=False`, the image observed by W&B is exactly the representation used by verl-omni's other downstream image consumers. Although this can change values near quantization boundaries relative to historical W&B logging, the unified behavior is more internally consistent and is the intended behavior of this PR.

#### Out of scope

- Worker-specific field registration or removal of unused `responses` from `old_log_prob` and `update_actor` payloads.
- Quantization of latent rollout outputs.
- Automatic `uint8`-to-float restoration for custom rewards.

> [!IMPORTANT]

  - All checks relevant to this PR passed. The `autogen-trainer-cfg` hook also reported a pre-existing mismatch between `omni_trainer.yaml` and `_generated_omni_trainer.yaml` on `upstream/main`; this PR does not modify either file or attempt to resolve that unrelated issue.
- [x] Add unit or end-to-end test(s) to [the CI workflow](https://github.com/verl-project/verl-omni/tree/main/.github/workflows) to cover all the code. The new CPU unit tests follow the existing `*_on_cpu.py` naming convention and are automatically discovered by the current CPU test workflow, so no workflow change is required.


jiefangxuanyan:
As discussed in the [2026/8/11 weekly meeting](https://docs.google.com/document/d/1UD8IbrFWIjJGCs1MF2yjPAV5xnXKdmIDsj7UchqhHrc/edit?tab=t.0), we prefer option 2 (Always expose `uint8` to custom rewards).

knlnguyen1802 (commented):
Could you add more statistic on loss, reward for first 50 steps of your experiment ?

verl_omni/workers/rollout/vllm_rollout/vllm_omni_async_server.py
```diff
@@ -69,6 +69,22 @@ def _drop_none_mapping_values(value: Any) -> Any:
...
+        output_type = (sampling_params.get("extra_args") or {}).get("output_type")
+    return output_type or "image"
+
+
+def _pixel_output_to_uint8(output: torch.Tensor) -> torch.Tensor:
+    """Quantize a rollout pixel tensor from float ``[0, 1]`` to uint8 once."""
+    if output.dtype == torch.uint8:
+        return output
+    output = output.detach().to(dtype=torch.float32)
+    output = torch.nan_to_num(output, nan=0.0, posinf=1.0, neginf=0.0).clamp_(0, 1)
```
zhtmike: throw nan as error directly, refer to https://github.com/verl-project/verl-omni/issues/388

verl_omni/utils/tracking.py
```diff
@@ -165,7 +165,7 @@ def wrap_val_samples_for_wandb(samples, fps=24, output_dir=None):
             media_to_log[media_key] = wandb.Video(video_path, format="mp4")
             media = media_key
         else:
-            media = wandb.Image(out.float(), file_type="jpg")
+            media = wandb.Image(out, file_type="jpg", normalize=False)
```
zhtmike, line 170: may need a verification on wandb logging, cc @NancyFyong

verl_omni/utils/reward_score/reward_utils.py
```diff
@@ -22,17 +22,15 @@
...
     PIL (not NumPy) frames avoid ``export_to_video`` rescaling already-uint8 input
     by 255, which would invert colors modulo 256.
     """
     if video.ndim != 4 or video.shape[1] != 3:
         raise ValueError(f"Expected an RGB video tensor with shape [T, 3, H, W], got {tuple(video.shape)}")
 
-    video = video.detach().permute(0, 2, 3, 1).to(dtype=torch.float32)
-    video = torch.nan_to_num(video, nan=0.0, posinf=1.0, neginf=0.0).clamp_(0, 1)
-    frames = video.mul_(255).round_().to(dtype=torch.uint8, device="cpu").contiguous().numpy()
+    frames = video.detach().permute(0, 2, 3, 1).to(device="cpu").contiguous().numpy()
```
zhtmike, line 35: add a dtype check after entering this func, throw error if it is not uint8

jiefangxuanyan:
@knlnguyen1802 Here are the actor and critic statistics for the first 50 steps, comparing the uint8 and float32 runs.

**Actor**

actor screenshot


**Critic**

critic screenshot

zhtmike:
and might need to update the doc under `docs/contributing` and `.agents/rules/reward.md` as well

knlnguyen1802:
> and might need to update the doc under `docs/contributing` and `.agents/rules/reward.md` as well

right cc: @AndyZhou952  for this

verl_omni/workers/rollout/vllm_rollout/vllm_omni_async_server.py
jiefangxuanyan: Thanks. I’ve changed it so that NaN, `+Inf`, and `-Inf` all raise a `ValueError`. Finite values outside `[0, 1]` are still clamped before quantization. Would this behavior be acceptable?

verl_omni/utils/reward_score/reward_utils.py
jiefangxuanyan: I found several places in the project that need the same dtype validation, so I added checks for W&B logging, generation dumping, reward processing, and the reusable video conversion helpers. Reward validation is handled at the common entry point.

zhtmike:
> and might need to update the doc under `docs/contributing` and `.agents/rules/reward.md` as well

right, pls fix this~

verl_omni/utils/tracking.py
jiefangxuanyan: Added an automated test using the real W&B SDK in offline mode. It passes a synthetic `torch.uint8` CHW image through `wrap_val_samples_for_wandb`, calls `wandb.log`, verifies that the run is offline and produces exactly one JPEG, then decodes the JPEG and checks its RGB values within JPEG compression tolerance. No W&B account or network access is required. Tested with W&B 0.28.1.

zhtmike (approved):

verl_omni/utils/reward_score/reward_utils.py
jiefangxuanyan: Correction: the implementation previously pushed in 56d40af3 incorrectly rejected floating-point latent responses; 75828892 fixes this by validating `responses` according to the active rollout `output_type`.

jiefangxuanyan:
> and might need to update the doc under `docs/contributing` and `.agents/rules/reward.md` as well

@zhtmike Addressed in 75828892: the response contract is documented in `docs/contributing/integrating_a_diffusion_model.md` under “4.1 Rollout response contract”, with the corresponding agent guidance updated in `.agents/rules/reward.md` and `.agents/skills/add-reward-score/SKILL.md`.

knlnguyen1802:
Please resolve conflict after vllm 0.27.0 update and fix cpu test

verl_omni/workers/rollout/vllm_rollout/vllm_omni_async_server.py
```diff
@@ -69,6 +69,24 @@ def _drop_none_mapping_values(value: Any) -> Any:
...
+
+def _pixel_output_to_uint8(output: torch.Tensor) -> torch.Tensor:
+    """Quantize a rollout pixel tensor from float ``[0, 1]`` to uint8 once."""
+    if output.dtype == torch.uint8:
+        return output
+    output = output.detach().to(dtype=torch.float32)
+    if not bool(torch.isfinite(output).all()):
+        raise ValueError("Pixel rollout output must contain only finite values")
+    output = output.clamp_(0, 1)
+    return output.mul_(255).round_().to(dtype=torch.uint8)
```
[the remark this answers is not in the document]
jiefangxuanyan: Fixed in 4a9cc52. I used `to(dtype=torch.float32, copy=True)` to avoid mutating float32 inputs without adding a redundant post-conversion clone for other dtypes, and added regression coverage for both float16 and float32 inputs.

jiefangxuanyan:
@knlnguyen1802 I’ve resolved the merge conflicts with the latest `main`.

The previous CPU test failure was caused by vLLM-Omni using `StrEnum`, which is unavailable in Python 3.10. The newly pinned vLLM-Omni version has replaced that code with a Python 3.10-compatible implementation, so the tests are expected to pass directly now.

PR #375, opened 2026-08-11
[trainer, rollout, recipe, doc] feat: support OPD training for Qwen3-Omni model

wsc16:
### What does this PR do?

Adds online policy distillation (OPD) support for the Qwen3-Omni model on the omni V1 trainer, building on verl's OPD (teacher loop / distillation) implementation. Extends verl's `MultiTeacherModelManager` to work with `vllm_omni` inference for OPD training of the Qwen3-Omni model.

### Checklist Before Starting

- [x] Search for similar PRs: https://github.com/verl-project/verl-omni/pulls?q=opd (no existing Qwen3-Omni OPD support)
- [x] Format the PR title as `[{modules}] {type}: {description}`
  - Title: `[trainer, rollout, recipe] feat: support OPD training for Qwen3-Omni model`

### Test

Validated by experiment on MMK12:

- **Setup**: the student is Qwen3-Omni-30B-A3B-Instruct with 25% noise added to its weights; the teacher is the original (un-noised) Qwen3-Omni-30B-A3B-Instruct, served by `vllm_omni` in AR mode.
- **Baseline**: the noised model self-trained (GSPO, no distillation).
- **Result**: OPD training runs end-to-end with `distillation.enabled=true` (KL distillation loss, teacher prompt logprobs flowing into the loss) and outperforms the self-training baseline.
- **Environment**: validated on 2x Ascend 910C machines — student rollout/actor on 16 GPUs of node 1, teacher model on 16 GPUs of node 2.

Launch the task as follows (replace `<head_ip>` and `<port>` with the actual values from the head node):

1. On the master node (node 1): `ray start --head`
2. On the slave node (node 2): `ray start --address='<head_ip>:<port>'`
3. Run this script on the master node to start the training task.

The OPD training results compared with the baseline are shown below — OPD converges faster than the self-training baseline:
image
image

### API and Usage Example

No breaking API changes. New config surface is the existing `distillation.*` group pointed at a `vllm_omni` teacher:

```bash
distillation.enabled=true \
distillation.teacher_models.teacher_model.model_path=<teacher_path> \
distillation.teacher_models.teacher_model.inference.name=vllm_omni \
distillation.teacher_models.teacher_model.inference.vllm_omni.output_mode="ar" \
distillation.teacher_models.teacher_model.inference.vllm_omni.pipeline_name="qwen3_omni_moe" \
distillation.distillation_loss.loss_mode=kl \
distillation.distillation_loss.use_policy_gradient=true
```

Full runnable script: `examples/gspo_trainer/qwen3_omni/run_qwen3_omni_thinker_gspo_lora_mmk12_v1_opd_npu.sh`

### Design & Code Changes

The default verl teacher manager assumes a text-only HF model (`HFModelConfig`). Omni teachers need the full processor pipeline (vision/audio towers, chat template, `dedup_pad_tokens`), hence an omni-specific manager rather than patching verl core.

- **`verl_omni/workers/config/omni/distillation.py`** (new): `OmniDistillationTeacherModelConfig` subclasses verl's `DistillationTeacherModelConfig` and overrides `_validate_topk_logprobs` to add a `case "vllm_omni"` branch (mirrors the `vllm` branch: seeds `engine_kwargs["vllm_omni"]["max_logprobs"]=topk` when unset, validates `>= topk`). `OmniDistillationConfig` subclasses verl's `DistillationConfig` and overrides `_resolve_teacher_models` to materialize each teacher via `OmniDistillationTeacherModelConfig` (verl hardcodes the base class at the `omega_conf_to_dataclass` call site, so inheritance alone is insufficient). `build_omni_distillation_config` builds the omni config via `hydra.instantiate` with `_target_` injected — works around OmegaConf's `unions of containers are not supported` error on `DistillationLossSettings.names: Union[str, list[str]]` that `omega_conf_to_dataclass(dataclass_type=OmniDistillationConfig)` would hit via `OmegaConf.structured`.
- **`verl_omni/workers/utils/padding.py`**: `patched_padding_template` — wraps verl's minimal padding template to extend teacher-side fields (`teacher_ids` filled with eos, `teacher_logprobs` zeroed) to the padded sequence length. Temporary; will be removed once verl pads these fields natively.
- **`verl_omni/workers/rollout/vllm_rollout/vllm_omni_async_server.py`**: `vLLMOmniHttpServer` extracts prompt logprobs (`extract_prompt_logprobs`) on the AR path so teacher distributions can be scored on student rollouts; `vLLMOmniReplica` accepts `is_teacher_model` / `name_suffix` for colocated teacher replicas.
- **`examples/gspo_trainer/qwen3_omni/run_qwen3_omni_thinker_gspo_lora_mmk12_v1_opd_npu.sh`** (new): end-to-end GSPO + LoRA OPD recipe on MMK12 for Ascend NPU, documents runtime deps (`math-verify`, `qwen-vl-utils`) and data prep; defaults `STUDENT_MODEL` to `Qwen3-Omni-30B-A3B-Instruct-Noised` and `TEACHER_MODEL` to `Qwen3-Omni-30B-A3B-Instruct`.
- **`examples/gspo_trainer/README.md`**: adds an "MMK12 On-Policy Distillation (OPD)" subsection documenting the recipe and launch steps.

Limitation: teacher and student must share the same tokenizer (same model family). Cross-vocabulary distillation not supported yet. The agent-loop `AsyncTeacherLLMServerManager` path still uses verl's base `DistillationConfig`; `forward_kl_topk` + `vllm_omni` would need additional work there, but the current `loss_mode=kl` recipe does not trigger it.

- [ ] Add unit or end-to-end test(s) to [the CI workflow](https://github.com/verl-project/verl-omni/tree/main/.github/workflows) to cover all the code. Not feasible: requires multi-GPU vLLM omni teacher + student cluster; validated by real training run instead (see Test section).

🤖 Generated with [Claude Code](https://claude.com/claude-code)


SamitHuang:
@zhtmike @cr-gao PTAL thx

zhtmike:
little busy on #492, @knlnguyen1802 @NancyFyong pls take a look~

SamitHuang (changes requested):
## Review (requesting changes)

OPD-on-omni is the right direction (OmniModelConfig + prompt-logprob extraction + replica teacher naming), but the wiring is too fragile to merge as-is. No CPU tests were added for the new manager or the padding patch; a 2-node training run does not cover the failure modes below.

### 1. `OmniPPOTrainerSync.init` — [ray_omni_trainer.py](https://github.com/verl-project/verl-omni/pull/375/files#diff-db9a7d8211f40080ad54ece52b0d443ccf4aa380ae49a4c98bbc3db7f531079aR78)
Toggling `use_teacher_policy` around `super().init()` and then monkey-patching `construct_minimal_padding_template` is a maintenance trap. `PPOTrainer._setup` already creates the agent/LLM server path and later `get_teacher_client()` reads `self.use_teacher_policy`; any new teacher work in `_setup` / `on_init_end` will silently skip. Prefer a hook that only swaps `HFModelConfig` → `OmniModelConfig` (or pass a manager class) instead of disabling the whole teacher branch. `omega_conf_to_dataclass(self.config.distillation)` also runs a second time after `OmniMultiTeacherModelManager` already converted it — `validate_and_prepare_for_distillation` mutates `prompt_length` / `response_length`; if that conversion writes back into OmegaConf, the second pass doubles the prompt length.

### 2. Copied teacher manager regresses verl — [teacher_model.py](https://github.com/verl-project/verl-omni/pull/375/files#diff-ff52dccba6261f50fe29f0d31586138dc3825e43cdbf83b421549097396c58c9R81)
`_initialize_llm_servers` / `_initialize_teacher_model_managers` duplicate verl's `TeacherModelManager` almost line-for-line but drop the pool-size `ValueError` and use `zip(..., strict=False)`. A pool/replica mismatch will spawn a truncated teacher set instead of failing. Override only the model-config construction and keep `strict=True`. Also add a CPU test that `vllm_omni` → `OmniModelConfig` and other engines still get `HFModelConfig`.

### 3. Padding helper is untested and imported globally — [padding.py](https://github.com/verl-project/verl-omni/pull/375/files#diff-719816b4d3b5ba882c4175cc7170e5042853f003e717e620f9a1689eb5d3593bR25)
`verl_omni/workers/utils/padding.py` is used by diffusion training; importing it now pulls `verl.trainer.ppo.padding_utils` at module import. Extend `tests/utils/test_padding_on_cpu.py` with a case where the source sample has `teacher_ids` / `teacher_logprobs` longer than the 2-token template and assert the patched sample matches `input_ids` length. The `new_full` / `new_zeros` rewrite is fine for synthetic pads, but it needs a regression test because this was already a production crash.

### 4. Recipe is not runnable as copied — [run_qwen3_omni_thinker_gspo_lora_mmk12_v1_opd.sh](https://github.com/verl-project/verl-omni/pull/375/files#diff-311c0ab1bb90c0945ae27e22c0fc0f9516b2d85ced0eb4db79a12a1eaa9386f5R28)
`STUDENT_MODEL` / `TEACHER_MODEL` default to `""`, unlike the existing MMK12 script which defaults `MODEL_PATH`. Document the same `$HOME/models/...` default (or fail fast if unset). The script is NPU-only (`VLLM_ASCEND_ENABLE_NZ`) but is not named `_npu`, and `examples/gspo_trainer/README.md` is not updated despite the PR checklist. `DistillationTeacherModelConfig._validate_topk_logprobs` still only allows `vllm` / `sglang`; this recipe uses `loss_mode=kl` so it slips through, but `forward_kl_topk` with `inference.name=vllm_omni` will raise `NotImplementedError` — either add a `vllm_omni` case (same `max_logprobs` as vLLM) or document that only estimator losses are supported.

### 5. Prompt logprobs — [vllm_omni_ar_strategy.py](https://github.com/verl-project/verl-omni/pull/375/files#diff-3bbca6a8e1919df55b632058b259cc881bb59d197b22a87f6944eaebca7c635cR241)
`extract_prompt_logprobs` is required: `AsyncTeacherLLMServerManager.compute_teacher_logprobs_single` reads `extra_fields["prompt_ids"]` / `prompt_logprobs`. Keep this on the AR output path after rebase onto the strategy split. A CPU test that a fake `RequestOutput` with `prompt_logprobs` fills `TokenOutput.extra_fields` would lock this in without a cluster.

wsc16:
> ## Review (requesting changes)
> OPD-on-omni is the right direction (OmniModelConfig + prompt-logprob extraction + replica teacher naming), but the wiring is too fragile to merge as-is. No CPU tests were added for the new manager or the padding patch; a 2-node training run does not cover the failure modes below.
> 
> ### 1. `OmniPPOTrainerSync.init` — [ray_omni_trainer.py](https://github.com/verl-project/verl-omni/pull/375/files#diff-db9a7d8211f40080ad54ece52b0d443ccf4aa380ae49a4c98bbc3db7f531079aR78)
> Toggling `use_teacher_policy` around `super().init()` and then monkey-patching `construct_minimal_padding_template` is a maintenance trap. `PPOTrainer._setup` already creates the agent/LLM server path and later `get_teacher_client()` reads `self.use_teacher_policy`; any new teacher work in `_setup` / `on_init_end` will silently skip. Prefer a hook that only swaps `HFModelConfig` → `OmniModelConfig` (or pass a manager class) instead of disabling the whole teacher branch. `omega_conf_to_dataclass(self.config.distillation)` also runs a second time after `OmniMultiTeacherModelManager` already converted it — `validate_and_prepare_for_distillation` mutates `prompt_length` / `response_length`; if that conversion writes back into OmegaConf, the second pass doubles the prompt length.
> 
> ### 2. Copied teacher manager regresses verl — [teacher_model.py](https://github.com/verl-project/verl-omni/pull/375/files#diff-ff52dccba6261f50fe29f0d31586138dc3825e43cdbf83b421549097396c58c9R81)
> `_initialize_llm_servers` / `_initialize_teacher_model_managers` duplicate verl's `TeacherModelManager` almost line-for-line but drop the pool-size `ValueError` and use `zip(..., strict=False)`. A pool/replica mismatch will spawn a truncated teacher set instead of failing. Override only the model-config construction and keep `strict=True`. Also add a CPU test that `vllm_omni` → `OmniModelConfig` and other engines still get `HFModelConfig`.
> 
> ### 3. Padding helper is untested and imported globally — [padding.py](https://github.com/verl-project/verl-omni/pull/375/files#diff-719816b4d3b5ba882c4175cc7170e5042853f003e717e620f9a1689eb5d3593bR25)
> `verl_omni/workers/utils/padding.py` is used by diffusion training; importing it now pulls `verl.trainer.ppo.padding_utils` at module import. Extend `tests/utils/test_padding_on_cpu.py` with a case where the source sample has `teacher_ids` / `teacher_logprobs` longer than the 2-token template and assert the patched sample matches `input_ids` length. The `new_full` / `new_zeros` rewrite is fine for synthetic pads, but it needs a regression test because this was already a production crash.
> 
> ### 4. Recipe is not runnable as copied — [run_qwen3_omni_thinker_gspo_lora_mmk12_v1_opd.sh](https://github.com/verl-project/verl-omni/pull/375/files#diff-311c0ab1bb90c0945ae27e22c0fc0f9516b2d85ced0eb4db79a12a1eaa9386f5R28)
> `STUDENT_MODEL` / `TEACHER_MODEL` default to `""`, unlike the existing MMK12 script which defaults `MODEL_PATH`. Document the same `$HOME/models/...` default (or fail fast if unset). The script is NPU-only (`VLLM_ASCEND_ENABLE_NZ`) but is not named `_npu`, and `examples/gspo_trainer/README.md` is not updated despite the PR checklist. `DistillationTeacherModelConfig._validate_topk_logprobs` still only allows `vllm` / `sglang`; this recipe uses `loss_mode=kl` so it slips through, but `forward_kl_topk` with `inference.name=vllm_omni` will raise `NotImplementedError` — either add a `vllm_omni` case (same `max_logprobs` as vLLM) or document that only estimator losses are supported.
> 
> ### 5. Prompt logprobs — [vllm_omni_ar_strategy.py](https://github.com/verl-project/verl-omni/pull/375/files#diff-3bbca6a8e1919df55b632058b259cc881bb59d197b22a87f6944eaebca7c635cR241)
> `extract_prompt_logprobs` is required: `AsyncTeacherLLMServerManager.compute_teacher_logprobs_single` reads `extra_fields["prompt_ids"]` / `prompt_logprobs`. Keep this on the AR output path after rebase onto the strategy split. A CPU test that a fake `RequestOutput` with `prompt_logprobs` fills `TokenOutput.extra_fields` would lock this in without a cluster.

Thanks for review, and the above issues have been fixed:
1. Temporarily set use_teacher_policy=False before super().init(), so verl’s default MultiTeacherModelManager path is skipped; rebuild teacher manager in subclass.
2. Fixed. Restore zip(strict=True) and raise ValueError when sub-pool size mismatches expected (num_replicas * per_replica_world_size).
3. Minimal runtime monkey-patch only when use_teacher_policy to extend padding for teacher_ids/teacher_logprobs (eos/zeros). No unit test added per project guidance.
4. Fixed: Add OmniDistillationTeacherModelConfig._validate_topk_logprobs with case "vllm_omni"; use OmniDistillationConfig._resolve_teacher_models to instantiate the omni subclass.

verl_omni/experimental/teacher_loop/__init__.py
zhtmike, line 1: I think we should follow the design done by @cr-gao, instead of introducing multiple teacher again (seems duplicated)

Ping @cr-gao , can you provide some suggestions?

wsc16: it is currently derived from the base class in Verl and we will expand it to support the full-vocab OPD, which may differ from the diffusion model. We can consider unifying them later after full-vocab OPD is supported to avoid interfering with the process of the diffusion model.

zhtmike: I see, it is better to share the same base class (especially the model manager part, since it is only responsible for the worker arrangement, and does not touch the core algo. or data pipeline) for all opd models.

I am fine with current implementation then.

verl_omni/utils/config.py
```diff
@@ -38,3 +43,14 @@ def validate_config(config: Any) -> None:
...
+
+def omni_omega_conf_to_dataclass(config: DictConfig | dict, dataclass_type: type[Any]):
+    """Dataclass materializer with OmegaConf-union fallback."""
+    try:
+        return _base_omega_conf_to_dataclass(config, dataclass_type=dataclass_type)
+    except OmegaConfBaseException:
+        cfg = OmegaConf.create(config)
+        if "_target_" not in cfg:
+            cfg["_target_"] = f"{dataclass_type.__module__}.{dataclass_type.__qualname__}"
+        return instantiate(cfg, _convert_="partial")
```
zhtmike: why do we need this?

wsc16: the original method will throw an error when the dataclass_type is specified

zhtmike: for example?

wsc16: In the specified "dataclass_type" branch, "loss_settings" is declared as a dict, but it will actually be passed as a "LossSettings", which will result in a validation failure. this issue can be fixed directly in "verl" later.

zhtmike: I think we can just overwrite the corresponding part of yaml file or DistillationLossConfig, instead of adding this.

verl_omni/workers/utils/padding.py
```diff
@@ -11,10 +11,34 @@
...
+    sample, tag = ori_padding_template(source_td, source_tag, eos_token_id)
+    sequence_length = sample["input_ids"].size(0)
+
+    teacher_ids = sample.get("teacher_ids")
+    if isinstance(teacher_ids, torch.Tensor):
+        sample["teacher_ids"] = teacher_ids.new_full((sequence_length, *teacher_ids.shape[1:]), eos_token_id)
+
+    teacher_logprobs = sample.get("teacher_logprobs")
+    if isinstance(teacher_logprobs, torch.Tensor):
+        sample["teacher_logprobs"] = teacher_logprobs.new_zeros((sequence_length, *teacher_logprobs.shape[1:]))
```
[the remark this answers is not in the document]
wsc16: The sequence_length is equal to teacher_ids.shape[0]. A dummy sample was constructed for the balance_batch scenario. All parameters were dummy data, but the teacher logprobs were omitted here.

verl_omni/utils/config.py
wsc16: Yes, it's better. I will fix

verl_omni/experimental/teacher_loop/__init__.py
cr-gao: I think we do need separate teacher managers for diffusion and omni opd. But it looks like OmniTeacherModelManager and OmniMultiTeacherModelManager mostly copies the upstream base class, and the only diff is building `OmniModelConfig(path=...)`. Since `vLLMOmniHttpServer` converts whatever it receives via the AR strategy's `init_model_config` (`model_config_cls = OmniModelConfig`), is it possible to directly use verl's managers?

cr-gao: @zhtmike Currently diffusion OPD doesn't have its own teacher model manager, the worker and resource pool is directly registered inside ray_diffusion_trainer.py. Diffusion teachers are fsdp forward on training side since vllm-omni doesn't provide good trajectory scoring interfaces, and for omni models teachers should be vllm-omni serving replicas, and therefore they may not share the same base class. 
It seems that we can directly reuse verl's TeacherModelManager on omni model so we don't need an extra Omni teacher manager class (needs verification). 
If we preserve that class and aim a unified interface then it's a refactor of the diffusion teachers as serving engines. In that case I can rewire it with some extra work on diffusion teacher manager side or after landing a vllm-omni upstream PR that provides the API.

zhtmike: > @zhtmike Currently diffusion OPD doesn't have its own teacher model manager, the worker and resource pool is directly registered inside ray_diffusion_trainer.py. Diffusion teachers are fsdp forward on training side since vllm-omni doesn't provide good trajectory scoring interfaces, and for omni models teachers should be vllm-omni serving replicas, and therefore they may not share the same base class. 
> 
> It seems that we can directly reuse verl's TeacherModelManager on omni model so we don't need an extra Omni teacher manager class (needs verification). 
> 
> If we preserve that class and aim a unified interface then it's a refactor of the diffusion teachers as serving engines. In that case I can rewire it with some extra work on diffusion teacher manager side or after landing a vllm-omni upstream PR that provides the API.

Agree. Let us separate the design for now.

zhtmike: > I think we do need separate teacher managers for diffusion and omni opd. But it looks like OmniTeacherModelManager and OmniMultiTeacherModelManager mostly copies the upstream base class, and the only diff is building `OmniModelConfig(path=...)`. Since `vLLMOmniHttpServer` converts whatever it receives via the AR strategy's `init_model_config` (`model_config_cls = OmniModelConfig`), is it possible to directly use verl's managers?

Ping @wsc16 , can you take a look?

wsc16: Absolutely! Current implementation is based on the adaptation of the previous version. I will modify it to completely reuse the implementation of verl.

verl_omni/workers/config/omni/distillation.py
```diff
@@ -0,0 +1,57 @@
...
+    def _validate_topk_logprobs(self, use_topk: bool, topk: Optional[int]) -> None:
+        if not use_topk:
+            return
+        if topk is None:
+            raise ValueError("topk must be specified when use_topk is True.")
+
+        engine_name = self.inference.name
+        engine_kwargs = self.inference.engine_kwargs
+        match engine_name:
+            case "vllm_omni":
```
knlnguyen1802: Just have question here, in what case it will fall back to other engine ?

wsc16: Theoretically No. This can be changed to throw an exception here. Thx~

verl_omni/experimental/teacher_loop/__init__.py
wsc16: fixed

verl_omni/utils/config.py
wsc16: fixed

verl_omni/workers/config/omni/distillation.py
wsc16: fixed

verl_omni/workers/config/omni/distillation.py
```diff
@@ -0,0 +1,58 @@
...
+                max_logprobs = omni_engine_kwargs.get("max_logprobs")
+                if max_logprobs is None:
+                    omni_engine_kwargs["max_logprobs"] = topk
+                    max_logprobs = topk
+                if max_logprobs < topk:
+                    raise ValueError(
+                        f"vllm_omni max_logprobs ({max_logprobs}) must be >= distillation_loss topk "
+                        f"({topk}) to enable distillation loss computation."
+                    )
+                engine_kwargs["vllm_omni"] = omni_engine_kwargs
```
zhtmike: drop compatibility code change @wsc16 , no need get(xxx, xxx) fallback

wsc16: what if max_logprobs is not set?

verl_omni/workers/utils/padding.py
```diff
@@ -70,3 +72,31 @@ def _to_nested(values: torch.Tensor, mask: torch.Tensor | None, key: str):
...
+
+    return sample, tag
+
+
+# Patch the module attribute so upsample_batch_to_divisible_size resolves the
+# patched template through verl's padding_utils namespace.
+_padding_utils.construct_minimal_padding_template = patched_padding_template
```
zhtmike, line 113: last thing, better to enable the patch only for opd, I don't think a global patch is good

verl_omni/workers/config/omni/distillation.py
```diff
@@ -0,0 +1,52 @@
...
+
+    def _validate_topk_logprobs(self, use_topk: bool, topk: Optional[int]) -> None:
+        if not use_topk:
+            return
+        if topk is None:
+            raise ValueError("topk must be specified when use_topk is True.")
+
+        assert self.inference.name == "vllm_omni", (
+            f"OmniDistillationTeacherModelConfig only supports inference.name == 'vllm_omni', got {self.inference.name}"
+        )
```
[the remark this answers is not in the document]
wsc16: changed to an if statement.

verl_omni/workers/rollout/vllm_rollout/npu_utils.py
```diff
@@ -102,18 +102,15 @@ def restore_moe_param_layout(model, hidden_size: int) -> None:
...
-        if data.ndim != 3:
-            continue
-        if "w2_weight" in name and data.shape[2] == hidden_size:
+    for name, param in model.named_parameters():
+        if "w2_weight" in name and param.shape[2] == hidden_size:
             parts = name.split(".")
             parent_module = model.get_submodule(".".join(parts[:-1]))
-            replace_parameter(parent_module, parts[-1], data.transpose(1, 2).contiguous())
-        elif "w13_weight" in name and data.shape[1] == hidden_size:
+            replace_parameter(parent_module, parts[-1], param.transpose(1, 2))
```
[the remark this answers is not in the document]
wsc16: fixed

verl_omni/workers/utils/padding.py
```diff
@@ -70,3 +76,36 @@ def _to_nested(values: torch.Tensor, mask: torch.Tensor | None, key: str):
...
+    The stock template only knows about student-token fields; teacher_ids /
+    teacher_logprobs ride along on the batch and must be extended to the padded
+    sequence length with eos / zeros respectively.
+    """
+    sample, tag = ori_padding_template(source_td, source_tag, eos_token_id)
+    logger.warning(
+        "Using patched_padding_template to pad teacher_ids / teacher_logprobs. "
+        "This is a temporary workaround; remove once verl's padding_utils pads "
+        "teacher-side fields natively."
+    )
```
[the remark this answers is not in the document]
wsc16: padding is needed only in balance_batch when the train_batch_size cannot be evenly divided by micro_batch.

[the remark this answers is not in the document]
zhtmike: but I agree it will warn many times, it will pollute the logging message

zhtmike (approved):

SamitHuang (changes requested):
Please update related docs before merge
1. Add OPD algo doc referring to https://verl-omni.readthedocs.io/en/latest/algo/diffusion_opd.html
2. update the models doc (https://verl-omni.readthedocs.io/en/latest/start/models.html) and performance reference doc (https://verl-omni.readthedocs.io/en/latest/algo/performance.html)

SamitHuang (approved):
LGTM

zhtmike:
All CI green.

PR #337, opened 2026-08-03
[trainer, tests, recipe] feat: add omni colocate-async V1 trainer

zhtmike:
### What does this PR do?

Phase 1 of [#320]: implement the `colocate_async` orchestration mode for the omni V1 trainer.

> **Update 2026-09-18:** rebased on current `main` and addressed review. The `vllm_omni_async_server.py` hunk is dropped (superseded by vllm-omni #497, so the `fix_sleep_v0.24.0` tail branch is no longer needed), and the `v1.omni_colocate_async` config stub + `on_train_begin` override are removed — colocate-async now reads the generic `trainer.v1.colocate_async` keys per #320 §3.6. CPU tests pass.

- Add `OmniPPOTrainerColocateAsync` with `@register_trainer("omni_colocate_async")`, a thin subclass of verl's `PPOTrainerColocateAsync`. The only override is `_init_tokenizer`, which wires `OmniModelConfig.tokenizer/processor` (identical to `OmniPPOTrainerSync`). Warmup (`on_train_begin`) and all async lifecycle hooks (abort/sleep/resume, weight sync, `FullyAsyncLLMServerClient`) are inherited from verl and read the generic `trainer.v1.colocate_async.num_warmup_batches` — no `v1.omni_*` config stub (#320 §3.6).
- Export `OmniPPOTrainerColocateAsync` in `verl_omni/trainer/omni/__init__.py` so `@register_trainer` fires on import.
- Add example script `run_qwen3_omni_thinker_gspo_lora_mmk12_colocate_async_v1.sh` — a small delta from the sync MMK12 reference (`trainer_mode`, `num_warmup_batches`, `experiment_name`). MMK12 is a multimodal thinking dataset with a measured long-tail response-length distribution (10 → 12288 tokens, 4.4% clip ratio on Qwen3-Omni-30B thinker), which is the workload colocate_async overlaps.
- Add CPU unit tests covering trainer registration, `_init_tokenizer` wiring, and the inherited `on_train_begin` warmup + skip guard.

Separate_async (`OmniDetachActorWorker`, `adapter_name` forwarding fix, separate_async config + example) has landed separately as #380; the two modes share no trainer-level code.

> Add **concise** overview of what this PR aims to achieve or accomplish. Reference related GitHub issues and PRs that help with the review.

### Checklist Before Starting

- [ ] Search for similar PRs. Paste at least one query link here: ...
- [ ] Format the PR title as `[{modules}] {type}: {description}` (This will be checked by the CI)
[...]

Issue #320, opened 2026-07-30
[RFC] Omni V1 Trainer: Colocate Async and Separate Async Modes


zhtmike:
Status: **updated 2026-09-03, ready to implement PR 2 / PR 3** — the engine foundation this RFC was blocked on is restored: #491 step 1 (sub-RFC #492: vllm-omni pin bump + async rollout server lifecycle redesign) landed in #497, merged 2026-09-03. This RFC now tracks the remaining PR 2 (`omni_colocate_async`) and PR 3 (`omni_separate_async`) of #491's stack.
Owners: @zhtmike — colocate_async (PR 2, landing vehicle: rebase of #337) · @AndyZhou952 — separate_async (PR 3, landing vehicle: rebase of #380) — per the split recorded in this issue's comments.
Design locked at the level described here. One open design question — resume ownership vs #517 — is called out in §6.3 and must be settled before PR 3's worker hunks are finalized.

---

## Revision note (2026-09-03) — what changed vs the 2026-07-30 body

The original body was written against the old vllm-omni pin `4444856`, before #491's stack landed. It has been rewritten against the shipped state of `main` @ `332cb352` (#497). Substantive corrections:

1. **Engine contract replaced wholesale.** The three blocking engine defects (unacked abort, frontend-flag-only pause, discarded partial tokens) are fixed upstream (vllm-omni #6084 / #6367 / #6327) and consumed by #497's redesigned server. The old "Verified Compatibility" table described pre-bump semantics (e.g. bulk abort = `pause_generation` only) that are now wrong and is replaced by §2.
2. **Resume ownership moved server-side (deviation from sub-RFC #492, shipped in #497).** Every ACK-validated `wake_up` now resumes admission inside the server (`vllm_omni_async_server.py:217`), restoring upstream verl parity — AR trainers work unmodified. The trainer-visible hold now lives **only on the abort path**, lifted by `resume_generation_replicas()`. The `omni_sync` bridge (`ray_omni_trainer.py:78-84`) stays as an init-time safety net. PR 2 / PR 3 add **no new per-trainer bridges**; see §6.3 for the #517 coordination.
[...]

PR #380, opened 2026-08-12
[trainer, cfg, tests, recipe] feat: add omni separate-async V1 trainer

AndyZhou952:
### What does this PR do?

Phase 2 of RFC #320 and follow-up to #337: add `trainer.v1.trainer_mode=omni_separate_async` for the omni V1 trainer (thin `PPOTrainerSeparateAsync` subclass + `OmniDetachActorWorker`). This is the #389 `omni_separate_async` landing vehicle, not a parallel path.

Duplicate check: `gh pr list --repo verl-project/verl-omni --state open --search "389 in:body"` / `"separate_async omni"` — this PR already owns the row.

Engine blockers from the Aug 14 thread (unacked abort, frontend-only pause, discarded partial tokens) were fixed by #497 (vllm-omni pin `ded89346`). This PR needs no server hunks.

### Review follow-up (zhtmike, 2026-09-08)

1. MMK12 example now satisfies `train_batch_size == parameter_sync_step * ppo_mini_batch_size` (`128 == 8 * 16`) and documents the constraint.
2. Dropped the dead `replay_buffer.parameter_sync_step` write; CPU test no longer asserts it.
3. Abort-semantics doc rewritten against RFC #320 §2.1 / post-#497 abort-then-pause with real partial-token continuation (`test_omni_rollout_recovery_on_cpu.py` budgets `[8, 6]`).
4. Decoupled PPO doc corrected: generated omni config ships `bypass_mode: false`, so `OmniDetachActorWorker` CPU save/restore is the live GPU path. `parameter_sync_step=8` engages it.
5. Dropped the smoke-only `pip3 install transformers==5.12.1 ...` pin.
6. PR body replaced (this text). Prior MMK12 curve from the thread is below; that run predates the `parameter_sync_step=8` identity fix, so the RFC #320 §6.2 convergence gate with Decoupled PPO engaged still needs a re-run on this recipe.

qwen3_async

Nice-to-have from review: `omni_sync` still uses verl's v1 actor worker, so LoRA adapter-only send now differs between omni modes — tracking separately (not in this PR).

### Test

CPU (registration, `parameter_sync_step` key, LoRA/detach wiring, recovery client, `adapter_name`):

```bash
pytest -s --asyncio-mode=auto \
    tests/trainer/omni/test_ray_omni_trainer_separate_async_on_cpu.py \
    tests/workers/rollout/test_omni_rollout_recovery_on_cpu.py \
    tests/workers/test_omni_fsdp_engine_on_cpu.py
```

GPU smoke (2 GPUs, tiny-random Thinker, 3 steps):

```bash
bash tests/special_e2e/run_gspo_qwen3_omni_thinker_lora_v1_separate_async_smoke.sh
```

Example:

```bash
bash examples/gspo_trainer/qwen3_omni/run_qwen3_omni_thinker_gspo_lora_mmk12_separate_async_v1.sh
```

### Design & Code Changes

- `@register_trainer("omni_separate_async")` thin subclass: tokenizer from `OmniModelConfig`, `parameter_sync_step` from `v1.separate_async`, LoRA-aware `OmniCheckpointEngineManager`.
- `OmniDetachActorWorker(ActorRolloutRefWorker, DetachActorWorker)`: omni LoRA sync wins the MRO; Detach supplies CPU save/restore for Decoupled PPO.
- `adapter_name` forwarded in `OmniFSDPEngine.get_per_tensor_param`.
- Thinker-only / single-stage AR (multi-stage abort still broken upstream).

- [x] Read the Contribute Guide.
- [x] Add / Update the documentation (`docs/algo/separate_async_omni.md`).
- [x] Add unit or end-to-end tests (CPU contracts + GPU smoke script).

AI assistance (Cursor) was used for this change. A human submitter reviewed every changed line.


zhtmike:
I have done an analysis with AI assistance. This PR is blocking until the following problems are fixed in vllm-omni side.
cc @SamitHuang @knlnguyen1802 

## 1. Async abort has no completion acknowledgment

`AsyncOmniEngine.abort_async()` (vllm_omni/engine/async_omni_engine.py:1423, pin fe478a95) just calls the fire-and-forget `abort()` and returns; there is no `AbortResultMessage` anywhere at the pin. Backend abort, route-binding release, and request cleanup all happen asynchronously, with no way for the caller to know when they finish. At every weight sync the trainer therefore proceeds to the NCCL weight overwrite while backend work from the aborted requests may still be running under old weights — mixed-weight samples and late outputs. The acknowledged-abort design exists in vllm-project/vllm-omni#6084; without it, abort failures are unobservable and the trainer cannot fail closed.

## 2. AR `pause_generation` is a frontend flag only

At the pin, `AsyncOmni.pause_generation()` (vllm_omni/entrypoints/async_omni.py:827) only sets `self._paused = True` on the frontend; `wait_for_inflight_requests` is an unimplemented TODO and nothing is forwarded to EngineCore. An empty `request_states` therefore proves nothing about the backend — there is no engine-side idle boundary before the weight overwrite, even when the ahead-batch finished early. This is the mechanism behind the "actor-updating takes extensive time" symptom: the trainer-side drain (up to 120 s) is compensating for an idle boundary the engine never provides.

## 3. StagePool abort discards partial tokens

`StagePool.abort_requests()` (vllm_omni/engine/stage_pool.py:1130, pin fe478a95) routes request IDs to replicas and deletes state, but never builds a terminal output from the output processor — no `finish_reason="abort"` output with the tokens generated so far. The AR client therefore resumes from the original prompt instead of `prompt + partial_tokens`, silently discarding the partial rollout at every hard abort. On MMK12's long tail this fires at most syncs, so partial-rollout continuation never happens and trajectory stats diverge from `omni_sync` — the likely cause of the convergence gap. Note: vllm-project/vllm-omni#6084 does not cover this one; it needs cumulative abort outputs for AR stages.

Suggested: track vllm-project/vllm-omni#6084 for 1 and 2, and file a separate vLLM-Omni issue for 3. Until then, hard-abort path behaves like a full restart and the sync boundary has no engine-side guarantee.

knlnguyen1802:
> I have done an analysis with AI assistance. This PR is blocking until the following problems are fixed in vllm-omni side. cc @SamitHuang @knlnguyen1802
> 
> ## 1. Async abort has no completion acknowledgment
> `AsyncOmniEngine.abort_async()` (vllm_omni/engine/async_omni_engine.py:1423, pin fe478a95) just calls the fire-and-forget `abort()` and returns; there is no `AbortResultMessage` anywhere at the pin. Backend abort, route-binding release, and request cleanup all happen asynchronously, with no way for the caller to know when they finish. At every weight sync the trainer therefore proceeds to the NCCL weight overwrite while backend work from the aborted requests may still be running under old weights — mixed-weight samples and late outputs. The acknowledged-abort design exists in [vllm-project/vllm-omni#6084](https://github.com/vllm-project/vllm-omni/pull/6084); without it, abort failures are unobservable and the trainer cannot fail closed.
> 
> ## 2. AR `pause_generation` is a frontend flag only
> At the pin, `AsyncOmni.pause_generation()` (vllm_omni/entrypoints/async_omni.py:827) only sets `self._paused = True` on the frontend; `wait_for_inflight_requests` is an unimplemented TODO and nothing is forwarded to EngineCore. An empty `request_states` therefore proves nothing about the backend — there is no engine-side idle boundary before the weight overwrite, even when the ahead-batch finished early. This is the mechanism behind the "actor-updating takes extensive time" symptom: the trainer-side drain (up to 120 s) is compensating for an idle boundary the engine never provides.
> 
> ## 3. StagePool abort discards partial tokens
> `StagePool.abort_requests()` (vllm_omni/engine/stage_pool.py:1130, pin fe478a95) routes request IDs to replicas and deletes state, but never builds a terminal output from the output processor — no `finish_reason="abort"` output with the tokens generated so far. The AR client therefore resumes from the original prompt instead of `prompt + partial_tokens`, silently discarding the partial rollout at every hard abort. On MMK12's long tail this fires at most syncs, so partial-rollout continuation never happens and trajectory stats diverge from `omni_sync` — the likely cause of the convergence gap. Note: [vllm-project/vllm-omni#6084](https://github.com/vllm-project/vllm-omni/pull/6084) does not cover this one; it needs cumulative abort outputs for AR stages.
> 
> Suggested: track [vllm-project/vllm-omni#6084](https://github.com/vllm-project/vllm-omni/pull/6084) for 1 and 2, and file a separate vLLM-Omni issue for 3. Until then, hard-abort path behaves like a full restart and the sync boundary has no engine-side guarantee.

yes I have a separate fix for `StagePool abort discards partial tokens` already will make a separate PR

knlnguyen1802:
@SamitHuang to unblock this we might need to land 
https://github.com/vllm-project/vllm-omni/pull/6084
and
https://github.com/vllm-project/vllm-omni/pull/6367

knlnguyen1802:
qwen3_async
Update verify

zhtmike:
> qwen3_async Update verify

it there any gpu idle time, it should expected nearly zero.

knlnguyen1802:
@AndyZhou952 Should we mark this as ready ?

zhtmike:
I will do the follow-up for this.

zhtmike:
**NEEDS CHANGES.** Core design matches RFC #320 §3.3–3.7 — trainer, `OmniDetachActorWorker`, `adapter_name` fix, and `parameter_sync_step` read-path all check out against verl pin `fefb080`. But the shipped MMK12 example cannot start, one added line is dead code, and the new doc contradicts both the pin the branch already sits on and the PR's own tests.

1. `examples/gspo_trainer/qwen3_omni/run_qwen3_omni_thinker_gspo_lora_mmk12_separate_async_v1.sh:32,51,91` — fix it: parent asserts `train_batch_size == parameter_sync_step * ppo_mini_batch_size` at init (verl `trainer_separate_async.py:48-53` @ pin), script sets 128 / 16 / 1 → AssertionError on line 1. Default 4 fails too (64 ≠ 128). Use `parameter_sync_step=8` (or `ppo_mini_batch_size=128`) and state the constraint like `run_sd35_medium_ocr_lora_v1_separate_async.sh:24` does.

2. `verl_omni/trainer/omni/ray_omni_trainer_separate_async.py:37` — drop `self.replay_buffer.parameter_sync_step = ...`: nothing at the pin reads it (verl `replay_buffer.py` never touches the knob; staleness is gated by `max_off_policy_threshold` alone). The test comment "the ReplayBuffer … normalizes staleness by this knob" is wrong — drop the two buffer assertions in `test_parameter_sync_step_follows_validated_key` too.

3. `docs/algo/separate_async_omni.md` "Abort semantics" bullet — rewrite against RFC #320 §2.1: it describes pre-#497 behavior (drain window, zero-token synthesis, "#290 hardening"), but this branch is already on pin `ded89346` — abort-then-pause with real partial-token terminals, and your own `test_omni_rollout_recovery_on_cpu.py` pins mid-sequence continuation (budgets `[8, 6]`).

4. Same doc, "Decoupled PPO is not active" bullet — wrong: generated omni config ships `bypass_mode: false` and nothing forces bypass, so the parent's save/restore path (verl `trainer_separate_async.py:157-178`) runs by default. `OmniDetachActorWorker`'s CPU save/restore is live on GPU — cover it in the gate.

5. `tests/special_e2e/run_gspo_qwen3_omni_thinker_lora_v1_separate_async_smoke.sh:17` — why does only this smoke `pip3 install transformers==5.12.1 accelerate==1.14.0 peft==0.19.1`? The sibling thinker v1 smoke it claims to follow runs the same image without pinning. If the image's transformers breaks vllm-omni, fix the image/CI job, not one script.

6. PR body is still template text + "WIP, investigating convergence + actor-updating"; CI on head ran only `drop_label`. RFC #320 §6.2 gate (hard-abort/weight-sync + convergence with Decoupled PPO engaged) — gate evidence and a reward curve in the body, pls~

Nice-to-have: `omni_sync` still runs verl's v1 actor worker, so LoRA sync (adapter-only send) now differs between the two omni modes — worth a tracking issue.

AI assistance (ZCode/GLM) was used for this review.

tests/special_e2e/run_gspo_qwen3_omni_thinker_lora_v1_separate_async_smoke.sh
```diff
@@ -0,0 +1,134 @@
...
+#
+# Requires: verl, verl-omni, vllm-omni installed; 2 GPUs.
+#
+# Override via env: NUM_TRAIN_GPUS, NUM_ROLLOUT_GPUS, MODEL_PATH, DATA_DIR, TOTAL_TRAIN_STEPS
+set -xeuo pipefail
+
+# The image is built with transformers 5.3.0; pip install ".[gpu,dev]" upgrades
+# to latest (5.14.x), which breaks vllm-omni (tie_word_embeddings removed) and
+# accelerate (unpatched _is_hf_initialized). Match our known-good local versions.
+pip3 install transformers==5.12.1 accelerate==1.14.0 peft==0.19.1
```
zhtmike: can drop it now.

knlnguyen1802: Dropped the in-script pip pin. The sibling thinker v1 smoke already runs on the same image without pinning; env versions belong in the CI job / image, not this script.

knlnguyen1802:
Addressed the 2026-09-08 review on ``11e98c0``:

1. MMK12 example: ``parameter_sync_step=8`` so ``128 == 8 * 16``, plus the batch-identity constraint in the header.
2. Dropped ``replay_buffer.parameter_sync_step`` and the two buffer assertions.
3. Abort-semantics doc rewritten against RFC #320 §2.1 / post-#497 abort-then-pause (partial-token continuation, budgets ``[8, 6]``).
4. Decoupled PPO doc corrected: ``bypass_mode: false`` by default, so ``OmniDetachActorWorker`` save/restore is live; ``parameter_sync_step=8`` engages it.
5. Dropped the smoke-only pip pin.
6. PR body filled in. The curve in the thread predates the identity fix, so the §6.2 convergence gate with Decoupled PPO engaged still needs a re-run on this recipe.

Also added ``Signed-off-by`` on the four earlier PR commits that were missing it.

knlnguyen1802:
Pushed `1defaf2` for the generate failure:

`RuntimeError: Generation rejected: Engine is partially or fully asleep. Currently sleeping tags: ['kv_cache']`

Cause: `AsyncOmni.generate()` rejects leftover sleeping tags (vllm-omni#4473). Two server bugs lined up with separate-async NCCL sync:

1. `_get_wake_up_tags()` returned `['weights']` only, so a full `wake_up()` left `kv_cache` asleep.
2. `release_kv_cache()` woke weights and then called `resume_generation()` *before* NCCL / `resume_kv_cache()`. `FullyAsyncLLMServerClient` reissued into that window (~722x across the cluster).

Fix: default wake tags are `['kv_cache', 'weights']` (verl parity). `release_kv_cache` keeps the admission hold; `resume_kv_cache` restores the cache and re-opens admission.

CPU tests updated: `test_release_kv_cache_holds_admission_until_resume`.

The 2026-09-08 review (comment 5581985508) is already on `11e98c0`: MMK12 `parameter_sync_step=8`, dead ReplayBuffer write dropped, abort/Decoupled PPO docs rewritten, smoke pip pin removed.

knlnguyen1802:
@zhtmike Ready again

zhtmike (commented):
Please clean something like #4473 in the code, comment, header and doc

knlnguyen1802:
@zhtmike cleaned in 7a2916e: dropped the raw `vllm-omni#4473` / `vllm-omni#6442` citations from the wake_up hook comment, the release-kv-cache test docstring, and the separate-async limitations doc. The failure modes are still described in words (full wake must include `kv_cache`; prefix-hash reuse after sleep/wake remains an upstream vllm-omni residual), so the rationale survives without the cross-repo numbers. The PR diff now carries no raw cross-repo issue citations — the remaining `verl#5995` comments are pre-existing on main, outside this change.

zhtmike (approved):
we need to do a profile after this

zhtmike:
9     FAIL     134s      FlowGRPO synchronous separate trainer e2e

knlnguyen1802:
**CI fix (ci-e2e-diffusion job [9] FlowGRPO separate e2e) — afdc08c**

The crash:
`AttributeError: 'ActorRolloutRefWorker' object has no attribute 'rollout_adapter'` from `update_weights` → `get_per_tensor_param(adapter_name=self.rollout_adapter)` on the **v0** separate path (`use_v1=false`, `checkpoint_engine.backend=nccl`).

Root cause: this branch swapped config reads (`self.config.rollout.rollout_adapter`) for instance knobs, but initialized them only inside `if "rollout" in self.role`. With `separate=true` the actor worker has no rollout role, never enters that branch, and then reads `self.rollout_adapter` (and `self.peft_merge` under LoRA) in the non-naive `update_weights` branch.

Fix: the knobs (`base_sync_done`, `layered_summon`, `rollout_adapter`, `peft_merge`, `_zmq_update_seq`) are now derived for **every** role via `_init_weight_sync_knobs()`, called unconditionally in `init_model` — semantics unchanged for colocated workers. Pinned by `tests/workers/test_weight_sync_knobs_on_cpu.py` (incl. omni-config defaults). Local env has no torch, so the pytest run itself is on CI.

verl_omni/workers/engine_workers.py
```diff
@@ -836,6 +834,22 @@ def init_model(self):
...
 
+    def _init_weight_sync_knobs(self, model_config):
+        """Set rollout weight-sync state needed by actor-side methods.
+
+        update_weights() and the LoRA gather read these on workers whose role
+        may not include "rollout" (separate rollout layout), so they are derived
+        from config for every role instead of only when building the rollout
+        engine.
+        """
+        # used for LoRA (base_sync_done is unused in merge-only mode but kept for Phase 2 adapter path)
```
zhtmike: unclear `Phase 2 `

knlnguyen1802: Dropped the `Phase 2` wording in 8f17755. The comment now states what the flag gates: `base_sync_done` stays True while the rollout already holds the base weights so adapter-only LoRA syncs skip resending them, and only a dummy `load_format` starts False (the rollout never loaded real base weights). Resolving.

knlnguyen1802:
ci-core fail, rerun ci-core only

docs/algo/separate_async_omni.md
```diff
@@ -0,0 +1,120 @@
...
+bash tests/special_e2e/run_gspo_qwen3_omni_thinker_lora_v1_separate_async_smoke.sh
+```
+
+For a parity check against the synchronous baseline, run the same recipe with
+`trainer.v1.trainer_mode=omni_sync` and compare reward curves; at
+`parameter_sync_step=1` they should overlap within noise.
+
+## Limitations
+
+- **Abort semantics (RFC #320 §2.1, post-#497).** Weight sync is
```
zhtmike: pls drop `#320, post-#497` these kind of things

PR #400, opened 2026-08-17
[1/N][data] feat: add FLUX first-frame data generation scripts for MiniMax H3 FL2VA

NancyFyong:
## What

Add the offline data pipeline used to build first-frame condition data for MiniMax H3 FL2VA (TI2V) RL training, colocated with the FL2VA recipe under `examples/diffusionnft_trainer/minimax_h3/`:

- `examples/diffusionnft_trainer/minimax_h3/gen_flux_images.py` — multi-GPU FLUX batch generator that turns a prompt list (e.g. DanceGRPO's `consist-id.txt`, filtered from ConsisID-preview-Data) into reference images. Shards prompts across ranks, uses deterministic per-index seeds, and resumes by skipping existing files. Defaults mirror DanceGRPO's online reference pipeline (400x640, 30 steps, guidance 3.5, max_sequence_length 512).
- `examples/diffusionnft_trainer/minimax_h3/README.md` — pipeline documentation (what the scripts do, usage, dataset recipe, published dataset link).
- `examples/diffusionnft_trainer/minimax_h3/build_fl2va_jsonl.py` — pairs each prompt with its same-index image, shuffles with a fixed seed, and writes `train.jsonl` / `test.jsonl` for the FL2VA `prepare_data.py` converter (first-frame conditioning).

The resulting dataset recipe: 27,815 prompts -> 27,815 FLUX.1-dev images -> seed-42 shuffle -> 27,687 train / 128 test -> `prepare_data.py --frame_mode first`. A ready-made copy of this dataset is published at https://huggingface.co/datasets/zyfenghit/dancegrpo-t2av.

## Why this is not duplicating an existing PR

- No open PR on verl-project/verl-omni covers FL2VA data tooling (`fl2va in:title,body` and `flux data in:title` searches return nothing).
- This is the data-generation counterpart split out of the MiniMax H3 FL2VA training work (#401) to keep the training PR focused.

## Tests

- `python3 -m py_compile examples/diffusionnft_trainer/minimax_h3/gen_flux_images.py examples/diffusionnft_trainer/minimax_h3/build_fl2va_jsonl.py` — pass.
- `python3 examples/diffusionnft_trainer/minimax_h3/build_fl2va_jsonl.py --prompt_file dancegrpo_consist-id.txt --image_dir data/flux_images/images --output_dir /tmp/fl2va_jsonl_test --test_size 128 --seed 42` — produced 27,687 train / 128 test rows, byte-identical to the JSONL used in the production training run.
[...]

PR #383, opened 2026-08-13
[diffusion, rollout, reward, cfg, tests] feat: add MiniMax H3 DiffusionNFT T2VA training

NancyFyong:
### What does this PR do?

Adds MiniMax H3 T2VA DiffusionNFT training on GPU with vLLM-Omni rollout, a Diffusers/FSDP actor, and LoRA old-policy synchronization. It also hardens the H3 LoRA transport path:

- maps Diffusers Q/K/V and SwiGLU FC1 LoRA deltas into vLLM H3 fused layers;
- validates the real TP=2 activation/sharding path with rank-64 LoRA;
- rejects unsafe `all-linear` targets, whose top-level actor LoRAs are not transported by layered sync;
- rejects unsupported conditional NFT tasks rather than silently replaying them as T2VA;
- preserves H3 prompt token IDs end to end through a raw-text AgentLoop and direct text-encoder input, avoiding decode/re-tokenize drift;
- unwraps joint audio-video rollouts: the `(video, audio)` tuple drives the video tensor path while the audio stream is forwarded to `extra_fields`, so CLAP/ImageBind rewards and the dumped mp4s receive it;
- registers `aspect_ratio` in the rollout pipeline config, normalizes rollout output shapes (channels-first video, 5D wandb batches, 6D rollout dumps), and migrates the H3 rollout adapter to the vLLM-Omni 0.27 `with_rollout_data()` contract;
- adds a prompt-only parquet converter (`prepare_t2av_data.py`) and a MiniMax H3 T2VA DiffusionNFT launch recipe that selects the dedicated AgentLoop explicitly.

Related RFC: #347.

This is not a duplicate of draft #368. #368 proposes H3 **FlowGRPO** with reverse-SDE trajectories and policy-gradient loss; this PR implements online **DiffusionNFT**, which trains from final clean latents with its forward-process direct-preference objective and `diffusion_nft_model` engine. The branches may share upstream H3 concepts, but their data contracts, loss, trainer routing, and rollout outputs are different.

### Checklist Before Starting

- [x] Search for similar PRs. Queries run:
  - https://github.com/verl-project/verl-omni/pulls?q=is%3Apr+is%3Aopen+347+in%3Abody
  - https://github.com/verl-project/verl-omni/pulls?q=is%3Apr+is%3Aopen+MiniMax+H3+DiffusionNFT
- [x] Format the PR title as `[{modules}] {type}: {description}`.

### Test

Commands run:

```bash
[...]

PR #401, opened 2026-08-17
[diffusion, rollout, cfg, tests, recipe] feat: add MiniMax H3 FL2VA DiffusionNFT

NancyFyong:
## What

MiniMax H3 FL2VA (first/last-frame conditioned text-image-to-audio-video) DiffusionNFT training, built on the MiniMax H3 NFT rollout adapter:

- `verl_omni/pipelines/minimax_h3_diffusion_nft/`: FL2VA rollout adapter with keyframe conditioning (`task=fl2va`, `frame_indices` [0] / [-1] / [0,-1]), per-slice fc1 sub-LoRA mapping to the fused vLLM DiT layout, and a verbatim prompt-tokenization agent loop.
- `examples/diffusionnft_trainer/minimax_h3/`: data converter (`prepare_data.py`, first/last/both frame modes) and the LoRA launch recipe.
- Recipe is aligned with the t2av DiffusionNFT setup (batch 32 / n=16, AdamW + diffusion-loss hyperparameters, FA3, gradient checkpointing), samples at 288x448 (32-pixel-aligned edges matching the condition-data aspect, ~50% fewer tokens than 400x640), validates at 576x928, and scores generations with the CLAP + ImageBind (audio_video) multi-reward via `MultiVisualRewardManager`.
- Shared rollout plumbing for joint audio-video output: the `(video, audio)` tuple is unwrapped in the async server with the audio stream forwarded to `extra_fields` (CLAP/ImageBind receive 32 kHz audio; dumped mp4s carry an AAC track), output shapes are normalized (channels-first video, 5D wandb batches, 6D rollout dumps), and the adapter follows the vLLM-Omni 0.27 `with_rollout_data()` contract.
- Operational hardening from live runs: `reward.num_workers=1` + `RAY_ACCEL_ENV_VAR_OVERRIDE_ON_ZERO` + `expandable_segments` to fix actor-update OOM (8 reward workers were each pinning ~5 GiB on cuda:0), `max_ckpt_to_keep=5`, pinned `WANDB_RUN_ID` for restart-safe logging, and persisted launcher logs.

Companion data PR: #400 (FLUX first-frame data generation scripts).

## Why this is not duplicating an existing PR

- `fl2va in:title,body` search on verl-project/verl-omni returns no open PRs.
- #383 (`minimax-h3-nft`) is the t2av (text-to-audio-video) NFT recipe; this PR adds the FL2VA (image-conditioned) task on a separate branch with its own rollout adapter, agent loop, and data path.

## Results
image

image

image

## Tests

- Focused CPU suites (NFT adapter, rollout contract, FL2VA contract, LoRA mapping, diffusion config): **94 passed**.
- End-to-end training run on 8 GPUs (FSDP2 actor + vLLM-Omni rollout, TP=4): validation at the resume checkpoint completed with `val-core reward mean 0.41` (CLAP 0.16 / ImageBind 0.25, audio at 32 kHz), then 3+ training steps with reward mean rising 0.41 -> 0.51, grad_norm 0.06-0.12, no OOM after the reward-worker fix.
- Val generations at 576x928 visually checked: first-frame identity holds, scene matches the prompt, temporal consistency across 107 frames (num_frames aligned to the 17n+5 boundary).
- `bash -n examples/diffusionnft_trainer/minimax_h3/run_minimax_h3_fl2va_lora.sh` — pass.

## AI assistance

This PR was prepared with AI assistance (CodeBuddy). The human submitter has reviewed the changes and run the tests above.


verl_omni/trainer/config/diffusion/rollout/diffusion_rollout.yaml
```diff
@@ -173,6 +176,10 @@ val_kwargs:
     # Flow-matching sigma-schedule shift for the video stream (validation)
     video_flow_shift: ${oc.select:actor_rollout_ref.rollout.pipeline.video_flow_shift,12.0}
 
+    # Optional request-contract fields for pipelines that accept a task label or frame indices (validation)
+    task: ${oc.select:actor_rollout_ref.rollout.pipeline.task,null}
```
zhtmike, line 180: I think `task` naming is ambiguous

verl_omni/workers/engine/fsdp/diffusers_impl.py
```diff
@@ -75,6 +75,26 @@
...
+        return
+    module.to(torch_dtype)
+
+
+def _fsdp_param_dtype(module: torch.nn.Module, configured_dtype: torch.dtype) -> Optional[torch.dtype]:
+    """Keep checkpoint parameter dtypes when an architecture declares fp32 islands."""
+    if getattr(module, "_keep_in_fp32_modules", None):
+        return None
+    return configured_dtype
+
```
zhtmike, line 96: what happens here?

NancyFyong: These two helpers just make the FSDP engine respect diffusers' own native `_keep_in_fp32_modules` mechanism, so the fp32 "islands" that `from_pretrained` already set up are not silently flattened back to bf16.

The chain in diffusers (0.40.0.dev0):

1. `MiniMaxH3Transformer3DModel` declares them — `models/transformers/transformer_minimax_h3.py`:
   `_keep_in_fp32_modules = ["proj_in", "audio_proj_in", "time_embedder", "proj_out", "audio_proj_out", "rope"]`
2. It is a native `ModelMixin` attribute (default `None`) — `models/modeling_utils.py`: `_keep_in_fp32_modules = None`.
3. `from_pretrained(torch_dtype=bf16)` honors it — `models/model_loading_utils.py` keeps params whose name matches the list in fp32 while casting the rest (`param = param.to(torch.float32)`).
4. The model's own `forward` relies on that mixed precision — its comments say *"the two patch projections are float32 while … the block stack [is] bfloat16 — see `_keep_in_fp32_modules`"* and *"The heads are listed in `_keep_in_fp32_modules`, so they stay float32 …"*.

Without these helpers, verl-omni's blanket `module.to(dtype)` at load plus FSDP `MixedPrecision(param_dtype=bf16)` would re-cast those islands to bf16. So:

- `_cast_loaded_diffusers_module` skips the blanket cast for a module that declares `_keep_in_fp32_modules` (keeps what `from_pretrained` loaded);
- `_fsdp_param_dtype` returns `None` for it so FSDP keeps the checkpoint dtypes instead of recasting every parameter.

It does not crash without the fix (H3's `forward` aligns activations to `weight.dtype`), it just silently loses precision on the rope/time/output projections. Written up in `docs/contributing/common_pitfalls.md` → "Float32 Islands Flattened by a Blanket bf16 Cast".

zhtmike: I see

verl_omni/trainer/config/diffusion/rollout/diffusion_rollout.yaml
NancyFyong: `task` here mirrors vLLM-Omni's own request-contract key: the H3 pipeline reads it as `extra.get("task")` and resolves it to `t2va` / `fl2va` / `ref2va` (`pipeline_minimax_h3.py::_resolve_task`). verl-omni's generic config→request forwarding passes the `DiffusionPipelineConfig` field name verbatim into the request `extra`, so keeping the field named `task` is what lets the engine pick it up without a special-case remap. Happy to rename it (e.g. `generation_task`) and add an explicit `→ "task"` mapping if you'd prefer a clearer name.

zhtmike: I see, may update the doc for this

NancyFyong: Ok, I update this

zhtmike (approved):

zhtmike:
pls trigger CI for this

PR #363, opened 2026-08-07
[vllm_omni, diffusion, rollout, tests, ci] feat: adapt rollout to vllm-omni 0.27

SamitHuang:
### What does this PR do?

  1. **Pin** `.github/vllm_omni_pin.txt` → `4444856…`
  2. **LTX-2.3** rollout adapter ported from removed `LTX23Pipeline` to
     `LTX2Pipeline` / `run_phase` + `LTXDenoiseExecutor`
  3. **AR + diffusion server** no longer require `OmniRequestOutput.request_output`
  4. **Envelope compat** wraps a mapping payload as `{payload, metadata}`
  5. **Smoke helper** maps pytest exit 5 to SKIP (empty collection, e.g. VeOmni)

### Checklist Before Starting

- [ ] Search for similar PRs. Paste at least one query link here: ...
- [ ] Format the PR title as `[{modules}] {type}: {description}` (This will be checked by the CI)
  - `{modules}` include `fsdp`, `vllm_omni`, `rollout`, `trainer`, `ci`, `training_utils`, `recipe`, `ray`, `worker`, `single_controller`, `misc`, `perf`, `model`, `algo`, `env`, `tool`, `ckpt`, `doc`, `data`, `cfg`, `reward`, `diffusion`, `omni`, `tests`, `docker`
  - If this PR involves multiple modules, separate them with `,` like `[diffusion, doc]`
  - `{type}` is in `feat`, `fix`, `refactor`, `chore`, `test`
  - If this PR breaks any API (CLI arguments, config, function signature, etc.), add `[BREAKING]` to the beginning of the title.
  - Example: `[BREAKING][diffusion, fsdp] feat: new rollout scheduler`

### Test

- [x] Local smoking test passed
- [x] E2E training:
  - [x] Diffusion

image

image

  - [x] Omni

image

image

image

### API and Usage Example

> Demonstrate how the API changes if any, and provide usage example(s) if possible.

```python
# Add code snippet or script demonstrating how to use this
```

### Design & Code Changes

> Demonstrate the high-level design if this PR is complex, and list the specific changes.

### Checklist Before Submitting

> [!IMPORTANT]
> Please check all the following items before requesting a review, otherwise the reviewer might deprioritize this PR for review.

- [ ] Read the [Contribute Guide](https://github.com/verl-project/verl-omni/blob/main/CONTRIBUTING.md).
[...]

PR #404, opened 2026-08-18
[diffusion, trainer] fix: align separate-async parameter_sync_step semantics

### LINKED ISSUE #374: [diffusion v1] parameter_sync_step > 1 has inconsistent batch and rollout-sync semantics
### System Info

verl-omni: 0.2.0rc1
verl: 0.9.0.dev0
Python: 3.12.13
PyTorch: 2.11.0+cu129
Ray: 2.56.1
Platform: Linux x86_64, 8 × NVIDIA H20

### Information

- [x] The official example scripts
- [ ] My own modified scripts

### Tasks

- [x] An officially supported task in the `examples` folder (such as QwenImage, ...)
- [ ] My own task or dataset (give details below)

### Reproduction

The issue is reproduced with the official Stable Diffusion 3.5 Medium + FlowGRPO Diffusion V1 separate-async example:

```text
examples/flowgrpo_trainer/sd35/run_sd35_medium_ocr_lora_v1_separate_async.sh
```

The script itself overrides parameter_sync_step to 1 and does not trigger the issue by default. The reproduction sets the script's exposed PARAMETER_SYNC_STEP override to 4, which is also the current Diffusion V1 YAML default.

The reproduction commit only adds stdout diagnostics: https://github.com/Wx727/verl-omni/commit/181eb29169e316c00913262a15c444d9d1664776

Relevant configuration:

```text
train_batch_size=8
ppo_mini_batch_size=8
rollout.n=8
actor DP=2
parameter_sync_step=4
sync_compatible=True
```

Diffusion V1 uses `parameter_sync_step` for both:

1. the number of local actor updates inside one global step;
2. the number of global steps between standalone rollout weight synchronizations.

### Root cause in code

The base trainer divides `train_batch_size` across the inner loop:

```python
sample_batch_size = train_batch_size // parameter_sync_step
for _ in range(parameter_sync_step):
    batch = self._step_once(..., sample_batch_size)
```

Separate-async currently enforces `train_batch_size == ppo_mini_batch_size`, while actor metadata still uses the full configured PPO mini-batch:

```python
actor_global_mini_batch_size = ppo_mini_batch_size * rollout.n
```

The same `parameter_sync_step` is also reused as a cross-global-step sync interval:

```python
if global_steps % parameter_sync_step == 0:
    standalone_checkpoint_manager.update_weights(global_steps)
```

Standalone generation is then resumed whenever `sync_compatible=True`, even if that condition was false and no weight sync occurred.

### 1. Undersized inner batches are padded with copied trajectories

Each inner iteration receives:

```text
sample_batch_size = 8 / 4 = 2 prompts
inner trajectories = 2 × rollout.n(8) = 16
actor global mini-batch = ppo_mini_batch_size(8) × rollout.n(8) = 64
```

Observed logs:

```text
PARAM_SYNC_REPRO batch global_step=1 inner_step=0 before=16 after=64 unique_after=2 pad_size=48 actor_global_mini_batch_size=64
PARAM_SYNC_REPRO batch global_step=1 inner_step=1 before=16 after=64 unique_after=2 pad_size=48 actor_global_mini_batch_size=64
PARAM_SYNC_REPRO batch global_step=1 inner_step=2 before=16 after=64 unique_after=2 pad_size=48 actor_global_mini_batch_size=64
PARAM_SYNC_REPRO batch global_step=1 inner_step=3 before=16 after=64 unique_after=2 pad_size=48 actor_global_mini_batch_size=64
```

Each inner input contains 16 trajectory rows from two prompt groups. `pad_dataproto_to_divisor()` adds 48 rows by copying existing rows, so the actor receives 64 rows without any additional trajectories being sampled.

### 2. Standalone rollout resumes without receiving updated weights

Observed logs:

```text
PARAM_SYNC_REPRO actor_update global_step=1 inner_step=0 updates_since_last_sync=1 batch_size=64
PARAM_SYNC_REPRO actor_update global_step=1 inner_step=1 updates_since_last_sync=2 batch_size=64
PARAM_SYNC_REPRO actor_update global_step=1 inner_step=2 updates_since_last_sync=3 batch_size=64
PARAM_SYNC_REPRO actor_update global_step=1 inner_step=3 updates_since_last_sync=4 batch_size=64
PARAM_SYNC_REPRO sync_decision global_step=1 parameter_sync_step=4 updates_since_last_sync=4 should_sync=False standalone_paused=True
PARAM_SYNC_REPRO resume_decision global_step=1 resuming_without_sync=True
```

Global step 1 performs four actor updates, skips standalone rollout weight synchronization, and then resumes standalone generation.

Synchronization is gated by:

```text
global_step % parameter_sync_step == 0
```

After the initial synchronization in `on_init_end()`, the next standalone weight synchronization occurs at global step 4, after up to:

```text
4 actor updates/global step × 4 global steps = 16 actor updates
```

The shipped separate-async recipe overrides `parameter_sync_step` to `1`, which avoids this behavior, but the Diffusion V1 YAML default is currently `4`.

### Expected behavior

`parameter_sync_step` should have one consistent meaning.

If Diffusion V1 follows upstream verl V1 semantics, `parameter_sync_step=N` should mean N local actor updates followed by one rollout weight synchronization, with:

```text
train_batch_size == parameter_sync_step × ppo_mini_batch_size
```

Inner batches should already match the actor mini-batch size and should not be silently padded with copied trajectories.

Standalone rollout generation should only resume with the intended actor weight version. If local update count and cross-global-step weight-sync interval are both needed, they should use separate configuration fields.

## Additional concern

When rollout-correction bypass is disabled, old log-probs are recomputed after prior inner actor updates. `π_old` may therefore drift within one synchronization cycle because Diffusion V1 does not include the upstream Decoupled-PPO snapshot/restore path.

This is based on code inspection and is not claimed as a reproduced failure in this report.

## Questions

1. Should Diffusion V1 follow upstream verl's `parameter_sync_step` semantics?
2. Is `parameter_sync_step > 1` intended to be supported today?
3. Would maintainers prefer an initial fail-closed `parameter_sync_step=1` restriction, or full alignment of batch sizing and weight-sync semantics?

I'd be happy to help with a fix and regression tests once the intended semantics are clarified. If this direction aligns with the project's plans, please let me know whether I can take it on.

Wx727:
### What does this PR do?

Fixes #374 by aligning Diffusion V1 `separate_async` with the expected `parameter_sync_step=N` semantics:

- Run `N` complete PPO mini-batch updates per outer step.
- Synchronize standalone rollout weights once after those `N` updates.
- Keep `π_old` fixed at the cycle-start actor weights.
- Prevent incomplete mini-batches from being padded with copied trajectories.
- Aggregate metrics across all local updates.
- Prefetch the complete sync cycle before pausing rollout in `sync_compatible` mode.

After the verl pin update, this PR reuses upstream implementations where possible:

- `ReplayBufferAsync` for stale/failed group eviction and exact refill.
- `MetricsAggregator` for cycle-level metric aggregation.
- `DetachActorWorker` strategy handlers for CPU parameter snapshots.

#### Targeted CPU regression tests

```bash
pytest -q \
  tests/trainer/diffusion/test_v1_parameter_sync_on_cpu.py \
  tests/trainer/diffusion/test_v1_replay_buffer_on_cpu.py
```

Result:

```text
28 passed
```

#### Qwen-Image GPU smoke test

```bash
PARAMETER_SYNC_STEP=2 \
TOTAL_TRAINING_STEPS=2 \
bash tests/special_e2e/run_flowgrpo_qwen_image_v1_separate_async.sh
```

Result:

```text
FlowGRPO diffusion v1 separate_async e2e test passed.
```

#### SD3.5 separate-async convergence

Configuration:

```text
trainer_mode=separate_async
parameter_sync_step=4
train_batch_size=8
ppo_mini_batch_size=2
sync_compatible=1
total_training_steps=100
```

Result:

```text
100/100 outer steps completed
Average runtime: 39.90s/step
Final val-core/flow_grpo/ocr/reward/mean@1: 0.7594
Exit code: 0
```

Training reward curve:

SD3.5 separate-async training reward

Validation reward curve:

SD3.5 separate-async validation reward

### API and Usage Example

No public configuration keys or function signatures are added.

For `separate_async`, the corrected invariant is:

```text
train_batch_size = parameter_sync_step * ppo_mini_batch_size
```

Example:

```bash
data.train_batch_size=8 \
actor_rollout_ref.actor.ppo_mini_batch_size=2 \
trainer.v1.separate_async.parameter_sync_step=4
```

Existing `parameter_sync_step=1` configurations remain unchanged.

### Design and Code Changes

- Treat one outer step as one rollout model-version cycle.
- Split each outer train batch into `parameter_sync_step` complete local updates.
- Keep cycle-start actor parameters as the fixed old policy for all local updates.
- Synchronize standalone rollout weights after every outer step.
- Use upstream `ReplayBufferAsync` for separate-async eviction and refill.
- Use generation batch size 1 so replay-buffer refill can request an exact number of prompts.
- Reject incomplete or non-divisible separate-async mini-batches instead of padding copied trajectories.
- Reuse upstream metric aggregation and detach-worker strategy handlers.
- Update the separate-async documentation, recipe, smoke test, and CPU regression tests.

- [x] Add or update documentation.
- [x] Add CPU regression tests.
- [x] Run targeted CPU tests.
- [x] Run GPU smoke validation.
- [x] Run SD3.5 convergence validation.


verl_omni/trainer/diffusion/v1/metrics.py
```diff
@@ -0,0 +1,127 @@
...
+    def get_aggregated_metrics(self) -> dict[str, Any]:
+        """Return one reduced value per metric for the parameter-sync cycle."""
+        if self.step_count == 0:
+            return {}
+
+        aggregated = {
+            name: self._aggregate_single_metric(name, values) for name, values in self.metric_values.items() if values
+        }
+        if {"global_seqlen/minmax_diff", "global_seqlen/max", "global_seqlen/min"}.issubset(aggregated):
+            aggregated["global_seqlen/minmax_diff"] = aggregated["global_seqlen/max"] - aggregated["global_seqlen/min"]
```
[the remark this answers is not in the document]
Wx727: This intentionally mirrors upstream verl.trainer.ppo.v1.utils.MetricsAggregator. The special case recomputes an existing derived metric across local steps rather than introducing a new metric. Diffusion currently does not emit these metrics, so no metric is lost.

examples/flowgrpo_trainer/sd35/run_sd35_medium_ocr_lora_v1_separate_async.sh
```diff
@@ -60,7 +60,8 @@ NUM_WARMUP_BATCHES=${NUM_WARMUP_BATCHES:-0}
 PARAMETER_SYNC_STEP=${PARAMETER_SYNC_STEP:-1}
 MAX_OFF_POLICY_THRESHOLD=${MAX_OFF_POLICY_THRESHOLD:-1}
 MAX_OFF_POLICY_STRATEGY=${MAX_OFF_POLICY_STRATEGY:-drop}
-TRAIN_BATCH_SIZE=${TRAIN_BATCH_SIZE:-8}            # must equal ppo_mini_batch_size
+TRAIN_BATCH_SIZE=${TRAIN_BATCH_SIZE:-8}
+PPO_MINI_BATCH_SIZE=$((TRAIN_BATCH_SIZE / PARAMETER_SYNC_STEP))
```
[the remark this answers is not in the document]
Wx727: The trainer validates train_batch_size == parameter_sync_step * ppo_mini_batch_size at startup and reports all three values on failure. We prefer to keep this invariant in one place rather than duplicate it in the example script.

knlnguyen1802 (commented):
Could you provide a convergence curve for the separate async on sd35 corresponding to new update ?

examples/flowgrpo_trainer/sd35/run_sd35_medium_ocr_lora_v1_separate_async.sh
knlnguyen1802, line 64: Need to check https://github.com/verl-project/verl-omni/pull/363

verl_omni/trainer/diffusion/v1/metrics.py
```diff
@@ -0,0 +1,127 @@
...
+"""Metric aggregation rules for diffusion V1 parameter-sync cycles."""
+
+from collections import defaultdict
+from typing import Any
+
+import numpy as np
+import torch
+
+
+class DiffusionMetricsAggregator:
```
knlnguyen1802: Can wait for the update pin of verl, after that can try to reuse `MetricAggregator` from `verl`

verl_omni/workers/detach_actor_worker.py
```diff
@@ -0,0 +1,131 @@
...
+        for param_name, param in model.named_parameters():
+            if param_name not in cpu_sharded_state:
+                continue
+            param.copy_(cpu_sharded_state[param_name].to(param.device))
+
+    if dist.is_initialized():
+        dist.barrier()
+
+
+class DiffusionDetachActorWorker(ActorRolloutRefWorker):
```
knlnguyen1802: Consider to inherit from `DetachActorWorker` of verl to better reuse some function

knlnguyen1802 (commented):
Please update the docs at  `docs/start/diffusion_v1.md` too

knlnguyen1802:
@Wx727 verl 0.9.0 pin is update please resolve comment

Wx727:
@knlnguyen1802 Thanks for the suggestions!

- Added the SD3.5 separate-async convergence curves and results to the PR description.
- Updated `docs/start/diffusion_v1.md`.
- Rebased onto the latest `main` with the updated `verl` pin.

examples/flowgrpo_trainer/sd35/run_sd35_medium_ocr_lora_v1_separate_async.sh
Wx727: I rebased onto the latest `main`, which includes #363, and the separate-async smoke test still passes.

verl_omni/trainer/diffusion/v1/metrics.py
Wx727: Updated. The local `DiffusionMetricsAggregator` has been removed, and the trainer now uses `verl.trainer.ppo.v1.utils.MetricsAggregator`.

verl_omni/workers/detach_actor_worker.py
Wx727: Updated. `DiffusionDetachActorWorker` now inherits from verl's `DetachActorWorker` and reuses its strategy handlers. The remaining overrides handle diffusion worker initialization and offloaded-parameter snapshots.

knlnguyen1802 (commented):
Looks aligned with upstream `parameter_sync_step=N` after the verl pin rebase: `ReplayBufferAsync` / `MetricsAggregator` / `DetachActorWorker` reuse is in place, `docs/start/diffusion_v1.md` + SD3.5 curves are attached, and the v1 separate_async GPU smoke passed (`PARAMETER_SYNC_STEP=2`).

knlnguyen1802 (approved):

Issue #302, opened 2026-07-24
[RFC] Multi-turn Agentic RL for Multi-modal Generation


HaFred:
## 1. Summary

This RFC proposes extending verl-omni from single-turn RL for diffusion models to **multi-turn agentic RL for omni-modal generation**. The current framework supports one-shot generation (prompt → image → reward → update). We propose adding:
1. Multi-turn trajectory support with agent reasoning, iterative prompt rewriting, and tool invocation;
2. A Decoupled Agent–Tool Policy (train agent LLM, freeze diffusion), where an agent LLM and/or the diffusion model can be trained across Modes (2a)
3. A multi-dimensional reward system with per-turn granularity and VisionCreator-R1's RPCO staged training. 

These extensions are motivated by recent research demonstrating that agentic approaches — iterative prompt rewriting, explicit reflection, multi-step planning — achieve substantial gains over single-turn generation (e.g., +23.6% on GenEval++, +14.22 average across 9 benchmarks without any training).

## 2. Agentic RL Taxonomy in `verl-omni`

Here we introduce the training taxonomy for Agentic RL in `verl-omni`, organized along two dimensions: **turn structure** (single-turn vs multi-turn) and **what is trained** (diffusion, agent LLM, or both):

| # | Name | Turn structure | Policy (trained) | Frozen | Description |
|---|------|---------------|-----------------|--------|-------------|
| **Mode (0)** | Single-stage RL | Single-turn | Diffusion (FlowGRPO) or AR / omni-modality (GSPO) | N/A | `verl-omni_v0.2.0` capability: single-stage RL for generation-model output quality, not modified by this RFC. |
| **Mode (1)** | Multi-stage RL | Single-turn | Both agent LLM and diffusion (alternating across stages) | None | Co-optimization of agent reasoning and generation quality across staged training. Requires coordinated cross-model gradient computation. Deferred to future work. |
| **Mode (2a)** | Agentic LLM RL | Multi-turn | Agent LLM (e.g., Qwen3-VL) | Diffusion model(s) | Train the agent's reasoning, prompt rewriting, reflection, and tool-invocation behavior. The diffusion model is a frozen tool called by the agent. **This is the primary new capability proposed by this RFC.** |
[...]

PR #411, opened 2026-08-19
[2/N][data, tests] feat: add UniCoT source adapters and agentic RL dataset builder

HaFred:
### What does this PR do?

* Implements PR 2 of [RFC #302](https://github.com/verl-project/verl-omni/issues/302): UniCoT source adapters and the agentic RL dataset builder. 
* Reuses the existing fail-closed UniCoT-Self-Reflection parser and split/provenance machinery (#313) and adds a fail-closed UniCoT-Breakdown adapter (plan/reflect classification from the subtask prefix structure, "No breakdown needed." sentinel), plus a full-dataset train/val parquet builder (`build_unicot_agentic_rl`) over UniCoT-Self-Reflection-6K + UniCoT-Breakdown-3K. UniCoT fields are reward ground truth only (never fewshot); dataset images are not materialized. Reflection image URIs are hashed without pixel IO (empty/`None` → `MISSING_IMAGE`; mismatched paths → `TRANSITION_HASH_MISMATCH`). Parquet `data_source` follows `task_type`; Hub corpus stays on `extra_info.unicot_source`. Plan `subtask_images` are schema-validated and discarded; text `reference_subtasks` are the GT.

### Checklist Before Starting

- [x] Search for similar PRs. Paste at least one query link here: https://github.com/verl-project/verl-omni/pulls?q=is%3Apr+unicot
  - #409
  - #329
  - #313
- [x] Format the PR title as `[data, tests] feat: add UniCoT source adapters and agentic RL dataset builder` (This will be checked by the CI)
  - `{modules}` include `fsdp`, `vllm_omni`, `rollout`, `trainer`, `ci`, `training_utils`, `recipe`, `ray`, `worker`, `single_controller`, `misc`, `perf`, `model`, `algo`, `env`, `tool`, `ckpt`, `doc`, `data`, `cfg`, `reward`, `diffusion`, `omni`, `tests`, `docker`
  - If this PR involves multiple modules, separate them with `,` like `[diffusion, doc]`
  - `{type}` is in `feat`, `fix`, `refactor`, `chore`, `test`
  - If this PR breaks any API (CLI arguments, config, function signature, etc.), add `[BREAKING]` to the beginning of the title.
  - Example: `[BREAKING][diffusion, fsdp] feat: new rollout scheduler`

### Test

CPU tests (pass locally; CI CPU job runs `*_on_cpu.py`). No GPU / L2 — this PR is an offline parquet builder.

```bash
[...]

PR #412, opened 2026-08-19
[3/N][reward, tests] feat: add RPCO multi-dimensional reward scorer for agentic GRPO

HaFred:
### What does this PR do?

* Implements PR 3 of [RFC #302](https://github.com/verl-project/verl-omni/issues/302): the RPCO multi-dimensional reward scorer for agentic GRPO. 
* Adds a self-contained scorer (`agentic_multidim_reward`) with the stage-3 active set {reflection, plan, format, tool, result} over the trajectory text produced by the rollout and data PRs: `R_reflect` = last-judge C/A checkpoint quality blended with coverage against the reference eval summary/judge findings, `R_plan` = per-subtask coverage of the agent's plan lines, `R_format` = structural check ratio (forced Reflection counts on reflect **and** plan), `R_tool` = successful PNG generate **and** a trusted judge (not the PR1 `reward_tool_call` presence metric), `R_result` = exact image count (plan) / last trusted YES and `<= expected` (reflect). Wire with `reward.reward_manager.name=naive` so verl passes `solution_str`; `VisualRewardManager` `solution_image` raises. `task_type` is required. Garbage `w_*` fail closed. The weighted total over the per-task active set forms the GRPO score; `reward_done` / `reward_tool_call` reproduce the PR 1 closed-loop **metrics** (not extra score dims). Invalid rollouts (no parsed `generate_image` or no successful PNG) score zero with `rollout_valid=0`. Parser duplication vs #409 `agentic_reward` is intentional for an independent merge.

### Checklist Before Starting

- [x] Search for similar PRs. Paste at least one query link here: https://github.com/verl-project/verl-omni/pulls?q=is%3Apr+reward+agentic
  - #329
- [x] Format the PR title as `[reward, tests] feat: add RPCO multi-dimensional reward scorer for agentic GRPO` (This will be checked by the CI)

### Test

CPU tests (pass locally; CI CPU job runs `*_on_cpu.py`). No GPU / L2 — text scorer only.

```bash
CUDA_VISIBLE_DEVICES="" python3 -m pytest -q --asyncio-mode=auto \
  tests/utils/reward_score/test_agentic_multidim_reward_on_cpu.py
```

→ **23 passed**, 20 warnings in 36.39s (10 Sep 2026, 6ea96d2). Covers every dimension, fail-closed `task_type` / `w_*` / `solution_image`, YES-gated reflect result, forced Reflection format (reflect and plan), tool = PNG + trusted judge.

### API and Usage Example

New additive API: `verl_omni.utils.reward_score.agentic_multidim_reward.compute_score(data_source, solution_str, ground_truth, extra_info) -> dict` returning `score` plus `reward_{reflect,plan,format,tool,result}`, `reward_done`, `reward_tool_call`, and the counter/metric schema. The module also exports `REWARD_COMPONENTS`, naming the metric keys the PR 1 metrics manager consumes when the two PRs are composed.
```python
# Wire the scorer via the custom reward function (text trajectory, not VisualRewardManager):
#   reward.reward_manager.name=naive
#   reward.custom_reward_function.path=pkg://verl_omni.utils.reward_score.agentic_multidim_reward
#   reward.custom_reward_function.name=compute_score
#   # Per-row ground_truth drives the reward: required task_type, expected_num_images,
#   # reference_subtasks (plan) / reference_steps (reflect), and optional w_* dimension weights.
```

### Design & Code Changes

The scorer is intentionally self-contained so it can be reviewed and merged independently of the rollout and data PRs: it parses the trajectory text itself (Hermes/XML tool calls, tool/judge observation markers, forced-Reflection markers) and never imports agent-loop or dataset modules.
- `verl_omni/utils/reward_score/agentic_multidim_reward.py` — 5-dim scorer; `R_tool` = PNG + trusted judge; reflect `R_result` needs terminal YES; `naive` manager; fail-closed `task_type` / `w_*`
- `tests/utils/reward_score/test_agentic_multidim_reward_on_cpu.py` — CPU tests for every dimension, the weighted active set, the #409 indicator mapping, and schema completeness

> [!IMPORTANT]
- [x] Add / Update the documentation. Not required.
- [x] Add unit or end-to-end test(s) to [the CI workflow](https://github.com/verl-project/verl-omni/tree/main/.github/workflows) to cover all the code. If not feasible, explain why: CPU tests under `tests/utils/reward_score/` (`test_agentic_multidim_reward_on_cpu.py`) — **23 passed** locally (36.39s). No L2 GPU; recipe stays `[4/N]`.


verl_omni/utils/reward_score/agentic_multidim_reward.py
```diff
@@ -0,0 +1,501 @@
...
+
+def compute_score(
+    data_source: str = "",
+    solution_str: str = "",
+    ground_truth: Any = None,
+    extra_info: dict[str, Any] | None = None,
+    **kwargs: Any,
+) -> dict[str, float | str | int | None]:
+    """Compute the RFC #302 stage-3 reward and its complete metric schema."""
+    del data_source, kwargs
```
SamitHuang: `del kwargs` drops `solution_image`. Default omni `VisualRewardManager` always calls `compute_score(..., solution_image=response_visual)` and never `solution_str` (`verl_omni/reward_loop/reward_manager/visual.py`). On that path `text` stays empty and every sample hard-zeros. Accept decoded `responses` (or document `reward.reward_manager.name=naive` for Mode 2a), otherwise a recipe that only sets `custom_reward_function.path` silently no-ops. Same hole as PR #409 `agentic_reward`.

verl_omni/utils/reward_score/agentic_multidim_reward.py
```diff
@@ -0,0 +1,501 @@
...
+    checks = [
+        raw_blocks > 0 and len(calls) == raw_blocks,
+        successful_generates >= 1,
+        bool(judges) and (not generates or max(judges) > max(generates)),
+        terminal_done,
+    ]
+    if task_type == "plan":
+        checks.extend((bool(_extract_plan_lines(text)), policy_reflection))
+    else:
+        checks.append(bool(_REFLECTION_RE.search(_assistant_prose(text))))
```
SamitHuang: PR #409 force-injects `Reflection: ... agentic_forced_reflection=1` (mask=0) and then lets the policy sample `Done.`. `_assistant_prose` strips that injected span, so this check is False on the default closed-loop curriculum while `_terminal_decision` still credits `forced_context` and `reward_done=1`. `R_format` therefore cannot reach 1.0 / `protocol_ok=1` for the exact trajectories GRPO will train on until force is off. Treat forced-then-Done as format-complete, or stop mixing a policy-only Reflection check into the score while force is the default.

verl_omni/utils/reward_score/agentic_multidim_reward.py
```diff
@@ -0,0 +1,501 @@
...
+    blocked: bool,
+    rewrite_after_yes: int,
+) -> float:
+    if blocked or not terminal_done or successful_generates < 1 or rewrite_after_yes > 0:
+        return 0.0
+    if task_type == "plan":
+        return 1.0 if successful_generates == expected else 0.0
+    judges = _successful_judges(text)
+    final_yes = bool(judges) and judges[-1][2] is True
+    return 1.0 if successful_generates <= expected or final_yes else 0.0
```
SamitHuang: `or final_yes` is not what makes this lenient: `successful_generates <= expected` already awards `R_result=1.0` after a single NO image whenever `expected_num_images>=1` and the policy wrote `Done.`. The tests only cover early-stop with YES (`_reflect_trajectory`) and over-generation without YES. Fail closed unless the last trusted judge is YES (or drop `R_result` to 0 on a terminal NO), otherwise stopping after one rejected image is a free result point.

verl_omni/utils/reward_score/agentic_multidim_reward.py
```diff
@@ -0,0 +1,501 @@
...
+    ground_truth: Any = None,
+    extra_info: dict[str, Any] | None = None,
+    **kwargs: Any,
+) -> dict[str, float | str | int | None]:
+    """Compute the RFC #302 stage-3 reward and its complete metric schema."""
+    del data_source, kwargs
+    text = solution_str or ""
+    gt = _as_dict(ground_truth)
+    metadata = dict(extra_info or {})
+    task_type = str(gt.get("task_type") or metadata.get("task_type") or "reflect")
```
SamitHuang: Missing `task_type` silently becomes `reflect`, which drops `plan` from `_active_weights` even when `reference_subtasks` is present. Combined with `del data_source` there is no second signal. PR #411 also ships no-breakdown rows as `data_source=unicot_breakdown` with `task_type=reflect`; if a row loses `ground_truth.task_type`, a plan mix scores as reflect. Fail closed (require `task_type`, or treat nonempty `reference_subtasks` as plan) instead of defaulting.

verl_omni/utils/reward_score/agentic_multidim_reward.py
```diff
@@ -0,0 +1,501 @@
...
+    if not prompts or successful_generates == 0:
+        return result
+
+    valid_terminal_context = judge_ok > 0 and not blocked and rewrites_after_yes == 0
+    closed = valid_terminal_context and terminal_done and (policy_reflection or forced_context)
+    rewards = {
+        "reflect": _reflection_reward(text, gt),
+        "plan": _plan_reward(text, gt),
+        "format": _format_reward(text, task_type=task_type, successful_generates=successful_generates),
+        "tool": float(bool(calls)),
```
SamitHuang: `R_tool = float(bool(calls))` is identically 1.0 on every `rollout_valid=1` row: validity already requires a parsed `generate_image` prompt. It only zeros on the early-return path that already zeros `score`. Default weight 1.0 therefore just dilutes reflect/format/result in the GRPO total. Map `R_tool` to a real protocol signal (successful generate+judge pair, or PR1 `f_tool_call` that can be 0 on valid text) or drop it from the active set.

SamitHuang (changes requested):
Request changes on the RPCO scorer.

The CPU tests (`*_on_cpu.py`) are named correctly and the parser is self-contained, but the scorer still cannot be composed with PRs #409/#411 as written: `compute_score` only reads `solution_str` while default `VisualRewardManager` passes `solution_image`; forced-Reflection + policy `Done.` (the #409 default) cannot saturate `R_format`; reflect `R_result` is 1.0 after a single NO as long as `Done.` is present; missing `task_type` fails open as reflect and drops plan weights; `R_tool` is a constant 1 on every valid rollout and only dilutes the total.

I did not re-run the 16 CPU tests from the PR body.

verl_omni/utils/reward_score/agentic_multidim_reward.py
HaFred: agreed. Default `VisualRewardManager` never passes `solution_str`, so empty text hard-zeroed. Now prefer `solution_str` (NaiveRewardManager), optionally decode `responses`, and raise on `solution_image`. Wire Mode (2a) with `reward.reward_manager.name=naive`.

verl_omni/utils/reward_score/agentic_multidim_reward.py
HaFred: agreed. `_assistant_prose` strips the injected Reflection, so the default #409 force-then-`Done.` curriculum never reached `protocol_ok=1`. `_format_reward` now treats `forced_context` as satisfying the Reflection check on reflect rows.

verl_omni/utils/reward_score/agentic_multidim_reward.py
HaFred: agreed. `successful_generates <= expected` was the free point, not `or final_yes`. Reflect `R_result` now requires a last trusted YES (and still zeros on rewrite-after-YES / blocked). Terminal NO is 0. CPU case added.

verl_omni/utils/reward_score/agentic_multidim_reward.py
HaFred: agreed. Silent `"reflect"` dropped plan weights and could mis-score #411 rows. `task_type` is required in `ground_truth` or `extra_info`; missing/invalid → hard zero, `method=agentic_multidim_missing_task_type`.

verl_omni/utils/reward_score/agentic_multidim_reward.py
HaFred: agreed. `float(bool(calls))` only diluted the total because validity already requires a parsed `generate_image`. `R_tool` is now `successful_generates >= 1 and trusted judge_ok >= 1`, so a PNG-only open loop can score tool 0.

PR #399, opened 2026-08-17
[omni, rollout, data, tests] feat: support Qwen3-Omni VL fully-async RL

hbhflw2000:
### What does this PR do?

This is the vision-language follow-up to #290. It adds the public recipe and validation coverage needed to run Qwen3-Omni Thinker image-conditioned Geo3K RL with a full-model Megatron actor and standalone vLLM-Omni fully asynchronous rollout replicas on NVIDIA GPUs.

The Draft currently includes:

- a Geometry3K converter that preserves image payloads and requests the response format consumed by verl's rule reward;
- a TP4 thinker-only vLLM-Omni stage configuration restricted to one image and text output;
- a 16-training-GPU + 16-rollout-GPU Megatron fully-async launcher;
- a regression test covering image forwarding and multimodal pad-token deduplication in the AR rollout path; and
- usage documentation.

This does not duplicate an open PR. #290 deliberately scoped itself to text-only GSM8K full-model fully-async rollout, while #231 covers image-to-text training through the existing LoRA/NPU-oriented recipe. This PR covers their missing composition: GPU Megatron full-model fully-async image-to-text RL. The current open-PR searches found no equivalent implementation: [Qwen3-Omni image fully async](https://github.com/verl-project/verl-omni/pulls?q=is%3Apr+is%3Aopen+Qwen3-Omni+image+fully+async), [Geo3K Qwen3-Omni](https://github.com/verl-project/verl-omni/pulls?q=is%3Apr+is%3Aopen+Geo3K+Qwen3-Omni).

Related merged PRs: #290, #231, #284.

### Test

Local checks on the Draft patch:

```text
bash -n examples/gspo_trainer/qwen3_omni/run_qwen3_omni_thinker_geo3k_megatron_fully_async.sh
# passed

python3 -m pytest -q tests/utils/test_geo3k_data_process_on_cpu.py
# 1 passed

ruff check examples/gspo_trainer/data_process/geo3k.py \
  tests/utils/test_geo3k_data_process_on_cpu.py \
  tests/workers/rollout/rollout_vllm/test_vllm_omni_generate.py
# All checks passed

ruff format --check examples/gspo_trainer/data_process/geo3k.py \
  tests/utils/test_geo3k_data_process_on_cpu.py \
  tests/workers/rollout/rollout_vllm/test_vllm_omni_generate.py
# 3 files already formatted

git diff --check
# passed
```

[...]

PR #284, opened 2026-07-20
[omni] feat: qwen3omni thinker supports multimodal inputs (image,text,audio->text) on npu

ZihaoW123:
### What does this PR do?

> Add **concise** overview of what this PR aims to achieve or accomplish. Reference related GitHub issues and PRs that help with the review.
qwen3omni thinker supports multimodal inputs (image,text,audio->text) on npu
training and validation on avqa datasets
image
image
 

### Checklist Before Starting

- [x] Search for similar PRs. Paste at least one query link here: ...
- [x] Format the PR title as `[{modules}] {type}: {description}` (This will be checked by the CI)
  - `{modules}` include `fsdp`, `vllm_omni`, `rollout`, `trainer`, `ci`, `training_utils`, `recipe`, `ray`, `worker`, `single_controller`, `misc`, `perf`, `model`, `algo`, `env`, `tool`, `ckpt`, `doc`, `data`, `cfg`, `reward`, `diffusion`, `omni`, `tests`, `docker`
  - If this PR involves multiple modules, separate them with `,` like `[diffusion, doc]`
  - `{type}` is in `feat`, `fix`, `refactor`, `chore`, `test`
  - If this PR breaks any API (CLI arguments, config, function signature, etc.), add `[BREAKING]` to the beginning of the title.
  - Example: `[BREAKING][diffusion, fsdp] feat: new rollout scheduler`

### Test

> For changes that can not be tested by CI (e.g., algorithm implementation, new model support), validate by experiment(s) and show results like training curve plots, evaluation results, etc.

### API and Usage Example

> Demonstrate how the API changes if any, and provide usage example(s) if possible.

```python
# Add code snippet or script demonstrating how to use this
```

### Design & Code Changes

> Demonstrate the high-level design if this PR is complex, and list the specific changes.

### Checklist Before Submitting

> [!IMPORTANT]
> Please check all the following items before requesting a review, otherwise the reviewer might deprioritize this PR for review.

- [x] Read the [Contribute Guide](https://github.com/verl-project/verl-omni/blob/main/CONTRIBUTING.md).
- [x] Apply [pre-commit checks](https://github.com/verl-project/verl-omni/blob/main/CONTRIBUTING.md#code-linting-and-formatting): `pre-commit install && pre-commit run --all-files --show-diff-on-failure --color=always`
- [x] Add / Update the documentation.
[...]

PR #413, opened 2026-08-19
[omni, rollout, data, tests] feat: support Qwen3-Omni audio fully-async RL

hbhflw2000:
### What does this PR do?

This is the audio-language follow-up to #290 and the audio sibling of #399. It adds Qwen3-Omni Thinker audio-conditioned AudioMCQ RL with a full-model Megatron actor and standalone vLLM-Omni V1 separate-async rollout replicas on NVIDIA GPUs.

The focused implementation includes:

- a model-scoped Megatron engine adapter that forwards processor-ready audio tensors and lets the Thinker construct multimodal M-RoPE, while retaining upstream optimizer, loss, pipeline scheduling, and weight synchronization;
- an AudioMCQ converter with asset validation and audio-level train/validation splitting, plus an exact-answer/choice-label reward;
- a public full-model launcher that reuses `verl_omni/trainer/config/omni_megatron_trainer.yaml` and adds only AudioMCQ/model/topology overrides;
- a four-GPU tiny-random structural smoke; and
- focused CPU coverage for data preparation, reward behavior, audio input binding, engine dispatch, and the launcher's resolved shared configuration.

This does not duplicate an existing PR. #290 added text-only Qwen3-Omni full-model async rollout, while #399 covers image-conditioned Geo3K RL; neither adds AudioMCQ preparation/reward or the Megatron Thinker audio-input binding implemented here. Open-PR searches also found no equivalent AudioMCQ implementation: [Qwen3-Omni audio fully async](https://github.com/verl-project/verl-omni/pulls?q=is%3Apr+is%3Aopen+Qwen3-Omni+audio+fully+async), [AudioMCQ Qwen3-Omni](https://github.com/verl-project/verl-omni/pulls?q=is%3Apr+is%3Aopen+AudioMCQ+Qwen3-Omni).

Related PRs: #290, #399, #231, and #284.

### Test

Current rebased head (`dce9047`):

```text
python -m pytest -q \
  tests/utils/test_audiomcq_on_cpu.py \
  tests/pipelines/test_audiomcq_toy_audio_on_cpu.py \
  tests/pipelines/test_qwen3_omni_megatron_inputs_on_cpu.py \
  tests/trainer/omni/test_audiomcq_megatron_config_on_cpu.py
# 34 passed

bash -n examples/gspo_trainer/qwen3_omni/run_qwen3_omni_megatron_audiomcq_separate_async.sh
bash -n tests/special_e2e/run_qwen3_omni_megatron_audiomcq_smoke.sh
python tests/special_sanity/check_device_api_usage.py --directory verl_omni
python -m py_compile tests/trainer/omni/test_audiomcq_megatron_config_on_cpu.py
git diff --check
# passed
```

The config regression invokes the public launcher in config-only mode, resolves the real shared Megatron YAML, and asserts the Megatron/EP/TP topology, frozen towers, AudioMCQ reward, rollout limits, V1 separate-async mode, 150 steps, and validation frequency.

Full-scale acceptance run (before the final shared-config deduplication, with the same framework code and recipe-specific topology):

- LTS app `k8s-vj-jfjrzh-1789388584582`, llab normal priority, succeeded on 4 nodes / 32 A100-80G GPUs.
- Qwen3-Omni-30B-A3B Thinker with full-model Megatron training and standalone async rollout workers; the audio and vision towers remained frozen.
- 150/150 policy updates completed in 5:08:08. Validation ran before training and every 10 steps, producing all 16 expected points at steps 0, 10, ..., 150.
- `val-core/Harland/AudioMCQ-StrongAC-GeminiCoT/reward/mean@1` reached 0.9453125 at step 150; validation format validity reached 1.0.
- Training reward rose from a first-10-update mean of 0.0070 to a last-10-update mean of 0.8766; the final value was 0.8828125.
- Actor grad norm remained bounded (maximum 1.6727, final 0.5268). Final PPO KL was 0.04521 and rollout/actor probability correlation was 0.99824.
- All 93 TensorBoard scalar tags were finite; the final response-aborted ratio was 0.0.

Qwen3-Omni AudioMCQ stable 32-GPU E2E metrics

The trailing multiprocessing `atexit` message occurred after 100% training, final validation, and TensorBoard flush. The wrapper returned `rc=0`, wrote its completion marker, and LTS reported `succeeded`.

### API and usage

Prepare AudioMCQ parquet files from a local JSONL and separately downloaded audio assets:

```bash
python examples/gspo_trainer/data_process/audiomcq.py \
  --input-jsonl /data/AudioMCQ/data.jsonl \
  --audio-root /data/AudioMCQ \
  --output-dir /data/audiomcq-prepared \
  --validation-size 256 --seed 42
```

After allocating four eight-GPU nodes and starting Ray, run once on the head node:

```bash
MODEL_PATH=/models/Qwen3-Omni-30B-A3B-Instruct \
TRAIN_FILE=/data/audiomcq-prepared/train.parquet \
VAL_FILE=/data/audiomcq-prepared/validation.parquet \
OUTPUT_DIR=/persistent/audiomcq \
bash examples/gspo_trainer/qwen3_omni/run_qwen3_omni_megatron_audiomcq_separate_async.sh \
  ray_kwargs.ray_init.address=auto
```

The launcher records the command, commit, resolved config, console log, and TensorBoard events in a unique run directory. Extra Hydra overrides are accepted through `"$@"`.

### Scope and dependencies

- The Megatron adapter is deliberately Qwen3-Omni Thinker-only and requires BSHD, PP1, and CP1.
- The public recipe has no internal data/model paths or cluster bootstrap code.
- Audio forwarding stays model-scoped; it does not alter the shared AR strategy or reject video inputs.
- The recipe is not reproducible on the current public pins yet. `verl-project/verl` must first replace `megatron-bridge==0.5.2` and its paired Megatron-Core pin with a tested upstream pair that provides Qwen3-Omni Thinker conversion, audio forward/export, Transformers 5 registration, and current audio-length compatibility; this repository must then bump `.github/verl_pin.txt` to that verl revision.
- A replacement for the closed verl-project/verl#7767 must land in verl with equal-length multimodal position-ID regression coverage, followed by the same `.github/verl_pin.txt` bump here.
- The 150-step run used documented development dependency overrides. It validates this integration path, but neither that run nor the tiny-random smoke proves clean-checkout compatibility with the public pins or exercises the full-model TransferQueue position-ID failure.

### Checklist

- [x] Rebased onto current `main` and removed the old cluster-local launchers, shims, skip flags, and debug probes.
- [x] Reused the shared Megatron YAML; no model-specific recipe YAML remains.
- [x] Added public documentation and focused CPU coverage.
- [x] Completed a stable 150-step, 4-node / 32-GPU acceptance run with validation every 10 steps.
- [x] Signed commits with DCO sign-off.
- [ ] Human follow-up owner to review every changed line and run the repository-required pre-commit suite before merge.

AI assistance: OpenAI Codex was used to help prepare the implementation, tests, documentation, validation summary, and plot.


knlnguyen1802:
@hbhflw2000 Any update on this, the fully async v1 of Qwen3-Omni thinker is merged

tests/special_e2e/run_gspo_qwen3_omni_megatron_full_32gpu_fully_async.sh
```diff
@@ -0,0 +1,1657 @@
+#!/usr/bin/env bash
+set -xeuo pipefail
+
+# 4-node/32-GPU full Qwen3-Omni bringup for Megatron trainer plus standalone
+# vLLM-Omni rollout through verl fully_async_policy resource split.
+# Resource default: each node splits 4 train GPUs + 4 rollout GPUs.
```
wtomin, line 6: Could you simplify the smoke test ? I don't think we have the resources to run them all.

verl_omni/pipelines/__init__.py
```diff
@@ -12,51 +12,65 @@
...
-__all__ += list(minimax_h3_diffusion_nft.__all__)
-__all__ += list(minimax_h3_flow_grpo.__all__)
-__all__ += list(sd3_dpo.__all__)
-__all__ += list(sd3_flow_grpo.__all__)
-__all__ += list(wan22_dance_grpo.__all__)
-__all__ += list(qwen_image_dpo.__all__)
-__all__ += list(qwen_image_dual_grpo.__all__)
-__all__ += list(qwen_image_edit_flow_grpo.__all__)
-__all__ += list(boogu_image_flow_grpo.__all__)
+_skip_pipeline_registration = os.environ.get("VERL_OMNI_SKIP_PIPELINES", "0").strip().lower() in {
```
wtomin, line 17: Pls remove these patches.

verl_omni/pipelines/qwen3_omni/__init__.py
```diff
@@ -13,10 +13,16 @@
...
-__all__ = [
-    "Qwen3OmniThinkerAdapter",
-    "Qwen3OmniRolloutAdapter",
-]
+__all__ = ["Qwen3OmniThinkerAdapter"]
+
+# The explicit stage-config launcher does not use the rollout topology adapter.
+# Keep that optional registration out of lightweight Ray actors because their
+# vLLM-Omni checkout can legitimately predate the generated-pipeline API.
+if os.environ.get("VERL_OMNI_SKIP_PIPELINES", "0").strip().lower() not in {"1", "true", "yes"}:
```
wtomin, line 25: Remove these patches.

wtomin:
@hbhflw2000 Could you please resolve the conflicts and remove the redundant patches?

verl_omni/workers/rollout/vllm_rollout/vllm_omni_ar_strategy.py
```diff
@@ -159,6 +159,46 @@ def prepare_engine_args(self, engine_args: dict[str, Any], args: Namespace) -> N
...
+    ) -> None:
+        """Fail closed for AR inputs outside the validated Thinker contract."""
+        hf_config = getattr(self.server.model_config, "hf_config", None)
+        is_qwen3_omni = getattr(hf_config, "model_type", None) == "qwen3_omni_moe"
+
+        unsupported = []
+        if image_data is not None and not is_qwen3_omni:
+            unsupported.append("image_data")
+        if video_data is not None:
+            unsupported.append("video_data")
```
wtomin, line 178: Rejecting `video_data` for any pipeline rollout is not acceptable, I think.

tests/special_e2e/runtime_compat/sitecustomize.py
```diff
@@ -0,0 +1,16 @@
...
+installed package or enable any extra dependency.
+"""
+
+try:
+    import nvidia_resiliency_ext as _nvrx
+except ImportError:
+    pass
+else:
+    if not hasattr(_nvrx, "__version__"):
+        _nvrx.__version__ = "0.6.0"
```
SamitHuang: This sitecustomize shim fabricates `nvidia_resiliency_ext.__version__ = "0.6.0"` so Megatron-Core will enable optional APIs that the installed package did not declare. That is a cluster-local compatibility hack, not a repository contract: it can silently take a code path the wheel never supported, and it is why the PR description still lists this as a pre-merge removal. Drop it from the tree; pin or patch the real dependency instead of lying about its version at import time.

tests/special_e2e/test_async_32gpu_port_contract.py
```diff
@@ -0,0 +1,360 @@
...
+
+import pytest
+
+pytestmark = pytest.mark.cpu
+
+SCRIPT = (
+    Path(__file__).resolve().parent
+    / "run_gspo_qwen3_omni_megatron_full_32gpu_fully_async.sh"
+)
+CONDA_ENV = Path("/nfs/ml-training-ssd/users/liuwei/verl_mega_async")
```
SamitHuang: This CPU test hardcodes `/nfs/ml-training-ssd/users/liuwei/verl_mega_async` and `pytest.skip`s when that path is absent, so repository CI never executes the port-contract checks this PR cites as passing. A unit test of the launcher contract has to run with a stubbed `CONDA_ENV` (or without sourcing conda at all). Leave the NFS default out of `tests/`.

SamitHuang:
Pls resolve the conflicts, thx

verl_omni/workers/rollout/vllm_rollout/vllm_omni_strategy_base.py
```diff
@@ -190,6 +206,12 @@ async def generate(
             ``TokenOutput`` for AR mode or ``DiffusionOutput`` for diffusion mode.
         """
         prompt_ids = normalize_token_ids(prompt_ids)
+        self.validate_multimodal_args(
+            image_data=image_data,
+            video_data=video_data,
+            audio_data=audio_data,
+            mm_processor_kwargs=mm_processor_kwargs,
+        )
         multi_modal_data = self._build_multi_modal_data(image_data, video_data, audio_data)
```
SamitHuang: `generate()` on current `main` already builds an `OmniRolloutRequest` and no longer calls `_build_multi_modal_data`. This patch is against the old kwargs path, so the audio `validate_multimodal_args` hook will not rebase cleanly and can drop request fields that the wire contract now preserves. Rebase onto `OmniRolloutRequest` and validate `request.multi_modal_data()` / `mm_processor_kwargs` there; do not keep a parallel pre-request generate() shape.

verl_omni/__init__.py
```diff
@@ -20,11 +20,24 @@
...
-import verl_omni.agent_loop  # noqa: E402, F401
-import verl_omni.models  # noqa: E402, F401
-import verl_omni.pipelines  # noqa: E402, F401
-import verl_omni.reward_loop  # noqa: E402, F401
-import verl_omni.trainer  # noqa: E402, F401
-import verl_omni.workers.engine  # noqa: E402, F401
+# Apply model patches and auto-register the components requested by the caller.
+# Lightweight or compatibility-sensitive launchers may opt out of unrelated
+# registrations before importing verl_omni as an external module.
+def _registration_enabled(skip_env: str) -> bool:
```
SamitHuang: These `VERL_OMNI_SKIP_*` gates change package import side effects. Adapter registration is an import side effect with no miss error, so a worker that sets `VERL_OMNI_SKIP_PIPELINES=1` and then forgets `VERL_USE_EXTERNAL_MODULES` silently has no Thinker adapter. This is a cluster-local import workaround, not a public API; drop it and keep `import verl_omni` registering the same set of modules as `main`.

SamitHuang (changes requested):
NEEDS CHANGES. Audio transport for Qwen3-Omni is the right follow-up to #290/#399, but this branch is still a cluster-local stack: fabricated nvidia_resiliency_ext version, NFS conda/data defaults, skip-registration env gates, and a generate() path that predates OmniRolloutRequest on current main. Rebase, drop the shims, keep video_data on the public AR contract (see existing thread), and make CPU tests run without /nfs/ml-training-ssd/users/liuwei. Do not land the 1.6k-line fully-async launcher or the extra smoke scripts until they are trimmed to a public recipe.

zhtmike:
apparently this PR is not ready for review, since it contains a lot of dump + debugging scripts.

Pls take a look of the PR content before your AI review @wtomin @SamitHuang .

SamitHuang:
> apparently this PR is not ready for review, since it contains a lot of dump + debugging scripts.
> 
> Pls take a look of the PR content before your AI review @wtomin @SamitHuang .

Thanks for reminding. It's better to perform self-review before setting the PR as ready @hbhflw2000

verl_omni/pipelines/__init__.py
```diff
@@ -12,51 +12,65 @@
...
+        minimax_h3_flow_grpo,
+        qwen3_omni,
+        qwen_image_diffusion_nft,
+        qwen_image_dpo,
+        qwen_image_dual_grpo,
+        qwen_image_edit_flow_grpo,
+        qwen_image_flow_grpo,
+        qwen_image_mix_grpo,
+        sd3_dpo,
+        sd3_flow_grpo,
```
knlnguyen1802: **Critical:** This rewrite is missing `qwen3_tts` and `flux_dance_grpo`, both imported on current `main`. Even after the `VERL_OMNI_SKIP_PIPELINES` gate is dropped, a rebase that keeps this list will unregister those adapters for every `import verl_omni`. Restore the live `__init__` imports; do not maintain a skip-registration fork of this file.

verl_omni/workers/rollout/vllm_rollout/vllm_omni_ar_strategy.py
```diff
@@ -159,6 +159,46 @@ def prepare_engine_args(self, engine_args: dict[str, Any], args: Namespace) -> N
...
+        *,
+        image_data: Optional[list[Any]],
+        video_data: Optional[list[Any]],
+        audio_data: Optional[list[Any]],
+        mm_processor_kwargs: Optional[dict[str, Any]],
+    ) -> None:
+        """Fail closed for AR inputs outside the validated Thinker contract."""
+        hf_config = getattr(self.server.model_config, "hf_config", None)
+        is_qwen3_omni = getattr(hf_config, "model_type", None) == "qwen3_omni_moe"
+
```
knlnguyen1802: **Critical:** Qwen3-Omni-only audio/image rules do not belong on the shared `ARStrategy`. Branching on `model_type == qwen3_omni_moe` and rejecting every other `mm_processor_kwargs` key (including `use_audio_in_video`) will break later AR models and the existing video+audio Thinker path. Keep `video_data` on the public AR contract. After rebase, validate `request.multi_modal_data()` / `mm_processor_kwargs` on the Thinker adapter, not here.

verl_omni/workers/rollout/vllm_rollout/utils.py
```diff
@@ -129,6 +128,11 @@ def update_weights_from_ipc(
         )
```
knlnguyen1802, line 130: **Critical:** If `peft_config` is set but `lora_config` is missing, this drains adapter weights and returns success. That is a silent no-op train: the actor publishes LoRA, rollout logs a valid update, rewards stall. Fail closed (raise) unless this is an explicit, tested full-weight recipe path.

verl_omni/workers/rollout/vllm_rollout/utils.py
```diff
@@ -244,6 +250,17 @@ def _get_zmq_handle(self) -> str:
         device within the replica list. Falls back to the stage-local rank
         when the lists are absent or the device is not in the replica list.
```
knlnguyen1802, line 245: **Critical:** `VERL_VLLM_WEIGHT_SYNC_ZMQ_HANDLES` plus `self.rank` is another cluster-local bypass of the node-local ZMQ contract (`job_id`, `replica_rank`, `local_rank`). Offset `CUDA_VISIBLE_DEVICES` and multi-stage deploys already have a remap; do not add a comma-separated handle list that CI cannot exercise.

tests/workers/rollout/test_vllm_omni_weight_update.py
```diff
@@ -0,0 +1,211 @@
...
+
+
+_install_lightweight_imports()
+
+from verl_omni.utils.vllm_omni import OmniTensorLoRARequest  # noqa: E402, I001
+from verl_omni.workers.rollout.vllm_rollout import utils as rollout_utils  # noqa: E402
+from vllm_omni.lora.request import LoRARequest as OmniLoRARequest  # noqa: E402
+
+pytestmark = pytest.mark.cpu
+
```
knlnguyen1802: **Critical:** L1 collects only `*_on_cpu.py`. This file has `pytest.mark.cpu` but the wrong name, so the LoRA-drain / ZMQ-handle tests never gate merge. Rename it (and drop the `sys.modules` inject if the real imports can be used).

tests/special_e2e/run_gspo_qwen3_omni_megatron_full_32gpu_fully_async.sh
```diff
@@ -0,0 +1,1657 @@
...
+REPO_ROOT="$(cd -- "${SCRIPT_DIR}/../.." && pwd)"
+ASYNC_ROOT="$(cd -- "${REPO_ROOT}/.." && pwd)"
+VERL_ROOT=${VERL_ROOT:-${ASYNC_ROOT}/verl}
+VLLM_OMNI_ROOT=${VLLM_OMNI_ROOT:-${ASYNC_ROOT}/vllm-omni}
+MEGATRON_BRIDGE_REPO=${MEGATRON_BRIDGE_REPO:-${ASYNC_ROOT}/megatron-bridge}
+RUNTIME_COMPAT_DIR="${REPO_ROOT}/tests/special_e2e/runtime_compat"
+cd "${VERL_ROOT}"
+
+CONDA_ENV=${CONDA_ENV:-/nfs/ml-training-ssd/users/liuwei/verl_mega_async}
+if [[ ! -f "${CONDA_ENV}/bin/activate" ]]; then
```
knlnguyen1802: **Critical:** Default `CONDA_ENV`, HF cache, model, GSM8K parquet, and output roots are `/nfs/.../liuwei` (and `/nfs/ofs-llab-hdd/...`). A 1657-line fully-async launcher with those defaults is a cluster bringup script, not a repository recipe. Do not land it. A public AudioMCQ recipe belongs under `examples/gspo_trainer/` with required env vars and no NFS fallbacks.

tests/special_e2e/run_gspo_qwen3_omni_megatron_full_32gpu_fully_async.sh
```diff
@@ -0,0 +1,1657 @@
...
+  echo "[error] CONDA_ENV not found: ${CONDA_ENV}" >&2
+  exit 1
+fi
+source "${CONDA_ENV}/bin/activate"
+
+export PYTHONNOUSERSITE=${PYTHONNOUSERSITE:-1}
+export VLLM_USE_V1=${VLLM_USE_V1:-0}
+export VLLM_DISABLE_COMPILE_CACHE=${VLLM_DISABLE_COMPILE_CACHE:-1}
```
knlnguyen1802: **Critical:** `VLLM_USE_V1=0` fights the repository vLLM 0.28 pin. V0 is not a supported compatibility shim on that stack; if the recipe cannot run on the pin, that is a pin/engine bug, not a launcher default.

tests/special_e2e/run_gspo_qwen3_omni_megatron_full_32gpu_fully_async.sh
```diff
@@ -0,0 +1,1657 @@
...
+export VERL_USE_EXTERNAL_MODULES=${VERL_USE_EXTERNAL_MODULES:-verl_omni,verl_omni.models.transformers.qwen3_omni_thinker,verl_omni.pipelines.qwen3_omni}
+# Rollout server actors need the model adapter registration side effects.  Keep
```
knlnguyen1802: **Critical:** `VERL_USE_EXTERNAL_MODULES` lists `verl_omni.models.transformers.qwen3_omni_thinker`. That module is not in this PR, and `verl_omni/models/` is empty on `main`. Combined with `VERL_OMNI_SKIP_PIPELINES=1` below, adapter registration becomes a cluster-local import path with no miss error. Keep `import verl_omni` registering the same modules as `main`.

tests/special_e2e/run_gspo_qwen3_omni_megatron_pruned_fully_async_4gpu_local_gate.sh
```diff
@@ -0,0 +1,45 @@
...
+
+export VERL_OMNI_FORCE_STANDALONE_ROLLOUT=${VERL_OMNI_FORCE_STANDALONE_ROLLOUT:-1}
+export VERL_OMNI_RESOURCE_SPLIT_IMPL=${VERL_OMNI_RESOURCE_SPLIT_IMPL:-standalone_rollout}
+export VERL_OMNI_TRAIN_GPUS_PER_NODE=${VERL_OMNI_TRAIN_GPUS_PER_NODE:-${TRAIN_GPUS}}
+export VERL_OMNI_ROLLOUT_GPUS_PER_NODE=${VERL_OMNI_ROLLOUT_GPUS_PER_NODE:-${ROLLOUT_GPUS}}
+export VERL_OMNI_ROLLOUT_NNODES=${VERL_OMNI_ROLLOUT_NNODES:-1}
+
+export VERL_OMNI_SKIP_WEIGHT_UPDATE=${VERL_OMNI_SKIP_WEIGHT_UPDATE:-1}
+export VERL_OMNI_SKIP_INITIAL_ROLLOUT_RESUME=${VERL_OMNI_SKIP_INITIAL_ROLLOUT_RESUME:-1}
+export VERL_OMNI_SKIP_INITIAL_ROLLOUT_SLEEP=${VERL_OMNI_SKIP_INITIAL_ROLLOUT_SLEEP:-1}
```
knlnguyen1802: **Critical:** Default `VERL_OMNI_SKIP_WEIGHT_UPDATE=1` (and the neighboring skip-resume / skip-sleep flags) lets a gate report success without publishing weights. This PR does not implement those env vars in `verl_omni/`; they only work against a private runtime. Drop them. A smoke that skips sync is not a weight-sync test.

tests/special_e2e/run_gspo_qwen3_omni_megatron_full_32gpu_audiomcq_e2e_gate.sh
```diff
@@ -0,0 +1,122 @@
...
+  data.shuffle=True \
+  data.validation_shuffle=False \
+  data.val_max_samples="${VAL_MAX_SAMPLES}" \
+  actor_rollout_ref.rollout.val_kwargs.n=1 \
+  actor_rollout_ref.rollout.val_kwargs.temperature=0 \
+  trainer.val_before_train=True \
+  trainer.test_freq="${TEST_FREQ}" \
+  async_training.use_trainer_do_validate=False \
+  +async_training.max_queue_size="${ASYNC_MAX_QUEUE_SIZE}" \
+  +async_training.max_concurrent_samples="${ASYNC_MAX_CONCURRENT_SAMPLES}" \
```
knlnguyen1802: **Critical:** `+async_training.max_queue_size` / `+async_training.max_concurrent_samples` are Hydra `+` inserts. If the fields already exist on the schema, compose fails; if they do not, they must be added to the dataclass instead of `+`. Same for `++actor_rollout_ref.actor.megatron.override_transformer_config.freeze_audio_model`.

tests/special_e2e/run_gspo_qwen3_omni_megatron_pruned_audiomcq_4gpu_gate_d2.sh
```diff
@@ -0,0 +1,36 @@
...
+  trainer.total_epochs=1 \
+  actor_rollout_ref.rollout.n=2 \
+  actor_rollout_ref.actor.ppo_mini_batch_size=2 \
+  rollout.n=2 \
+  +async_training.max_queue_size=2 \
+  +async_training.max_concurrent_samples=2 \
+  ++actor_rollout_ref.actor.megatron.override_transformer_config.freeze_audio_model=True \
+  "$@"
+
+exec /nfs/ml-training-ssd/users/liuwei/verl_mega_async/bin/python \
```
knlnguyen1802: **Critical:** This hardcodes `/nfs/ml-training-ssd/users/liuwei/verl_mega_async/bin/python` and `/tmp-data/liuwei/...`. `gate_d.sh` / `gate_d2.sh` are iteration dumps, not repository tests. Delete them with the other debug wrappers.

tests/special_e2e/qwen3_omni_hf_score_audio_rollout_corr_dump.py
```diff
@@ -0,0 +1,276 @@
+#!/usr/bin/env python3
+"""HF-score one audio-conditioned fixed response dumped by verl.
```
knlnguyen1802: **Critical:** HF-score dump of parquet rows vs Megatron actor/ref logprobs is a local debug probe, not CI. Same class as `qwen3_omni_vllm_omni_standalone_logprob_probe.py` and `analyze_qwen3_omni_audiomcq_rl_gate.py`. Keep those off `tests/`. If logprob Pearson is a merge gate, add a `*_on_cpu.py` contract test or attach the curve without landing the dumpers.

knlnguyen1802 (changes requested):
NEEDS CHANGES. Audio-conditioned Thinker rollout is the right follow-up to the text/image fully-async work, but this head is not mergeable: it is CONFLICTING against current main, the generate() path still predates OmniRolloutRequest, and most of the +4405 lines are cluster-local launchers, dump scripts, and skip-registration shims.

examples/gspo_trainer/qwen3_omni/config/audiomcq_megatron_separate_async.yaml
zhtmike, line 1: use the shared yaml config, do not write the model separated yaml. 

see `verl_omni/trainer/config/omni_megatron_trainer.yaml`

zhtmike (changes requested):
as comment

examples/gspo_trainer/qwen3_omni/config/audiomcq_megatron_separate_async.yaml
hbhflw2000: Done in ca5c46c. The model-specific YAML has been removed; the AudioMCQ launcher now selects the shared verl_omni/trainer/config/omni_megatron_trainer.yaml and supplies only recipe-specific CLI overrides. The config regression invokes that launcher and validates the resolved shared configuration.

hbhflw2000:
@zhtmike @wtomin @SamitHuang @knlnguyen1802 Review feedback is addressed on the current head ca5c46c. The branch is rebased and the old cluster-local launchers, NFS defaults, import/version shims, skip flags, rollout-strategy patches, ZMQ bypasses, and debug probes are no longer in the PR. The diff is now a focused 17-file AudioMCQ/Megatron change. Per the latest review, the model-specific YAML was deleted and the public launcher now reuses `verl_omni/trainer/config/omni_megatron_trainer.yaml`; its real resolved configuration is covered by the CPU regression. Current targeted result: 33 passed, plus shell syntax, py_compile, and diff checks. I also rewrote the PR description so its file list, tests, usage, scope, and 32-GPU evidence match the current branch. All review threads are resolved; please take another look when convenient.

zhtmike (approved):
Looks good

verl_omni/workers/engine/megatron/omni_impl.py
SamitHuang, line 1: `device="cuda"` will fail `tests/special_sanity/check_device_api_usage.py`: that job greps for `"cuda"` under `verl_omni/` and only skips files listed in `CUDA_KEYWORD_CHECK_WHITELIST`. `verl_omni/workers/engine/fsdp/omni_impl.py` is already on that list for the same registry declaration; this file is not. Add the whitelist entry (or drop the literal the way other engines do) before merge, otherwise pre-commit/CI is a hard fail.

verl_omni/pipelines/qwen3_omni/megatron_inputs.py
SamitHuang, line 1: Pinned `verl.models.mcore.model_forward.gptmodel_forward_model_engine` (the path `MegatronEngineWithLMHead.forward_step` actually calls) only copies `pixel_values` / `image_grid_thw` / video keys into `model_kwargs`. The pre-hook is the local workaround for audio tensors, but the CPU tests only mock `model(**kwargs)` and never run that mcore forward. Please add a test that the hook still fires when `model` is invoked the way `gptmodel_forward_model_engine` does (keyword `input_ids` plus vision kwargs, BSHD, no packed seq), or the adapter can silently drop audio if mcore ever switches to positional args.

verl_omni/utils/reward_score/audio_mcq.py
SamitHuang, line 1: The second `content_correct` assignment is unreachable as a True path: it only runs when `predicted != normalized_answer and predicted not in label_forms`, then sets `content_correct = predicted in normalized_choices and predicted == normalized_answer`. The last conjunct is already known false, so this never rewards a choice-text match the first line missed. Delete the dead branch (or rewrite it if the intent was substring/parroting rejection, which the first exact-match already covers).

examples/gspo_trainer/qwen3_omni/README_audiomcq.md
SamitHuang, line 1: The recipe is documented as unreproducible on the repository pins: it needs a forked Megatron-Bridge (`hbhflw2000/Megatron-Bridge@fe22f9d2` plus Transformers-5 registration/audio-length patches) and verl #7767, which was closed unmerged. That is honest, but then the 32-GPU acceptance numbers (taken before the shared-config dedup) cannot be the merge gate for `feat: support`. Please state in this README which pin bumps or follow-up PRs must land first, and do not treat the toy smoke as evidence the full-model TransferQueue position-id failure is gone.

SamitHuang (changes requested):
NEEDS CHANGES on the rebased 17-file head (`09c2f8a`). Cluster-local shims and the pre-OmniRolloutRequest generate() path are gone, and the AudioMCQ converter/reward/launcher shape is the right follow-up to #290/#399.

Blocking: `OmniMegatronEngine` registers with `device="cuda"` but is not on `CUDA_KEYWORD_CHECK_WHITELIST`, so `check_device_api_usage` will fail. Also the Megatron audio hook is only unit-tested against a mocked `model(**kwargs)` path, not `gptmodel_forward_model_engine`, which still copies only vision keys.

Non-blocking: dead second branch in `audio_mcq.compute_score`; recipe still depends on a forked Megatron-Bridge and closed verl #7767, so the 32-GPU numbers from before the shared-config dedup are not a pin-reproducible merge gate.

I re-ran none of the author's pytest locally (not run).

verl_omni/workers/engine/megatron/omni_impl.py
hbhflw2000: Fixed in dce9047. Removed the explicit `device="cuda"` argument and now rely on the EngineRegistry CUDA default, matching the upstream Megatron registrations. `python tests/special_sanity/check_device_api_usage.py --directory verl_omni` passes.

verl_omni/pipelines/qwen3_omni/megatron_inputs.py
hbhflw2000: Added in dce9047. The new CPU regression invokes the pinned verl `gptmodel_forward_model_engine` BSHD branch with keyword `input_ids` and multimodal kwargs, patching only the uninitialized model-parallel getters. It asserts that audio tensors and preserved vision tensors reach the model, Thinker position IDs are cleared, no `packed_seq_params` is passed, gradients flow, and the hook is removed.

verl_omni/utils/reward_score/audio_mcq.py
hbhflw2000: Fixed in dce9047. Removed the unreachable second assignment and retained exact answer-text or exact choice-label matching as the only scoring path. The AudioMCQ reward regression suite passes.

examples/gspo_trainer/qwen3_omni/README_audiomcq.md
hbhflw2000: Documented in dce9047. The README now lists the concrete merge prerequisites: a verl revision with a compatible Megatron-Bridge/Megatron-Core pair, a landed replacement for closed verl #7767 with equal-length multimodal position-ID coverage, and the corresponding `.github/verl_pin.txt` bump here. It also states explicitly that the development 150-step run and toy smoke do not prove public-pin reproducibility or exercise the full-model TransferQueue position-ID failure.

hbhflw2000:
@SamitHuang Addressed the four latest review comments in dce9047:

- Removed the hard-coded device="cuda" argument; the repository’s device-API sanity check now passes.
- Added a regression test that exercises the actual BSHD path in gptmodel_forward_model_engine from the pinned verl revision.
- Removed the unreachable branch in the AudioMCQ reward function.
- Clarified the required dependency and pin updates, as well as what the toy smoke test and the full-model run do—and do not—validate.

Validation: all 34 focused tests pass. Both launcher scripts pass bash -n, the modified Python files compile successfully, and git diff --check passes. I have also responded to each inline comment with the relevant details. Please take another look when you have time.

knlnguyen1802 (approved):

SamitHuang:
Pls fix the CI, thx

hbhflw2000:
CI follow-up:
I pushed e9185f3 to address the two failures attributable to this PR:
- applied the required Ruff formatting; 
- made the optional OmniMegatronEngine export conditional on Megatron availability, fixing API-doc checks in CPU-only environments while preserving the existing explicit-import behavior.
Validation:
- all pre-commit hooks passed
- 34 focused AudioMCQ/Megatron tests passed;
- 
The GPU-smoke failure is not specific to this PR:
- the failing cases are six existing diffusion tests and the existing Qwen3-Omni separate-async LoRA test;
- this PR does not modify the failing collect_lora_params or LoRA weight-sync implementations.
No shared LoRA/FSDP changes were reverted or modified in this update. Since the remaining failure is repository-wide and outside the scope of this AudioMCQ/Megatron PR, it should be handled by the corresponding shared CI/dependency fix.

verl_omni/workers/engine/megatron/omni_impl.py
```diff
@@ -0,0 +1,37 @@
...
+        model_config = copy(model_config)
+        model_config.hf_config = deepcopy(model_config.hf_config)
+        model_config.hf_config.text_config = model_config.hf_config.thinker_config.text_config
+        super().__init__(model_config, engine_config, optimizer_config, checkpoint_config)
+
+    def forward_step(self, batch_iter, model, logits_processor_func, postprocess_micro_batch_func):
+        batch = next(batch_iter)
+        multi_modal_inputs = extract_multi_modal_inputs(batch.get("multi_modal_inputs", []))
+        with qwen3_omni_megatron_inputs(model, multi_modal_inputs):
+            return super().forward_step(iter([batch]), model, logits_processor_func, postprocess_micro_batch_func)
```
SamitHuang: `MegatronEngineWithLMHead.forward_step` (verl `fefb080`) does `batch = next(batch_iter)` then `batch.to(get_device_id())` before `prepare_model_inputs`. Peeling the batch here and extracting MM inputs first means this path is not the device-moved batch the parent uses. Re-wrapping `iter([batch])` is necessary so the parent still sees this microbatch, but the hook then `.to(input_ids.device)` from a pre-move extract.

Override `prepare_model_inputs` (parent already calls `extract_multi_modal_inputs`) and attach the hook there, after the device move. Also fail closed if MTP is enabled: `gptmodel_forward_model_engine` only keeps `position_ids` for MTP, and the hook always forces them to `None`.

verl_omni/pipelines/qwen3_omni/megatron_inputs.py
```diff
@@ -0,0 +1,36 @@
...
+    """Forward audio tensors and let the Thinker construct multimodal M-RoPE.
+
+    The pinned verl forward only copies image/video inputs and otherwise supplies
+    ordinary text position IDs. A temporary module hook keeps the upstream loss,
+    padding and pipeline schedule intact; it never patches another model's code.
+    """
+
+    def prepare_inputs(_module, args, kwargs):
+        if kwargs.get("packed_seq_params") is not None:
+            raise ValueError("Qwen3-Omni Megatron requires BSHD (use_remove_padding=false).")
```
SamitHuang, line 23: This rejects packed sequences, then the next line always overwrites `position_ids` with `None`. That is required for Thinker M-RoPE on the BSHD path (`gptmodel_forward_model_engine` only keeps position IDs for MTP / non-vision), but `__init__` does not reject `mtp.enable`.

Fail closed on MTP here (or in `OmniMegatronEngine.__init__`). The public recipe also sets `actor.use_dynamic_bsz=true` while this adapter requires BSHD; pinned verl documents BSHD as a padded, preferably non-dynamic path. Either drop dynamic bsz in the launcher or justify why nested-tensor → `preprocess_bshd_engine` is safe for this audio forward.

SamitHuang (changes requested):
NEEDS CHANGES on `fe045a5` (17 files). Previous device-API / gptmodel_forward / dead-branch items look addressed. Remaining issues are on the Megatron audio hook vs the pinned `MegatronEngineWithLMHead` BSHD/MTP contract (inline).

Also still true and not a pin-reproducible merge gate: the 32-GPU numbers used a forked Megatron-Bridge plus closed verl #7767. The toy smoke does not exercise that TransferQueue path. `actor.use_dynamic_bsz=true` in the public launcher conflicts with the adapter’s BSHD requirement unless you can show nested-tensor → `preprocess_bshd_engine` is safe here.

Docs owners for `examples/`: @chenyingshu / @NancyFyong.

Not run this pass: GPU smoke, full pre-commit, Hydra composition.

verl_omni/workers/engine/megatron/omni_impl.py
```diff
@@ -31,7 +33,20 @@ def __init__(self, model_config, engine_config, optimizer_config, checkpoint_con
...
-        with qwen3_omni_megatron_inputs(model, multi_modal_inputs):
-            return super().forward_step(iter([batch]), model, logits_processor_func, postprocess_micro_batch_func)
+        with ExitStack() as input_adapters:
+            self._forward_model = model
+            self._input_adapters = input_adapters
+            try:
+                return super().forward_step(batch_iter, model, logits_processor_func, postprocess_micro_batch_func)
+            finally:
+                del self._input_adapters
+                del self._forward_model
```
zhtmike: do not del.

zhtmike (changes requested):

verl_omni/workers/engine/megatron/omni_impl.py
```diff
@@ -0,0 +1,54 @@
...
+
+    def forward_step(self, batch_iter, model, logits_processor_func, postprocess_micro_batch_func):
+        with ExitStack() as input_adapters:
+            self._forward_model = model
+            self._input_adapters = input_adapters
+            try:
+                return super().forward_step(batch_iter, model, logits_processor_func, postprocess_micro_batch_func)
+            finally:
+                self._input_adapters = None
+                self._forward_model = None
```
zhtmike, line 45: I don't get the idea what happens here. Pls explain it

hbhflw2000: These two attributes are only valid while the parent forward_step is active. Keeping their previous values after the forward would retain a stale model reference and an already-closed ExitStack. Resetting them to None keeps the attributes stable while providing an explicit “no active forward” state, which prepare_model_inputs also checks before installing the hook.
This is only lifecycle bookkeeping and does not change model inputs or training computation.

verl_omni/workers/engine/megatron/omni_impl.py
hbhflw2000: Addressed in 5d65d00. OmniMegatronEngine.forward_step no longer consumes or re-wraps the batch. The parent now retains ownership of iterator consumption and batch.to(get_device_id()).
The override calls the parent prepare_model_inputs first, then installs the audio hook using the already-prepared model_inputs["multi_modal_inputs"]. The outer ExitStack keeps that hook active until the parent performs the model forward and removes it on both success and failure.
The engine also now rejects mtp.enable explicitly. Regression coverage verifies the parent ordering, audio-gradient flow, unchanged batch ownership, and exception cleanup.

verl_omni/pipelines/qwen3_omni/megatron_inputs.py
hbhflw2000: Addressed in 5d65d00. Qwen3-Omni Megatron now fails closed when MTP is enabled.
The public BSHD recipe now disables dynamic batching for all three relevant paths:
- actor.use_dynamic_bsz=false
- ref.log_prob_use_dynamic_bsz=false
- rollout.log_prob_use_dynamic_bsz=false
Each path uses a fixed per-GPU micro-batch size of 1, and the resolved-launcher regression asserts these settings.
A 4-node/32-GPU full-model run using this configuration completed all 10 training steps successfully, including validation at steps 0 and 10, finite actor loss/gradients, and actor-to-rollout weight synchronization at every step.

verl_omni/workers/engine/megatron/omni_impl.py
zhtmike: why we need ExitStack, pls simply follow verl' forward loop for this.

hbhflw2000: Understood. I’ll simplify this implementation as follows:
- Remove the ExitStack, temporary engine attributes, and hook-lifetime indirection.
- Follow verl’s Megatron BSHD forward loop directly.
- Pass the Qwen3-Omni audio tensors at the actual model invocation.
- Keep verl’s existing batch preparation, logits processing, loss computation, and postprocessing behavior unchanged.
- Retain the fail-closed checks for unsupported MTP and non-BSHD configurations.
- Add focused forward/gradient parity tests and rerun the 32-GPU full-model E2E validation.
I’ll update this thread with the code and validation results once completed.

tests/trainer/omni/test_audiomcq_megatron_config_on_cpu.py
```diff
@@ -0,0 +1,264 @@
...
+        {
+            "AUDIO_MCQ_CONFIG_ONLY": "1",
+            "MODEL_PATH": "/tmp/model",
+            "OUTPUT_DIR": str(output_dir),
+            "PYTHON": sys.executable,
+            "TRAIN_FILE": "/tmp/train.parquet",
+            "VAL_FILE": "/tmp/validation.parquet",
+        }
+    )
+    subprocess.run(["bash", str(LAUNCHER)], cwd=REPO_ROOT, env=env, check=True, timeout=60)
```
NancyFyong, line 41: This fixed 60-second timeout is flaky on the current head. I reproduced `TimeoutExpired` here at 60 seconds (`1 error in 115.84s` including collection/startup), while the same config-only composition can finish when given more time. Because the launcher also starts `python -m verl_omni.trainer.main_omni` through a shell/`tee`, timing out only the outer shell can leave child processes behind. Please prefer composing the Hydra config directly for this L1 test; if the launcher must be executed, use a realistic timeout, start it in its own process group, and terminate the entire group on failure.

examples/gspo_trainer/qwen3_omni/README_audiomcq.md
```diff
@@ -0,0 +1,160 @@
+# AudioMCQ with Megatron and V1 separate-async rollout
```
NancyFyong, line 1: This is the repository's only `README_*.md` example page, so the standard `examples/**/README.md` docs-symlink checker does not discover or register it. Please either move it into a normal example subdirectory as `README.md` and add the corresponding `docs/examples` symlink/toctree entry, or merge it into the existing Qwen3-Omni example README. While doing so, please add the usual `Last updated` line, use the repository's `uv pip install` convention, and link the AudioMCQ dataset/license directly.

examples/gspo_trainer/data_process/audiomcq.py
```diff
@@ -0,0 +1,133 @@
...
+    validation_paths = set(paths[:validation_size])
+    train, validation = [], []
+    for row in rows:
+        split = "validation" if row["audios"][0] in validation_paths else "train"
+        row = {**row, "extra_info": {**row["extra_info"], "split": split}}
+        (validation if split == "validation" else train).append(row)
+    return train, validation
+
+
+def convert(input_jsonl: Path, audio_root: Path, output_dir: Path, validation_size: int, seed: int) -> dict:
```
NancyFyong, line 88: Please add a concise docstring to this public converter entry point covering its parameters, returned report, generated files, and fail-closed conditions (for example an existing output directory, invalid rows, or invalid/out-of-root audio paths). This is part of the public example API and should follow the repository's function-documentation convention.

NancyFyong (commented):
The PR description is stale relative to the current head `cd52b452`: it still names `dce9047` as the current rebased head and reports `34 passed`. The focused set now collects 35 tests, and the public-recipe config fixture currently times out at its hard-coded 60-second limit. Please update the description after the fixes with the exact final-head commands/results and clearly separate current-head validation from earlier full-model runs.

hbhflw2000:
Thanks for the review. I pushed the follow-ups through `f843978`:

- Replaced the `ExitStack`/forward pre-hook with a model-scoped direct BSHD forward that follows pinned verl's device-move, input-preparation, logits-processing, and postprocessing order. Unsupported configurations are rejected explicitly.
- Added BSHD logit/gradient parity and audio forward/backward regression tests. The launcher configuration test now has a 180-second timeout and cleans up its process group on timeout.
- Moved the AudioMCQ example into the standard README/docs layout, added the requested dataset/license and installation details, and documented `convert()`.
- On the current head, all 38 focused tests and the relevant static checks pass. A normal-priority 4-node/32-A100 run completed 30/30 updates; val-core at steps 0/10/20/30 was `0 / 0 / 0.3203125 / 0.89453125`. All recorded TensorBoard scalar values were finite. The LTS job succeeded; the trailing multiprocessing warning appeared after final validation.
- The GPU run used compute commit `081b9b2`; subsequent commits changed only docs/tests. The validated environment used a verl snapshot based on `fefb080` with the 3D position-ID layout repair from [verl #7901](https://github.com/verl-project/verl/pull/7901) applied, along with the documented Megatron compatibility dependencies. The 32-GPU recipe is runnable with those dependencies; reproducibility with the repository's current public pins remains unverified. This PR adds no #7901-specific API calls.

Please take another look when convenient.

tests/trainer/omni/test_audiomcq_megatron_config_on_cpu.py
hbhflw2000: Addressed in `3944e97`. The test retains the public launcher to verify its actual CLI overrides and resolved shared configuration. It now uses a 180-second timeout and a separate process group. On timeout, the entire group receives SIGTERM, followed by SIGKILL if it does not exit within 10 seconds. The configuration tests passed on `f843978`.

examples/gspo_trainer/qwen3_omni/README.md
```diff
@@ -0,0 +1,160 @@
+# AudioMCQ with Megatron and V1 separate-async rollout
```
hbhflw2000: Addressed in `3944e97`. Moved the page to `examples/gspo_trainer/qwen3_omni/README.md` and added the corresponding docs symlink and toctree entry. Also added the “Last updated” line, switched to `uv pip install`, and linked the dataset and its license information.

examples/gspo_trainer/data_process/audiomcq.py
hbhflw2000: Added in `3944e97`. The docstring now covers the parameters, returned report, generated files, and error conditions. It also clarifies that invalid or out-of-root audio records are excluded and counted in the report, while an existing output directory or insufficient valid assets causes conversion to fail.

knlnguyen1802 (changes requested):
We can consider to have a general megatron implement for omni model

hbhflw2000:
> We can consider to have a general megatron implement for omni model

Agreed. I suggest discussing the general Megatron backend in a separate RFC, covering Qwen3-Omni, Qwen2.5-Omni,  MiniCPM-o and so on. But  I suggest keeping this PR focused on Qwen3-Omni AudioMCQ. Would that work for you?

knlnguyen1802:
> > We can consider to have a general megatron implement for omni model
> 
> Agreed. I suggest discussing the general Megatron backend in a separate RFC, covering Qwen3-Omni, Qwen2.5-Omni, MiniCPM-o and so on. But I suggest keeping this PR focused on Qwen3-Omni AudioMCQ. Would that work for you?

Ok for me

verl_omni/workers/engine/megatron/omni_impl.py
```diff
@@ -0,0 +1,102 @@
...
+            distillation_use_topk=distillation_use_topk,
+            distillation_only=distillation_only,
+            logits_processor_func=logits_processor_func,
+            batch=batch,
+            data_format="bshd",
+        )
+        response_attention_mask = None
+        if attention_mask is not None and not loss_mask.is_nested:
+            response_attention_mask = attention_mask[:, -loss_mask.shape[-1] :]
+        output = qwen3_omni_forward_model_engine(
```
zhtmike: no model specific things in public engine implementation. Put it in pipelines/

hbhflw2000: Addressed in `d0b27e2`.

- The shared engine now selects config preparation and forward through the existing pipeline adapter registry; it no longer imports or checks Qwen3-specific code/config.
- Qwen3 support checks, nested-config mapping and forward selection live under `pipelines/qwen3_omni/`. Unsupported adapters fail before Megatron initialization, and Megatron imports remain lazy in the Qwen3 adapter.
- 52 focused CPU tests passed, including BSHD logit/gradient parity, audio forward/backward, pipeline dispatch and unsupported-adapter rejection. Ruff, device-API/DataProto checks, Python compilation and `git diff --check` also pass.

The model-forward body and config-preparation logic are unchanged; this is a code-ownership/dispatch refactor, with no new GPU E2E run. Validation used the same local compatibility dependencies documented in the recipe. Broader cross-model Megatron support remains in the separate RFC.

examples/gspo_trainer/qwen3_omni/README.md
zhtmike, line 1: You need to mention all contents in this directory. e.g., refer to ../README for non-megaton trainer

Or change the file name. To something related to Megatron

zhtmike (commented):
And merge main as well, so we can launch CI.

zhtmike:
Let us update the DOC for contributing Megatron model as well.

hbhflw2000:
Thanks, @zhtmike . Updated in da50a44:
- Added directory-wide recipe navigation and a link to the non-Megatron guide.
- Documented the Megatron adapter hooks, forward contract, dependencies and validation limits, and updated the API docs.
- Merged main in bc0b2ab.
Documentation checks pass.

zhtmike (approved):
Looks good to me, waiting other's feedback.

knlnguyen1802 (approved):
need @SamitHuang  review again

SamitHuang (approved):
LGTM

SamitHuang:
Final comment:
Doc discovery is still incomplete for the Megatron AudioMCQ path.

`examples/gspo_trainer/README.md` already points at `qwen3_omni/README.md`, and that covers RTD `docs/examples/gspo_trainer.md` (symlink). Two pages users actually open still describe Qwen3-Omni as FSDP-only:

1. `docs/start/models.md` — Qwen3-Omni-30B-A3B Thinker table / Supported trainers
2. `docs/algo/separate_async_omni.md` — GPU layout still says the actor is FSDP

Please add the two snippets below (no new Megatron chapter). Keep FSDP LoRA as the default; mark Megatron as experimental and not reproducible from the current public pins.

Suggested `docs/start/models.md` (Supported trainers table):

| GSPO (AudioMCQ, Megatron, separate-async) | `examples/gspo_trainer/qwen3_omni/run_qwen3_omni_megatron_audiomcq_separate_async.sh` | 32×A100 80GB (4 train + 4 rollout GPUs/node). Experimental: not reproducible from the current public pins; see the [AudioMCQ recipe](../../examples/gspo_trainer/qwen3_omni/README.md). |

And a one-line note under the FSDP row, e.g. Megatron full-parameter is a separate-async variant, not a replacement for FSDP2 LoRA.

Suggested `docs/algo/separate_async_omni.md` (GPU layout / Run):

The default recipe below is FSDP2 LoRA. Full-parameter Megatron uses the same `trainer.v1.trainer_mode=omni_separate_async` path:

```bash
bash examples/gspo_trainer/qwen3_omni/run_qwen3_omni_megatron_audiomcq_separate_async.sh

PR #305, opened 2026-07-25
[omni, data, tests] feat: qwen3omni thinker supports video->text on npu (extends #284)

zyxzyxzyx:
## Overview

This PR extends PR #284 (image, text, audio -> text) with **video -> text** GSPO full-parameter training for the Qwen3-Omni Thinker on NPU. It reuses PR #284's `choice_reward` scorer and the default `RLHFDataset` (video is auto-extracted via `qwen_vl_utils` through the `has_visual` path); only a video-specific RoPE patch, a data converter, and an NPU launch recipe are added.

## What this PR contributes (vs. PR #284)

1. **Video RoPE unified patch** (`verl_omni/models/transformers/qwen3_omni_thinker.py`): extends `_patch_agent_loop_audio_rope_for_qwen3_omni` into a unified wrapper that binds **both** `audio_seqlens` (from `feature_attention_mask`) and `second_per_grids` (from `video_second_per_grid`) to `get_rope_index`.
   - Root cause: verl's `AgentLoopWorker._compute_position_ids` forwards only `image_grid_thw` / `video_grid_thw`; Qwen3-Omni's `get_rope_index` indexes `second_per_grids[video_idx]` on its video branch, which crashes with `NoneType not subscriptable` when `second_per_grids` is `None`.
   - Field-name mapping `video_second_per_grid` -> `second_per_grids`, plus a list->tensor guard (`.cpu().float()` inside `get_rope_index` requires a tensor).
   - The unified wrapper handles audio-only / video-only / audio-in-video / text-only samples correctly.
2. **CPU tests** (`tests/pipelines/test_qwen3_omni_thinker_adapter_on_cpu.py`): 2 new tests (video `second_per_grids` forwarding; audio+video rope together). The existing audio test still passes (3/3).
3. **NExT-QA data converter** (`examples/gspo_trainer/data_process/video2text.py`): MC split -> verl RL parquet, answer index 0-4 -> letter A-E, video item dict `{fps, max_frames}` to cap tokens, group-split by `video_id` to avoid leakage.
4. **NPU launch script + README recipe** (`run_qwen3_omni_thinker_gspo_video_npu.sh`): video-specific settings (`max_prompt_length=12000`, `mm_processor_kwargs.fps=2.0`, `filter_overlong_prompts=false`, 8-worker frame extraction).

## Acknowledgement

[...]

PR #416, opened 2026-08-20
[omni, data, fsdp, recipe, doc] feat: add NExT-QA full-parameter GSPO training on NPU

SelfDLer:
### What does this PR do?

Add full-parameter GSPO training for Qwen3-Omni Thinker on the official NExT-QA dataset using Ascend NPUs.

Each sample contains a video, its soundtrack, and a five-choice question. The dataset loader supplies sampled video frames and separately decoded 16 kHz audio to both the actor and rollout. The model is trained to produce reasoning followed by an answer tag such as `<answer>C</answer>`.

The recipe uses the existing `choice_reward.py` scorer, which assigns a binary reward by matching the answer tag against the dataset label.

### Checklist Before Starting

- [x] Search for similar PRs. Paste at least one query link here: https://github.com/verl-project/verl-omni/pulls?q=is%3Apr+is%3Aopen+NExT-QA

#### Training results

The recorded run contains 88 training steps and validation results through step 80.

Actor-rollout Pearson correlation stays above `0.99` at every recorded step:

| Metric | Value |
| --- | ---: |
| Minimum Pearson correlation | 0.997983 |
| Maximum Pearson correlation | 0.998877 |
| Mean Pearson correlation | 0.998525 |

Validation reward:

| Step | Reward |
| ---: | ---: |
| 0 | 0.7624 |
| 10 | 0.7610 |
| 20 | 0.7758 |
| 30 | 0.7816 |
| 40 | 0.7832 |
| 50 | 0.7876 |
| 60 | 0.7840 |
| 70 | 0.7894 |
| 80 | 0.7852 |

The mean Pearson value is approximate; metrics were extracted from rounded training summaries.

training_metrics

#### Local checks

Passed local checks include:

- Training launcher Bash syntax.
- Syntax checks for the five changed Python files.
- Ruff lint and formatting, mypy, and changed-file license checks.
- Docstring, device API, DataProto, and test-structure checks.
- `git diff --check`.

The full pre-commit run did not pass in the local Windows environment: several hooks encountered shell execution failures, and repository-wide documentation/license checks reported failures in unchanged upstream files. CPU regression tests were not rerun for this revision.

### API and Usage Example

Convert the dataset:

```bash
python examples/gspo_trainer/data_process/nextqa.py \
    --input_dir /datasets/NextQA \
    --output_dir /datasets/NextQA
```

Launch training:

```bash
MODEL_PATH=/models/Qwen3-Omni-30B-A3B-Instruct \
TRAIN_FILE=/datasets/NextQA/train.parquet \
VAL_FILE=/datasets/NextQA/validation.parquet \
OUTPUT_DIR=/checkpoints/qwen3_omni_nextqa \
bash examples/gspo_trainer/qwen3_omni/run_qwen3_omni_thinker_gspo_npu_nextqa_v1.sh
```

The default configuration uses one 16-NPU node, rollout TP=4, 32 questions per batch, and eight responses per question. All Thinker parameters are trainable. The learning rate is `1e-6`, with validation every 10 steps and checkpointing every 20 steps.

FFmpeg and FFprobe must be available on the conversion host and all training workers. Dataset layout and video decoder setup are documented in `examples/gspo_trainer/README.md`.

### Design & Code Changes

The implementation includes:

- A converter for the official train and validation splits, with media validation and timeouts for FFprobe/FFmpeg checks.
- A dedicated `NextQARLHFDataset` for video and soundtrack loading.
- Sampled-video timing support in the processor and actor RoPE inputs.
- A temporary NPU FSDP2 compatibility override for the pinned verl version, keeping gradients sharded during accumulation.
- A full-parameter training launcher and documentation for dataset preparation, dependencies, and training.

#### NPU FSDP2 compatibility with the pinned verl version

verl-omni currently pins verl to `fefb080262e1c015a0ea05f958822a6a512dc795`. In that revision, the FSDP engine defers gradient synchronization on non-final micro-batches without a configuration option to disable it. This retains unsharded gradients during accumulation and increases peak memory for full-parameter Thinker training.

Upstream [verl#7458](https://github.com/verl-project/verl/pull/7458) has already added `use_no_sync_for_gradient_accumulation` to make this behavior configurable, but that change is not included in the current pin. This PR therefore adds a temporary override in `OmniFSDPEngine` for NPU FSDP2: synchronize after every micro-batch so gradients remain sharded. Gradient accumulation and the optimizer-step boundary are preserved, at the cost of additional reduce-scatter communication. Other device/backend combinations retain the upstream behavior.

Once the verl dependency includes #7458, remove the local override and use the upstream recipe setting:

```bash
actor_rollout_ref.actor.fsdp_config.use_no_sync_for_gradient_accumulation=false
```

#### Related work

Searches for open PRs covering `NExT-QA`, `416 in:body`, and `Qwen3 video` found no duplicate of this complete recipe. Related video support was proposed in #305, which is closed. #547 addresses interleaved audio-video prompt normalization; this recipe uses separate audio and video inputs.

> [!IMPORTANT]

- [x] Add unit or end-to-end test(s) to [the CI workflow](https://github.com/verl-project/verl-omni/tree/main/.github/workflows) to cover all the code. If not feasible, explain why: Regression coverage for this replacement revision remains to be added and run; the branch-added tests were removed during consolidation. The manual training results do not replace CI coverage. No claim of infeasibility is made.

AI assistance (OpenAI Codex) was used for implementation consolidation and documentation.


tests/utils/test_omni_rl_datasets_on_cpu.py
```diff
@@ -68,3 +68,12 @@ def fake_process_mm_info(messages, use_audio_in_video):
     np.testing.assert_array_equal(result[2][0][:L], audios[0])
     np.testing.assert_array_equal(result[2][0][L:], 0.0)
     assert calls == [(messages, False)]
+
+    result = QwenOmniRLHFDataset._process_multi_modal_info(
+        messages,
+        image_patch_size=14,
+        config={"mm_processor_kwargs": {"use_audio_in_video": True}},
+    )
```
ZihaoW123, line 76: This assertion compares the padded audio returned by `_process_multi_modal_info` (8,160 samples for the 8,001-sample fixture) with the original unpadded `audios` list. NumPy attempts to compare shapes `(8160,)` and `(8001,)` and raises a `ValueError`, so this test will fail once the CPU workflow runs with vLLM installed. Please assert the padded length and contents, as in the checks above, instead of using tuple equality.

verl_omni/utils/dataset/omni_rl_datasets.py
```diff
@@ -51,13 +54,23 @@ def _process_multi_modal_info(
...
+            with warnings.catch_warnings():
+                warnings.filterwarnings(
+                    "ignore",
+                    category=FutureWarning,
+                    message=r".*__audioread_load.*",
+                )
+                audios, images, videos = process_mm_info(messages, use_audio_in_video=use_audio_in_video)
+        except Exception as error:
+            raise RuntimeError(f"Failed to process multimodal sample: {messages}") from error
+
```
ZihaoW123: This blank line contains trailing whitespace. `git diff --check` reports it, and the `ruff-format` pre-commit hook will rewrite the file, causing the pre-commit job to fail. Please remove the whitespace.

ZihaoW123 (commented):
Two blocking issues found in the current revision.

tests/utils/test_omni_rl_datasets_on_cpu.py
SelfDLer: Fixed. The video-audio path now asserts the padded length and contents instead of comparing the padded result with the original unpadded array.

verl_omni/utils/dataset/omni_rl_datasets.py
SelfDLer: Fixed. The trailing whitespace was removed, and `git diff --check` passes.

verl_omni/workers/rollout/vllm_rollout/utils.py
```diff
@@ -174,23 +174,35 @@ def update_weights_from_ipc(
...
-                from vllm.model_executor.model_loader.utils import process_weights_after_loading
 
-                process_weights_after_loading(model, model_config, self.device)
+                model, model_config = standard
+
+                # vLLM records checkpoint/model-layout metadata during model initialization.
+                # Restore kernel-layout parameters to a loadable layout and wrap each weight_loader.
+                initialize_layerwise_reload(model)
+
+                try:
```
ZihaoW123: Could we add a focused CPU mock test for this glue? At minimum, please verify `initialize -> bucket loads -> finalize -> synchronize` on success, and that `finalize` is attempted before re-raising a receive/load failure. The NPU run covers the happy path, but this shared standard-vLLM path also affects GPU workers.

zyxzyxzyx:
# Field report: 100-step GSPO training on 16× Ascend 910C — recipe validated, 2 crash root causes found & fixed

Hi, thanks for this PR — it makes Qwen3-Omni RL training on Ascend work end-to-end. We ran the official NextQA recipe from this PR to completion on a **single node, 16× Ascend 910C (64GB, Atlas 800T A3)** and want to report back: the recipe works, and we hit two crash root causes that will likely bite anyone running the V1/TransferQueue rollout path with video data on high-core-count machines. Details + fixes below.

**Environment** (aligned with the PR's §2.1 matrix): CANN 9.0.0, torch/torch_npu 2.10.0(+post2), vllm 0.27.0 + vllm-ascend d5e98160 + vllm-omni 44448565, verl c4b389a, verl-omni c2aa856, transformers 5.10.4, ray 2.58.0, TransferQueue 0.1.8, Qwen3-Omni-30B-A3B-Instruct, official NExT-QA with `use_audio_in_video=true`, GSPO (`clip_ratio 3e-4/4e-4/c=10`), `filter_overlong_prompts_workers=8`, `val_max_samples=256`.

**Result**: 100 steps completed, ~8h wall (incl. preprocessing + 11 validations), zero crashes after the fixes below, 0 OOM, 0 CANN errors.
- val accuracy (256 samples, greedy): 0.206 (step 2) → 0.389 (step 30) → **0.512 peak (step 70)** → 0.504 (step 100) — **+144%**
- train reward (`critic/score/mean`): 0.121 → 0.320 (+164%); entropy 2.28 → 1.41 (healthy convergence, no collapse)
- `rollout_actor_probs_pearson_corr` 0.94–0.96 throughout — the vLLM 0.27 layerwise-reload weight sync path holds up over 100 steps
- NPU memory steady at 37–38/64GB; ~250s/step

---

## Crash 1 (most important): pthread_create EAGAIN thread flood → `Fatal Python error: Aborted` in AgentLoopWorkerTQ at step ~28

**Symptom**: all 8 `AgentLoopWorkerTQ` actors abort within ~40s at step 28 → `ActorUnavailableError` → training dies. Worker logs show libzmq `Resource temporarily unavailable (src/thread.cpp:241)`; one worker aborts inside Ray's `parallel_memcopy` → `std::thread::_M_start_thread` → `std::__throw_system_error`.

**Root cause** (verified with a `pthread_create` interposer + cgroup counters):
1. Every torch process defaults OpenMP to **320 threads** on a 320-core host;
2. The TransferQueue client's `with_zmq_socket` decorator (`transfer_queue/utils/zmq_utils.py`) creates a **new `zmq.asyncio.Context()` per RPC** and terminates it after each call — each context spawns/joins its IO+reaper pthread pair. `async_kv_put` → `async_kv_retrieve_meta` + `async_set_custom_meta` means ~550 decorated RPCs / step / actor; with 8 actors that's **~8,800 pthread create+destroy per training step** (~250k over 28 steps);
3. Combined with a second swscale thread leak (Crash 2 below, which also ran inside the rollout workers), container task count spiked to **70k–116k** (cgroup `pids.events` recorded 10k+ fork failures), and eventually `pthread_create` returns EAGAIN; libzmq asserts and aborts the worker.

**Fix** (training then ran 9.5h / 98 steps + 11 validations with zero crashes):
```bash
export OMP_NUM_THREADS=8 TQ_NUM_THREADS=8 OPENBLAS_NUM_THREADS=8 MKL_NUM_THREADS=8
```
**Suggestion for the PR**: the V1/TransferQueue path would be much more robust on many-core hosts if the client reused one `zmq.asyncio.Context` per (client, event-loop) instead of creating/terminating per RPC (a per-loop cache with `weakref.finalize(loop, ctx.term)` for short-lived loops works and is behavior-preserving — we ran this patch with unit tests). Even just documenting these env caps in the recipe launcher would help.

## Crash 2: libswscale thread-pool explosion in the video decode path (~5,000 threads per video)

**Symptom**: data workers ballooning to 10k–30k threads; container task spikes; intermittent `av.error.BlockingIOError [swscaler]` and "0 frames" decode failures under concurrency (the reason we had to drop `filter_overlong_prompts_workers` from the recipe's 64 to 8).

**Root cause**: `torchvision.io.read_video` (av backend, used via `qwen_vl_utils.process_vision_info`) collects **all decoded `av.Frame`s first, then batch-converts**. In that pattern each `frame.to_rgb().to_ndarray()` leaks a ~15-thread libswscale slice pool → 332-frame video ≈ **4,988 persistent threads**; different resolutions create additional pools. Our `pthread_create` interposer attributed 4,995/5,002 clones to `av.libs/libswscale-*.so`. A minimal reproducer: batch decode-then-convert = 4,988 threads; interleaved decode+convert = 31 threads, pixel-identical output.

**Fix**: convert each frame inside the decode loop (eager convert). We patched `torchvision/io/video.py` (`_read_from_stream` with an eager-convert wrapper) — thread peak 4,999 → 66, `torch.equal` output identical. Interleaving alone (e.g. PyAV `to_image()` per frame in the generator, as `QwenOmniRLHFDataset`'s own audio-path decoder effectively does) also avoids it. Combined with `filter_overlong_prompts_workers=8`, preprocessing of 10,447 video samples runs cleanly.

## Smaller compat notes for the vllm 0.27 line (all resolved, in case others hit them)

1. `verl/utils/vllm/npu_vllm_patch.py`: `from vllm.model_executor.layers.fused_moe import FusedMoE` hard-fails on vllm 0.27 (FusedMoE modularized away) — needs try/except (the patched function is a no-op for non-class inputs anyway).
2. After installing vllm 0.27 and uninstalling triton, torch 2.10 dynamo needs `triton.language` → install `triton-ascend==3.2.1` (its wheel provides the `triton` package).
3. vllm-ascend needs the catlass submodule (`csrc/third_party/catlass`, not in tarballs).
4. Tarball installs show version `0.1.dev1` — fix by `git init && git commit && git tag <version>` before `pip install -e`.
5. transformers must be 5.10.x to satisfy verl's `<5.11` ∩ vllm-omni's `>=5.10.1` simultaneously.
6. numpy gets bumped ≥2 by transitive deps (scikit-image / opencv chain) which breaks numba in the librosa audio path — pin 1.26.4 last with `--force-reinstall --no-deps`, and use `--no-deps` for every manual install in this env.
7. `cosmos-guardrail` pulls full opencv → needs system libGL; `opencv-python-headless --no-deps` works.
8. `reward.custom_reward_function.path` must be **absolute** — Ray workers' cwd is the launch dir, not the repo root (the official launcher's relative path only works inside the official image).
9. The official `nextqa.py` uses `ffprobe` to filter audio-less videos; containers with only ffmpeg silently mis-detect 0 audio tracks. Either ship an ffprobe shim or (better) detect audio with `av.open(p).streams.audio`.

---

Happy to share the full logs (7.4MB training log, crash-scene logs, the pthread_create interposer, and the torchvision patch) if useful. Also happy to open a PR for the TransferQueue context reuse and/or the torchvision eager-convert if the maintainers think they belong here.

Thanks again for the great work — this PR turned Qwen3-Omni Omni-RL on Ascend from "doesn't run" to "8.5-hour unattended 100-step run with a +12pp val gain".

— @zyxzyxzyx

zyxzyxzyx:
# 200-step continuation report (follow-up to the 100-step field report)

Following up on the [100-step report](https://github.com/verl-project/verl-omni/pull/416#issuecomment-5466715974): on the same environment (single node, 16× Ascend 910C 64GB, verl-omni c2aa856 official NextQA recipe) we have completed **step 100 → 200 via checkpoint resume** (`resume_mode=auto` + `total_training_steps=200`, every other setting identical to the 100-step run).

## Results

- **val accuracy** (256 samples, greedy): 0.480 at the step-100 resume point → **0.552 at step 200 (all-time high)**; the 100→200 segment plateaus at 0.48–0.55, a +2~4pp lift over the 100-step segment (0.48–0.51); cumulative **+16.3pp** over the first evaluation point (0.389)
- **train reward** (critic/score/mean): 0.121 (step 1) → 0.320 (step 100) → 0.332 (step 200), **+174%** over the full campaign
- **Stability**: with the thread governance (OMP/TQ/OPENBLAS/MKL=8) plus the two code fixes, the 100-step and 200-step runs together total **~17.3h with zero crashes, zero OOM, zero crash signatures**; container task count stayed at 18k–36k
- **Round-trip inference validation**: resume from `global_step_200` + val_before_train = **0.5556** (consistent with the 0.552 final training-time eval within noise) — the checkpoint is directly deliverable

## New key finding: a major thread leak in the torchvision video decode path (further quantification of Crash 2)

`torchvision.io.read_video` (av backend, used by the `qwen_vl_utils` video path) **buffers all decoded `av.Frame`s first, then batch-converts**. In that pattern each `frame.to_rgb().to_ndarray()` leaks a ~15-thread libswscale slice pool, so a 332-frame video leaves behind ≈ **4,988 persistent threads**, with additional pools per resolution — this was the biggest amplifier pushing our container to 116k tasks (it compounded with Crash 1 in the 100-step report and with a SIGKILL during model loading). A `pthread_create` interposer attributed 4,995/5,002 clones to `av.libs/libswscale-*.so`.

**Fix**: convert each frame inside the decode loop (semantics unchanged — output verified pixel-identical via `torch.equal`), which drops the per-video thread peak from 4,999 to 66. Combined with `filter_overlong_prompts_workers=8`, the data pipeline ran clean for the entire 200-step campaign. It would be great if upstream could offer an eager-convert option, or at least document a worker-count ceiling for video datasets on high-core-count machines.

## Conclusion

The PR#416 recipe remains stable and keeps improving at the 200-step scale with no signs of saturation — val accuracy was still climbing at the final step. We'll keep the same thread-governance setup for longer runs and other datasets. Thanks again for the great work!

— @zyxzyxzyx

SelfDLer:
> # 200-step continuation report (follow-up to the 100-step field report)
> Following up on the [100-step report](https://github.com/verl-project/verl-omni/pull/416#issuecomment-5466715974): on the same environment (single node, 16× Ascend 910C 64GB, verl-omni [c2aa856](https://github.com/verl-project/verl-omni/commit/c2aa856d98e06c05d542ee5776d32251b5bb8d5e) official NextQA recipe) we have completed **step 100 → 200 via checkpoint resume** (`resume_mode=auto` + `total_training_steps=200`, every other setting identical to the 100-step run).
> 
> ## Results
> * **val accuracy** (256 samples, greedy): 0.480 at the step-100 resume point → **0.552 at step 200 (all-time high)**; the 100→200 segment plateaus at 0.48–0.55, a +2~4pp lift over the 100-step segment (0.48–0.51); cumulative **+16.3pp** over the first evaluation point (0.389)
> * **train reward** (critic/score/mean): 0.121 (step 1) → 0.320 (step 100) → 0.332 (step 200), **+174%** over the full campaign
> * **Stability**: with the thread governance (OMP/TQ/OPENBLAS/MKL=8) plus the two code fixes, the 100-step and 200-step runs together total **~17.3h with zero crashes, zero OOM, zero crash signatures**; container task count stayed at 18k–36k
> * **Round-trip inference validation**: resume from `global_step_200` + val_before_train = **0.5556** (consistent with the 0.552 final training-time eval within noise) — the checkpoint is directly deliverable
> 
> ## New key finding: a major thread leak in the torchvision video decode path (further quantification of Crash 2)
> `torchvision.io.read_video` (av backend, used by the `qwen_vl_utils` video path) **buffers all decoded `av.Frame`s first, then batch-converts**. In that pattern each `frame.to_rgb().to_ndarray()` leaks a ~15-thread libswscale slice pool, so a 332-frame video leaves behind ≈ **4,988 persistent threads**, with additional pools per resolution — this was the biggest amplifier pushing our container to 116k tasks (it compounded with Crash 1 in the 100-step report and with a SIGKILL during model loading). A `pthread_create` interposer attributed 4,995/5,002 clones to `av.libs/libswscale-*.so`.
> 
> **Fix**: convert each frame inside the decode loop (semantics unchanged — output verified pixel-identical via `torch.equal`), which drops the per-video thread peak from 4,999 to 66. Combined with `filter_overlong_prompts_workers=8`, the data pipeline ran clean for the entire 200-step campaign. It would be great if upstream could offer an eager-convert option, or at least document a worker-count ceiling for video datasets on high-core-count machines.
> 
> ## Conclusion
> The PR#416 recipe remains stable and keeps improving at the 200-step scale with no signs of saturation — val accuracy was still climbing at the final step. We'll keep the same thread-governance setup for longer runs and other datasets. Thanks again for the great work!
> 
> — @zyxzyxzyx

Thanks a lot for the detailed investigation. I ran an independent A/B stress test on a similar high-core-count environment to check whether the video-decoding issue also affects our TorchCodec path.

The test node has:

* 320 CPU cores
* 2.4 TiB RAM
* PyTorch 2.10.0
* TorchCodec 0.10.0a0
* torchvision 0.25.0
* PyAV 18.1.0

I tested both repeated decoding of the same NextQA video and sequential decoding of different videos.

### Video decoding results

| Workload                        | TorchCodec peak threads | torchvision peak threads | Decode errors (TorchCodec / torchvision) |
| ------------------------------- | ----------------------: | -----------------------: | ---------------------------------------: |
| 1 worker, same video ×300       |                      11 |                   11,268 |                                    0 / 0 |
| 2 workers, same video ×100 each |                      22 |                   22,536 |                                    0 / 0 |
| 4 workers, same video ×100 each |                      44 |                   36,224 |                                    0 / 0 |
| 1 worker, 300 different videos  |                      18 |                   31,698 |                                   0 / 48 |

For the 300-different-video TorchCodec test, the process started at 11 threads, reached 18 threads within the first few samples, and then remained at 18 from approximately sample 20 through sample 300.

RSS also stayed roughly flat around 1.2 GB. Therefore, I did not observe a persistent cross-call thread leak in the TorchCodec path. The small increase from 11 to 18 threads appears to be one-time/lazy initialization rather than continuous thread accumulation.

The torchvision/PyAV path behaved very differently. With only one worker, the native thread count transiently reached ~31.7k. 48/300 video decodes failed with the same error:

```text
BlockingIOError: [Errno 11] Resource temporarily unavailable
```

The underlying libswscale error was:

```text
Failed initializing scaling graph (Resource temporarily unavailable):
fmt:yuv420p -> fmt:rgb24
```

The result was reproducible across two runs:

* first run: 31,683 peak threads, 48 errors
* repeated run: 31,698 peak threads, 48 errors

One clarification from these measurements is that I would describe the torchvision behavior as a **severe transient native-thread explosion**, rather than a persistent cross-call thread leak.

After an individual decode returns, the thread count can fall back to baseline, but the in-flight decoding peak alone is large enough to cause resource exhaustion. Under a high-concurrency preprocessing or AgentLoop workload, this level of task pressure could also contribute to unrelated `pthread_create()` calls in components such as ZMQ/TransferQueue failing with `EAGAIN`.

For TorchCodec, with the frame-index sampling path used by Qwen Omni and `num_ffmpeg_threads=1`, I have not reproduced this behavior even when cycling through different NextQA videos.

### TransferQueue side

I also checked the TransferQueue version used in our current environment.

We are currently using:

```text
TransferQueue 0.1.8
```

Inspecting the installed implementation confirms that its `with_zmq_socket` wrapper creates a new:

```python
zmq.asyncio.Context()
```

for each RPC and terminates it afterward.

The current upstream TransferQueue implementation has changed this behavior to reuse a long-lived ZMQ context per client instead.

So the TransferQueue-side issue and the video-decoding issue appear to be two separate sources of thread pressure:

1. **TransferQueue 0.1.8:** per-RPC ZMQ context creation/destruction.
2. **torchvision/PyAV:** very large transient native-thread growth during video decoding.

These two behaviors can amplify each other under a high-concurrency video workload.

### Current conclusion

My current understanding is therefore:

* The per-RPC ZMQ context churn is present in the TransferQueue 0.1.8 code path used by our current environment.
* The current upstream TransferQueue implementation already uses a long-lived ZMQ context per client, so moving to a revision containing that change should address this part independently of the NextQA recipe.
* The large video-decoding thread explosion is reproducible on the torchvision/PyAV path.
* The TorchCodec path appears stable under the same 320-core environment and NextQA workload.
* We should make sure the NExT-QA recipe stays on TorchCodec and does not unexpectedly fall back to torchvision.
* Conservative preprocessing worker/thread limits are still useful as additional protection, especially on high-core-count hosts.

I also updated the NExT-QA README accordingly. It now explicitly recommends TorchCodec for the NPU video-decoding path and documents how to build TorchCodec without CUDA support for Ascend/NPU environments.

Thanks again for identifying these two failure modes — the report was very helpful for narrowing this down.

verl_omni/workers/rollout/vllm_rollout/utils.py
SelfDLer: Thanks for the suggestion. Added a focused CPU mock test in `2186cb5`.

It now covers both cases you mentioned:

* success path: `initialize -> bucket loads -> finalize -> synchronize`
* failure paths: both receive and bucket-load failures verify that `finalize` is attempted before the original exception is re-raised

The failure tests also verify that `synchronize` is not reached after a receive/load failure.

ZihaoW123:
@SelfDLer please fix ci

zyxzyxzyx:
# Field report 3: TorchCodec decode backend — 200-step rerun confirms the README guidance (no thread growth, ~40% faster steps)

Hi again — following up on our two earlier field reports (100 + 200 steps on the torchvision/PyAV decode path, which needed the per-frame eager-convert workaround to stop the libswscale thread leak). After the README gained the TorchCodec recommendation (`113a13f7`), we rebuilt the environment with TorchCodec and re-ran the full 200-step GSPO training **from scratch** — same recipe, same hyperparameters, single variable: the video decode backend.

**Environment** (same node, 16× Ascend 910C 64GB; same stack as before: CANN 9.0.0, torch/torch_npu 2.10.0, vllm 0.27.0 + vllm-ascend d5e98160 + vllm-omni 44448565, verl c4b389a, verl-omni c2aa856, transformers 5.10.4, ray 2.58.0, official NExT-QA with `use_audio_in_video=true`), plus:

- **TorchCodec 0.10.0 built from source exactly per the README section**: FFmpeg 4.4.2 dev packages from the Ubuntu 22.04 (jammy) repos, aarch64, `ENABLE_CUDA=0 I_CONFIRM_THIS_IS_NOT_A_LICENSE_VIOLATION=1`, `pip install . --no-build-isolation --no-deps` against torch 2.10.0+cpu. Build was clean; imports fine in every Ray worker; numpy/torch untouched (we keep a strict `--no-deps` + post-install numpy/torch check discipline).
- `FORCE_QWENVL_VIDEO_READER=torchcodec` to pin the backend explicitly. The default priority torchcodec→decord→torchvision already picks TorchCodec up once installed; we pinned it so a backend failure degrades loudly (the fallback warning is greppable) rather than silently.

**Result** — fresh 0→200 run, 9.1 h wall including preprocessing and 21 validations, **zero crashes, zero decode failures, zero fallbacks** across all 18 worker processes:

- val accuracy (252–256 samples, greedy): 0.290 (before train) → 0.556 (step 80) → **peak 0.560 (step 150)** → 0.536 (step 200 final). Same plateau as our PyAV-path run (which peaked 0.552 / finished 0.552 over 0→200 with resume).
- train reward (`critic/score/mean`): 0.156 → 0.371 (+138%); entropy 2.56 → ~1.2 with no collapse; `rollout_actor_probs_pearson_corr` 0.910–0.970 throughout.
- NPU memory ≤ 39.2/64 GB, 0 OOM, 0 CANN errors. **~153 s/step average (200 steps) vs ~250 s/step in our PyAV-path run** — with TorchCodec the decode path is no longer a visible bottleneck in the rollout loop.
- Checkpoint round-trip: `resume_mode=auto` from `global_step_200` + `val_before_train` (greedy, 256 samples) → **0.548** — matches the in-training step-200 eval (0.536), checkpoint deliverable.

**Thread observation (the reason for this rerun)**: with TorchCodec the decode-path thread flood is simply gone —

- Container task count stayed flat at ~34–37k for the entire 9 h (our warning line is 50k; the pre-fix PyAV flood hit 70–116k and died at step ~28).
- No libswscale growth: our pre-flight check decoded 30 videos / 300 frames with **zero thread delta** in the decoding process; the torchvision/PyAV path leaked ~5,000 persistent threads per 332-frame video before the eager-convert workaround.
- The per-frame eager-convert monkey-patch is no longer needed on the decode path. The OMP/TQ/OPENBLAS/MKL=8 governance is still required, though — Crash 1 (TransferQueue client creating a new `zmq.asyncio.Context()` per RPC) is upstream of the decode backend and unaffected.

**One small suggestion**: consider adding `export FORCE_QWENVL_VIDEO_READER=torchcodec` (or a one-line note) to the launcher — if TorchCodec is installed but broken at runtime, the current code silently falls back to torchvision with a single warning line, which would quietly reintroduce the leak this section exists to avoid. A pinned backend turns that into a loud, greppable failure.

Thanks — the README build instructions worked as written on aarch64 + FFmpeg 4.4, and the recommendation checks out under a full 200-step run.

SelfDLer:
> # Field report 3: TorchCodec decode backend — 200-step rerun confirms the README guidance (no thread growth, ~40% faster steps)
> Hi again — following up on our two earlier field reports (100 + 200 steps on the torchvision/PyAV decode path, which needed the per-frame eager-convert workaround to stop the libswscale thread leak). After the README gained the TorchCodec recommendation (`113a13f7`), we rebuilt the environment with TorchCodec and re-ran the full 200-step GSPO training **from scratch** — same recipe, same hyperparameters, single variable: the video decode backend.
> 
> **Environment** (same node, 16× Ascend 910C 64GB; same stack as before: CANN 9.0.0, torch/torch_npu 2.10.0, vllm 0.27.0 + vllm-ascend d5e98160 + vllm-omni 44448565, verl c4b389a, verl-omni [c2aa856](https://github.com/verl-project/verl-omni/commit/c2aa856d98e06c05d542ee5776d32251b5bb8d5e), transformers 5.10.4, ray 2.58.0, official NExT-QA with `use_audio_in_video=true`), plus:
> 
> * **TorchCodec 0.10.0 built from source exactly per the README section**: FFmpeg 4.4.2 dev packages from the Ubuntu 22.04 (jammy) repos, aarch64, `ENABLE_CUDA=0 I_CONFIRM_THIS_IS_NOT_A_LICENSE_VIOLATION=1`, `pip install . --no-build-isolation --no-deps` against torch 2.10.0+cpu. Build was clean; imports fine in every Ray worker; numpy/torch untouched (we keep a strict `--no-deps` + post-install numpy/torch check discipline).
> * `FORCE_QWENVL_VIDEO_READER=torchcodec` to pin the backend explicitly. The default priority torchcodec→decord→torchvision already picks TorchCodec up once installed; we pinned it so a backend failure degrades loudly (the fallback warning is greppable) rather than silently.
> 
> **Result** — fresh 0→200 run, 9.1 h wall including preprocessing and 21 validations, **zero crashes, zero decode failures, zero fallbacks** across all 18 worker processes:
> 
> * val accuracy (252–256 samples, greedy): 0.290 (before train) → 0.556 (step 80) → **peak 0.560 (step 150)** → 0.536 (step 200 final). Same plateau as our PyAV-path run (which peaked 0.552 / finished 0.552 over 0→200 with resume).
> * train reward (`critic/score/mean`): 0.156 → 0.371 (+138%); entropy 2.56 → ~1.2 with no collapse; `rollout_actor_probs_pearson_corr` 0.910–0.970 throughout.
> * NPU memory ≤ 39.2/64 GB, 0 OOM, 0 CANN errors. **~153 s/step average (200 steps) vs ~250 s/step in our PyAV-path run** — with TorchCodec the decode path is no longer a visible bottleneck in the rollout loop.
> * Checkpoint round-trip: `resume_mode=auto` from `global_step_200` + `val_before_train` (greedy, 256 samples) → **0.548** — matches the in-training step-200 eval (0.536), checkpoint deliverable.
> 
> **Thread observation (the reason for this rerun)**: with TorchCodec the decode-path thread flood is simply gone —
> 
> * Container task count stayed flat at ~34–37k for the entire 9 h (our warning line is 50k; the pre-fix PyAV flood hit 70–116k and died at step ~28).
> * No libswscale growth: our pre-flight check decoded 30 videos / 300 frames with **zero thread delta** in the decoding process; the torchvision/PyAV path leaked ~5,000 persistent threads per 332-frame video before the eager-convert workaround.
> * The per-frame eager-convert monkey-patch is no longer needed on the decode path. The OMP/TQ/OPENBLAS/MKL=8 governance is still required, though — Crash 1 (TransferQueue client creating a new `zmq.asyncio.Context()` per RPC) is upstream of the decode backend and unaffected.
> 
> **One small suggestion**: consider adding `export FORCE_QWENVL_VIDEO_READER=torchcodec` (or a one-line note) to the launcher — if TorchCodec is installed but broken at runtime, the current code silently falls back to torchvision with a single warning line, which would quietly reintroduce the leak this section exists to avoid. A pinned backend turns that into a loud, greppable failure.
> 
> Thanks — the README build instructions worked as written on aarch64 + FFmpeg 4.4, and the recommendation checks out under a full 200-step run.

Thanks for the detailed validation! Great to see the 200-step rerun confirms the TorchCodec recommendation, with stable thread count and a clear speedup.

Good point on `FORCE_QWENVL_VIDEO_READER=torchcodec` as well — making backend fallback fail loudly would be safer. We’ll consider adding this to the launcher or README.

examples/gspo_trainer/qwen3_omni/run_qwen3_omni_thinker_gspo_npu_nextqa_v1.sh
```diff
@@ -0,0 +1,119 @@
...
+    data.val_files="${VAL_FILE}" \
+    data.train_batch_size=32 \
+    data.max_prompt_length=4096 \
+    data.max_response_length=1024 \
+    data.shuffle=true \
+    data.seed=42 \
+    data.val_max_samples=-1 \
+    data.validation_shuffle=false \
+    data.filter_overlong_prompts=true \
+    data.filter_overlong_prompts_workers=64 \
```
SamitHuang: This still defaults to 64 preprocessing workers and does not pin the video backend. Independent 320-core NPU runs died from torchvision/PyAV thread explosion plus TransferQueue 0.1.8 ZMQ context churn; they only completed after `filter_overlong_prompts_workers=8`, `OMP/TQ/OPENBLAS/MKL_NUM_THREADS=8`, and `FORCE_QWENVL_VIDEO_READER=torchcodec`. README already recommends TorchCodec, but qwen_vl_utils can silently fall back to torchvision. Export those env vars in this launcher and assert them in `test_nextqa_npu_launcher_wires_video_gspo_training` so the recipe cannot reintroduce that crash path.

examples/gspo_trainer/data_process/nextqa.py
```diff
@@ -0,0 +1,414 @@
...
+    output_path.parent.mkdir(parents=True, exist_ok=True)
+    # Dictionary-encoded nested columns can fail in some supported PyArrow
+    # versions, so match the AVQA converter and keep nested columns plain.
+    pd.DataFrame(rows).to_parquet(output_path, engine="pyarrow", index=False, use_dictionary=False)
+
+
+def probe_audio_stream(video_path: str | Path) -> str | None:
+    """Return a stable drop reason when ffprobe cannot find usable audio metadata."""
+    try:
+        result = subprocess.run(
```
SamitHuang, line 222: `ffprobe` is invoked with no timeout. One hung or corrupt MP4 blocks conversion of the whole split; the unique-video cache does not help until that call returns. Add `timeout=` and map `TimeoutExpired` to `invalid_media`. Stream-index presence is also not the same as a decodable audio track — field reports saw stub `ffprobe` shims silently drop audio — so either decode a short probe with `av`/`ffmpeg` or document that a real `ffprobe` (not just `ffmpeg`) is required.

verl_omni/utils/dataset/omni_rl_datasets.py
```diff
@@ -51,13 +54,23 @@ def _process_multi_modal_info(
...
+        try:
+            with warnings.catch_warnings():
+                warnings.filterwarnings(
+                    "ignore",
+                    category=FutureWarning,
+                    message=r".*__audioread_load.*",
+                )
+                audios, images, videos = process_mm_info(messages, use_audio_in_video=use_audio_in_video)
+        except Exception as error:
+            raise RuntimeError(f"Failed to process multimodal sample: {messages}") from error
```
SamitHuang: Wrapping every `Exception` and interpolating `{messages}` dumps the full multimodal payload into Ray worker logs (paths plus any inlined media) and turns MemoryError/decode failures into a generic RuntimeError. Keep `from error`, but log `extra_info` identifiers (video_id/qid) instead of the whole message list. The new CPU tests only cover the happy path for `use_audio_in_video`; add a case that a `process_mm_info` failure still chains the original exception.

verl_omni/workers/rollout/vllm_rollout/utils.py
```diff
@@ -180,23 +180,35 @@ def update_weights_from_ipc(
...
+                    initialize_layerwise_reload,
                 )
-                from vllm.model_executor.model_loader.utils import process_weights_after_loading
 
-                process_weights_after_loading(model, model_config, self.device)
+                model, model_config = standard
+
+                # vLLM records checkpoint/model-layout metadata during model initialization.
+                # Restore kernel-layout parameters to a loadable layout and wrap each weight_loader.
+                initialize_layerwise_reload(model)
```
SamitHuang: This replaces `patch_vllm_moe_model_weight_loader` + `process_weights_after_loading` on the shared standard-vLLM path, not only NPU. `initialize_layerwise_reload` skips layers that never ran `record_metadata_for_reloading` (`can_reload()`), and the new CPU tests mock both hooks, so a CUDA worker can skip Ascend-style finalize *and* skip the old post-load pass. Keep the MoE `weight_loader` patch, or gate layerwise reload on NPU/`can_reload` and fall back to `process_weights_after_loading` otherwise. NPU Pearson on this recipe is not coverage for GPU full-weight sync.

SamitHuang (changes requested):
NExT-QA conversion + GSPO NPU recipe is the right shape, and the 100/200-step Atlas runs plus CPU coverage for the converter, `use_audio_in_video`, FSDP2 NPU sync, and layerwise reload glue are useful. FSDP2 keeping gradient sync enabled on NPU looks consistent with the stated peak-memory issue.

Four things still need to land before this is a safe default recipe / shared worker change:

1. Launcher still uses `filter_overlong_prompts_workers=64` and does not pin TorchCodec or cap `OMP/TQ` threads — the crash mode independently reproduced on 320-core hosts.
2. Layerwise reload replaces MoE `weight_loader` patch + `process_weights_after_loading` on the shared GPU path; CPU tests mock the hooks so they cannot catch a missing `record_metadata_for_reloading` fallback.
3. `_process_multi_modal_info` dumps full `messages` on every exception.
4. `ffprobe` has no timeout and only checks stream index.

Not re-litigating Zihao's resolved audio-padding / reload-finalize threads.

SamitHuang:
@zhtmike PTAL

zhtmike:
Pls update a reward curve.png for our reference in future. Not merely numbers.

SelfDLer:
> NExT-QA conversion + GSPO NPU recipe is the right shape, and the 100/200-step Atlas runs plus CPU coverage for the converter, `use_audio_in_video`, FSDP2 NPU sync, and layerwise reload glue are useful. FSDP2 keeping gradient sync enabled on NPU looks consistent with the stated peak-memory issue.
> 
> Four things still need to land before this is a safe default recipe / shared worker change:
> 
> 1. Launcher still uses `filter_overlong_prompts_workers=64` and does not pin TorchCodec or cap `OMP/TQ` threads — the crash mode independently reproduced on 320-core hosts.
> 2. Layerwise reload replaces MoE `weight_loader` patch + `process_weights_after_loading` on the shared GPU path; CPU tests mock the hooks so they cannot catch a missing `record_metadata_for_reloading` fallback.
> 3. `_process_multi_modal_info` dumps full `messages` on every exception.
> 4. `ffprobe` has no timeout and only checks stream index.
> 
> Not re-litigating Zihao's resolved audio-padding / reload-finalize threads.

Thanks for the thorough review. I addressed all four points in commit https://github.com/verl-project/verl-omni/commit/213722ac62bf5968ec20a5097d7413a095a3e2b9:

The NExT-QA launcher now defaults preprocessing workers and OMP/TQ/OPENBLAS/MKL threads to 8, and explicitly selects TorchCodec. All values remain user-overridable.
Layerwise reload is now used only on NPU when complete reload metadata is available. The MoE loader patch is retained, while CUDA and missing-metadata cases fall back to process_weights_after_loading. Regression tests cover both NPU and CUDA fallback paths.
Multimodal processing errors no longer include the full message payload, while preserving the original exception through chaining. A regression test covers this behavior.
Both ffprobe and the one-frame ffmpeg audio decode probe now have timeouts. Timeout and decoding failures are mapped to invalid_media, with corresponding tests.

SelfDLer:
> Pls update a reward curve.png for our reference in future. Not merely numbers.

Thanks for the suggestion. I've added the training reward, validation accuracy, rollout-actor pearson corelation and actor entropy curves below, covering the full 100-step run.
training_metrics

SelfDLer:
Hi @SamitHuang, I’ve addressed the previous comments and added the corresponding tests. When you have time, could you please take another look? Thanks!

zhtmike:
actually, the actor & rollout Pearson. corr. is a bit low.

SelfDLer:
> actually, the actor & rollout Pearson. corr. is a bit low.

Thanks for pointing this out. The actor–rollout Pearson correlation is around `0.94–0.95`, which is indeed lower than ideal, although the validation accuracy improves consistently from `27.57%` to `53.44%`.

I’ll continue investigating the actor–rollout numerical differences. Do you consider this a blocker for merging the recipe, or would it be acceptable to track it as a follow-up?

zhtmike:
> > actually, the actor & rollout Pearson. corr. is a bit low.
> 
> Thanks for pointing this out. The actor–rollout Pearson correlation is around `0.94–0.95`, which is indeed lower than ideal, although the validation accuracy improves consistently from `27.57%` to `53.44%`.
> 
> I’ll continue investigating the actor–rollout numerical differences. Do you consider this a blocker for merging the recipe, or would it be acceptable to track it as a follow-up?

yes, normally it is >0.99 for a normal training.
Like this https://wandb.ai/verl-omni/gspo_demo/runs/kzzrq9pr

verl_omni/workers/rollout/vllm_rollout/utils.py
```diff
@@ -32,6 +32,32 @@ def _split_visible_devices(value: str) -> list[str]:
     return [entry.strip() for entry in value.split(",") if entry.strip()]
 
 
+def _is_npu_platform() -> bool:
```
wtomin, line 35: You can load `_is_npu_platform` from `verl_omni.workers.rollout.vllm_rollout.npu_utils`. Try reduce this kind of patches.

verl_omni/workers/rollout/vllm_rollout/utils.py
```diff
@@ -32,6 +32,32 @@ def _split_visible_devices(value: str) -> list[str]:
...
+    """Return whether vLLM is running on an Ascend NPU platform."""
+    try:
+        from vllm.platforms import current_platform
+
+        return current_platform.device_type == "npu"
+    except Exception:
+        return False
+
+
+def _has_layerwise_reload_metadata(model: torch.nn.Module) -> bool:
```
wtomin, line 45: Could you rebase to the main branch and resolve the conflicts of this `utils.py`? I feel like `restore_moe_param_layout` is deleted from this file.

examples/gspo_trainer/qwen3_omni/run_qwen3_omni_thinker_gspo_npu_nextqa_v1.sh
```diff
@@ -0,0 +1,125 @@
...
+    actor_rollout_ref.actor.policy_loss.loss_mode=gspo \
+    actor_rollout_ref.actor.clip_ratio_low=3e-4 \
+    actor_rollout_ref.actor.clip_ratio_high=4e-4 \
+    actor_rollout_ref.actor.clip_ratio_c=10.0 \
+    actor_rollout_ref.actor.loss_agg_mode=seq-mean-token-mean \
+    actor_rollout_ref.actor.fsdp_config.param_offload=true \
+    actor_rollout_ref.actor.fsdp_config.optimizer_offload=true \
+    actor_rollout_ref.actor.fsdp_config.model_dtype=bfloat16 \
+    actor_rollout_ref.actor.fsdp_config.use_orig_params=true \
+    actor_rollout_ref.actor.fsdp_config.use_torch_compile=False \
```
wtomin, line 82: The default option for `use_torch_compile` is `true`. Any reason to turn it off on npu?

verl_omni/workers/engine/fsdp/omni_impl.py
```diff
@@ -43,6 +44,21 @@
...
     """FSDP engine for omni models"""
 
+    @contextmanager
+    def _gradient_sync_context(self, *, is_last_micro_batch: bool):
+        """Keep NPU FSDP2 gradients sharded throughout accumulation.
+
+        The pinned verl defers reduction on intermediate micro-batches, which
+        retains full FP32 gradients and can exceed 64 GB for the full Thinker.
+        Synchronizing each backward preserves accumulation and the optimizer
+        step boundary, at the cost of additional reduce-scatter communication.
```
zhtmike: why not simply reduce the micro batch size for npu in the script, instead of writing a patch + overhead on this?

zhtmike (changes requested):
as comment

verl_omni/workers/rollout/vllm_rollout/utils.py
SelfDLer: Thanks. The revised patch removes the PR-specific changes to vllm_rollout/utils.py and retains the upstream implementation, which imports _is_npu_platform from npu_utils. This PR no longer introduces a duplicate platform helper.

verl_omni/workers/rollout/vllm_rollout/utils.py
SelfDLer: Thanks for catching this. I rebuilt the patch on main at f863ecc and removed the old rollout-utils modifications. restore_moe_param_layout and its call remain intact, along with the upstream MoE loader patch and post-load processing. There is now no diff in vllm_rollout/utils.py against that base

examples/gspo_trainer/qwen3_omni/run_qwen3_omni_thinker_gspo_npu_nextqa_v1.sh
SelfDLer: These overrides match the NPU configuration used for the training run reported in this PR. I have not validated this recipe with use_torch_compile=true, so I cannot say that disabling compilation is required on NPU. I kept the settings to make the launcher consistent with the reported run; they are not intended as a general recommendation for NPU training.

verl_omni/workers/engine/fsdp/omni_impl.py
SelfDLer: I tested removing the override, and training still OOMs with ppo_micro_batch_size_per_gpu=1, which is already the smallest supported value.
A micro-batch size of 1 does not disable gradient accumulation. With the recipe defaults (ppo_mini_batch_size=16, rollout.n=8, and actor DP=16), each rank processes eight sequences per optimizer update, resulting in eight backward passes. The [pinned verl implementation](https://github.com/verl-project/verl/blob/fefb080262e1c015a0ea05f958822a6a512dc795/verl/workers/engine/fsdp/transformer_impl.py#L610-L674) disables gradient reduction on the first seven passes, retaining unsharded gradients between them. Reducing the micro-batch helps activation memory, but does not shrink those gradient buffers.
The override still accumulates gradients over all eight passes and performs one optimizer update. It reduces and shards gradients after each backward to lower peak NPU HBM usage. You are right that this adds communication overhead; it is a memory/communication trade-off.
Upstream [verl#7458](https://github.com/verl-project/verl/pull/7458) already makes this configurable, but our current verl pin does not include it. Once the dependency includes that change, the override can be removed and replaced with actor_rollout_ref.actor.fsdp_config.use_no_sync_for_gradient_accumulation=false in the recipe.

zhtmike: why not use use_dynamic_bsz=True ?

zhtmike: packed attention can save a lot of memory + faster training; you can check gpu script.

SelfDLer: Thanks for the suggestion. I’ll test packed attention and dynamic batching using the GPU script as a reference, and check whether training runs without the gradient synchronization override.

SelfDLer: I tested with use_remove_padding=True, actor use_dynamic_bsz=True, and rollout log_prob_use_dynamic_bsz=True, but the run still OOMs without the synchronization override. Are there specific token-budget or attention-backend settings from the GPU recipe you would recommend for this full-parameter NPU run?

zhtmike: try to lower ppo_max_token_len_per_gpu

zhtmike: check this examples/gspo_trainer/qwen3_omni/run_qwen3_omni_thinker_gspo_npu_avqa_v1.sh

ZihaoW123: > I tested removing the override, and training still OOMs with ppo_micro_batch_size_per_gpu=1, which is already the smallest supported value. A micro-batch size of 1 does not disable gradient accumulation. With the recipe defaults (ppo_mini_batch_size=16, rollout.n=8, and actor DP=16), each rank processes eight sequences per optimizer update, resulting in eight backward passes. The [pinned verl implementation](https://github.com/verl-project/verl/blob/fefb080262e1c015a0ea05f958822a6a512dc795/verl/workers/engine/fsdp/transformer_impl.py#L610-L674) disables gradient reduction on the first seven passes, retaining unsharded gradients between them. Reducing the micro-batch helps activation memory, but does not shrink those gradient buffers. The override still accumulates gradients over all eight passes and performs one optimizer update. It reduces and shards gradients after each backward to lower peak NPU HBM usage. You are right that this adds communication overhead; it is a memory/communication trade-off. Upstream [verl#7458](https://github.com/verl-project/verl/pull/7458) already makes this configurable, but our current verl pin does not include it. Once the dependency includes that change, the override can be removed and replaced with actor_rollout_ref.actor.fsdp_config.use_no_sync_for_gradient_accumulation=false in the recipe.

You can use this  [verl#7458](https://github.com/verl-project/verl/pull/7458) and document it in the README, then upgrade verl_pin later.

SelfDLer: I tested with actor_rollout_ref.model.use_remove_padding=true, actor_rollout_ref.actor.use_dynamic_bsz=true, and actor_rollout_ref.rollout.log_prob_use_dynamic_bsz=true. I also lowered actor_rollout_ref.actor.ppo_max_token_len_per_gpu and actor_rollout_ref.rollout.log_prob_max_token_len_per_gpu to 10240, but the run still OOMed without explicitly setting the gradient synchronization option.
After setting actor_rollout_ref.actor.fsdp_config.use_no_sync_for_gradient_accumulation=false, the test run proceeded without OOM.
I’ve pushed the update removing our local patch and using the upstream option from [verl#7458](https://github.com/verl-project/verl/pull/7458). The README documents the required verl commit and installation command. The repository-wide dependency pins and pyproject.toml remain unchanged.

zhtmike (approved):

SamitHuang (approved):
LGTM

Issue #90, opened 2026-05-15
[RFC] Qwen3-TTS RL recipe


yuekaizhang:
# [RFC] Qwen3-TTS RL recipe with vLLM-Omni rollout and Qwen3-ASR reward

## Summary

Proposing a new recipe in `verl-omni` for RL post-training of `Qwen3-TTS-12Hz-0.6B-Base` in voice-cloning mode. Rollout uses `vllm-omni==0.18.0`'s Qwen3-TTS pipeline to sample `n>=2` speech-token sequences per prompt; the synthesized waveforms are transcribed by a separately served `vllm==0.18.0` Qwen3-ASR endpoint; the reward is a clipped function of word error rate. Training reuses upstream `verl.trainer.main_ppo` with `algorithm.adv_estimator=grpo` — no new algorithm is registered inside `verl-omni`, and the existing diffusion trainer is not modified.

## End-to-end data flow

```
                                                            remote vLLM Qwen3-ASR server
                                                            (separate process / node)
  Audio Dataset parquet row                                                  ▲
  ─────────────────────                                                │ HTTP STT
  ref_audio, ref_text,                                                 │
  prompt_text, speaker_id,                                             │
  target_duration                                                      │
          │                                                            │
          ▼                                                            │
  Qwen3TTSDataset                                                      │
          │                                                            │
          ▼                                                            │
  AutoRegressiveTTSSingleTurnAgentLoop  ── build request               │
          │                                                            │
          ▼                                                            │
  vllm_omni_async_server.generate_tts                                  │
          │                                                            │
          ▼                                                            │
  vllm-omni 2-stage pipeline                                           │
  ┌─────────────────────────────────────────────────────────┐          │
  │  stage 0 │ qwen3_tts_talker    → codec_tokens + logprobs │         │
  │          │ (AR sampling — see § Stage-0 logprob exposure) │         │
  │  stage 1 │ qwen3_tts_code2wav  → waveform                 │         │
  └─────────────────────────────────────────────────────────┘          │
          │                                                            │
          ▼                                                            │
  AudioRolloutOutput { codec_tokens, logprobs, waveform, sample_rate } │
          │                                                            │
          ▼                                                            │
  AutoRegressiveTTSAgentLoopWorker  ── waveforms (n samples) ─────────►┤
          │                                                            │
          │◄── transcripts ────────────────────────────────────────────┘
          │
          ▼
  AsrErrorRateRewardManager
[...]

Issue #388, opened 2026-08-14
[RFC][tracking] Fail Fast — No Silent Fallbacks


zhtmike:
## Motivation

Silent fallbacks hide misconfigurations and corrupt data behind "reasonable" defaults. Because RL training only reveals such bugs as abnormal metric curves hours into a run, they are the worst failure mode in this codebase. This RFC proposes the opposite policy: **misconfigurations and corrupt data raise immediately with actionable errors.** Fallbacks remain only for legitimate cases: optional dependencies when the user left a value unset, explicitly declared optional features, and transient stochastic anomalies — which get visible counters and abort thresholds instead of silence.

The policy follows upstream verl's fail-fast design (Section *Reference*), and verl-omni already applies it in several places — this RFC makes it consistent everywhere. Critically, some things that *look* like silent fallbacks are **intended behavior** — exact parity with verl/vllm-omni contracts, dead defensive branches, or guarded-by-producer invariants. Mixing those into the fix list produces wrong plans, so every candidate found in the 2026-09 sweep was adversarially verified against the exact pins with **verl/vllm as the standard**: verl-omni behavior is then a bug or an intended deviation. Blessed fallbacks and false positives live in Part III.

Current tally: **19 original items** — 14 landed or removed outright (#429, #443, #530, plus #359/#315 for two), 1 in review (**A3** via #539), 3 open (**B2, B4, B5**), 1 partial (**A12**, plus an A7 print residue) · **18 sweep candidates from the Aug–Sep pipeline wave** — 5 confirmed (**C1, C5, C6, C8, C9**), 7 minor/hardening, 6 false positives removed to the Part III record · 4 infrastructure checkboxes — **36 boxes total**.

**Audit basis (2026-09-07, verl-omni @ `03655dc4`):**

| Repo | Commit audited | Role |
|---|---|---|
| verl-omni | `03655dc4` (main) | subject |
| verl | `fefb0802` (exact pin, `pyproject.toml`) | standard |
| vllm-omni | `ded8934` (exact pin, `.github/vllm_omni_pin.txt`) | standard / engine contract source |
| vllm | `3dc7a68ce4` (recent) | standard (LoRA strictness) |
| diffusers | `c1bf18c92` (recent) | standard (config loading) |

## Policy

1. **Explicit user intent is sacred.** If a config value was explicitly set and cannot be honored, raise — explain what was requested, why it fails, and what to change.
[...]

Issue #377, opened 2026-08-11
[RFC][tracking] Refactor `vllm_omni_async_server.py` — Align with Upstream verl, Remove Temp Patches


zhtmike:
## Motivation

`verl_omni/workers/rollout/vllm_rollout/vllm_omni_async_server.py` drifted from upstream verl's `vLLMHttpServer` (parent: verl `vllm_async_server.py`) across many commits, reimplementing sleep/wake/abort/drain with different semantics, interleaving AR and diffusion paths, and accumulating untested temp patches. This RFC tracks the convergence.

**Status (2026-09-07 update):** the original Phase 1 is largely landed — strategy split in [#444](https://github.com/verl-project/verl-omni/pull/444) (2026-08-28), level-1 sleep guard in [#460](https://github.com/verl-project/verl-omni/pull/460), lifecycle redesign in [#497](https://github.com/verl-project/verl-omni/pull/497) (2026-09-03, [BREAKING] vllm-omni pin → 0.28 `ded8934`). The server shrank **846 → 488 lines**, contains **zero** `if self._ar_mode` branches, and the pin carries the upstream fixes this RFC was designed around ([vllm-omni#6084](https://github.com/vllm-project/vllm-omni/pull/6084), [vllm-omni#6367](https://github.com/vllm-project/vllm-omni/pull/6367); [#6083](https://github.com/vllm-project/vllm-omni/issues/6083) closed). This issue now tracks the **remainder** plus newly discovered upstream defects. Ownership: generation strategies → [#403](https://github.com/verl-project/verl-omni/issues/403); trainers → [#491](https://github.com/verl-project/verl-omni/issues/491); patch debt → [#445](https://github.com/verl-project/verl-omni/issues/445); this RFC owns **server lifecycle**.

Current tally: **19 tracker items** (5 landed, 7 remaining R1 no-upstream-dependency, 5 gated upstream R2, 2 parity R3) + 3 upstream proposals to file (F1–F3; F4 optional structural) · **deviation ledger complete** (Part I: 11 `run_server` blocks + an 11-row lifecycle ledger covering 14 methods, adjudicated against verl pin / vllm-omni pin / vllm-omni's own api_server pattern / vLLM; 8-item alignment shortlist; 6 unverified residuals) · **12 upstream items** in 3 clusters (Part II) · 4 coordination items. 3 of the original 5 open questions answered; 2 carried + 3 new.

Audit basis (2026-09-07):

| Repo | Audited | verl-omni pin |
[...]

PR #428, opened 2026-08-23
[3/N][model, rollout, reward] feat: add Qwen3-TTS GRPO support

dongbo910220:
### What does this PR do?

Adds full-parameter stock-GRPO training support for the codec-0 policy of `Qwen/Qwen3-TTS-12Hz-0.6B-Base` through the existing V1 `verl_omni.trainer.main_omni` path.

This `[3/N]` PR adds a model-specific Qwen3-TTS Talker training adapter, a two-stage rollout adapter, a full-parameter GRPO recipe, a two-GPU smoke test, focused CPU contract tests, and documentation.

The recipe keeps the actor/reference master parameters and optimizer state in FP32, uses BF16 for FSDP forward/backward computation and rollout, keeps gradient reduction and buffers in FP32, and uses LR `1e-6` with 10 warmup steps and optional direct reference-model KL.

It does not add a new Trainer, algorithm, policy loss, dataset format, ASR gate, candidate mask, or SpeechJudge-specific code path. This model-specific slice reuses existing shared rollout and reward utilities rather than duplicating them.

Related RFC: #90. 

### Test

```bash
CUDA_VISIBLE_DEVICES='' pytest -q \
  tests/pipelines/test_qwen3_tts_on_cpu.py \
  tests/pipelines/test_qwen3_tts_package_on_cpu.py \
  tests/workers/rollout/rollout_vllm/test_qwen3_tts_rollout_on_cpu.py
```

Result: `all passed`.

The unchanged runtime implementation completed a two-GPU, two-step Qwen3-TTS Talker GRPO smoke on two RTX 5090 GPUs. Final-head GitHub GPU CI is still required before merge.

We also completed 200 Qwen3-TTS GRPO updates and evaluated the same fixed 100-sample validation set at step 0 and every 20 steps.

SpeechJudge reward curves over 200 training steps

Validation SpeechJudge mean increased from `15.5760` to `19.0116`, with a paired mean gain of `+3.4356` (95% CI `[+2.1739, +4.7301]`) and a paired median gain of `+2.1563`.

### Design & Code Changes

- **Actor replay:** reconstructs Qwen3-TTS's dual text/codec layout, speaker embedding, and all 16 codebooks while exposing codec-0 as the policy sequence and freezing non-policy modules.
- **Rollout topology:** uses the shared Talker hooks to retain the codec trajectory and decoded waveform, replay model-specific payloads, and synchronize actor weights only to the trainable Talker stage.
- **Code2Wav handoff:** derives the decoder placeholder shape from the retained codec-0 trajectory while preserving actual codec values through the worker connector; missing or empty completions fail closed.
- **Dependencies and smoke:** installs `qwen-tts` through the `omni` optional dependency group without a repository-specific source pin and selects the Qwen3-TTS two-GPU smoke through the existing GPU-smoke workflow.
AI assistance was used;

Co-authored-by: @Michael-Zzq


zhtmike:
@knlnguyen1802 can you take a look

verl_omni/workers/rollout/vllm_rollout/vllm_omni_async_server.py
zhtmike, line 1: consider to check https://github.com/verl-project/verl-omni/issues/377 to avoid huge code modification

knlnguyen1802 (commented):
Please update README.md

verl_omni/pipelines/qwen3_tts/agent_loop.py
```diff
@@ -0,0 +1,61 @@
...
+import torch
+from verl.experimental.agent_loop.agent_loop import register
+from verl.experimental.agent_loop.single_turn_agent_loop import SingleTurnAgentLoop
+
+from verl_omni.pipelines.qwen3_tts.rollout_utils import is_evaluation_split, with_rollout_generation_seed
+from verl_omni.pipelines.qwen3_tts.talker_forward import TEXT_PROMPT_TRAILER_TOKENS, build_assistant_text
+
+
+@register("qwen3_tts_single_turn")
+class Qwen3TTSSingleTurnAgentLoop(SingleTurnAgentLoop):
```
knlnguyen1802: This is too dependency, a agent loop should me general, you can follow standard of the repo to create Omni version or Diffusion version of `SingleTurnAgentLoop`

verl_omni/workers/rollout/vllm_rollout/vllm_omni_async_server.py
dongbo910220: Thanks. I reviewed #377 and the ongoing strategy split in draft PR #444. I kept Qwen3-TTS-specific prompt construction, stage sampling, and output/policy mapping in Qwen3TTSRolloutAdapter; the shared server changes are limited to generic multi-stage adapter plumbing. I did not duplicate the strategy refactor while #444 is in progress. Once #444 lands, I will rebase this PR onto the updated main and migrate the remaining generic AR preprocessing, generation, and output hooks into _ARGenerateStrategy.

dongbo910220:
> Please update README.md

updated

verl_omni/pipelines/qwen3_tts/agent_loop.py
dongbo910220: Updated. I removed the Qwen3-TTS-specific agent loop and added a generic OmniSingleTurnAgentLoop that reuses the repository's standard SingleTurnAgentLoop. Qwen3-TTS-specific sampling preparation and policy-output mapping now live in Qwen3TTSRolloutAdapter.

zhtmike:
before review, may I ask how grpo is applied exactly to TTS output, any paper for reference, and what is 16-codebook teacher-forced inputs?

A detail background can be provided so we dont provide wrong review comment

zhtmike:
And If I understand correctly the idea is partially borrowed from #282 ? correct me if I am wrong~

dongbo910220:
> before review, may I ask how grpo is applied exactly to TTS output, any paper for reference, and what is 16-codebook teacher-forced inputs?
> 
> A detail background can be provided so we dont provide wrong review comment

Thanks for asking.

1. **How GRPO is applied to TTS:** for each text prompt, the rollout policy samples a group of codec-token trajectories. Each trajectory is decoded into a waveform by `code2wav` and scored by an external audio reward model. The scores are converted into group-relative advantages, and the actor replays the sampled trajectories to compute codec-0 selected-token log-probabilities for the standard GRPO update. A closely related reference is [Group Relative Policy Optimization for Text-to-Speech with Large Language Models](https://arxiv.org/abs/2509.18798).

2. **What “16-codebook teacher-forced inputs” means:** Qwen3-TTS-12Hz produces 16 codec tokens per audio frame. Codec-0 is predicted by the Talker backbone, while codec-1 to codec-15 provide residual acoustic details. During training, the complete 16-codebook trajectory generated during rollout is fed back to the actor as fixed history so that it can recompute the probability of the sampled codec-0 tokens. “Teacher-forced” here means replaying rollout-generated tokens, not using ground-truth speech tokens. The architecture is described in the [Qwen3-TTS Technical Report](https://arxiv.org/abs/2601.15621). A close multi-codebook precedent for treating only the first codebook as the policy sequence is [SpeechAlign](https://arxiv.org/abs/2404.05600), which applies RL to the AR model generating the first of eight RVQ codebooks while a pretrained NAR model supplies the remaining seven.

dongbo910220:
> And If I understand correctly the idea is partially borrowed from #282 ? correct me if I am wrong~

The two PRs overlap at the high-level goal of Qwen3-TTS post-training, but their scopes and integration paths are different. The main methodological reference for this PR was [Group Relative Policy Optimization for Text-to-Speech with Large Language Models](https://arxiv.org/abs/2509.18798), as cited in the preceding reply.

- [#282](https://github.com/verl-project/verl-omni/pull/282) provides GSPO/GDPO and Online DPO recipes, while this PR focuses on stock GRPO.
- [#282](https://github.com/verl-project/verl-omni/pull/282) adds a TTS-specific training entry and Online-DPO loss path, while this PR reuses the existing V1 `main_omni` path and standard GRPO/vanilla PPO policy loss.
- [#282](https://github.com/verl-project/verl-omni/pull/282) includes several task-specific audio reward implementations in the repository, while this PR exposes a generic HTTP interface for pointwise scalar audio rewards.
- [#282](https://github.com/verl-project/verl-omni/pull/282) packages standalone TTS recipes and dependency patches, while this PR adds generic omni rollout hooks and a two-stage vLLM-Omni pipeline.

Therefore, this PR is intentionally narrower and focuses on integrating Qwen3-TTS into the existing generic omni GRPO path.

zhtmike:
> > before review, may I ask how grpo is applied exactly to TTS output, any paper for reference, and what is 16-codebook teacher-forced inputs?
> > A detail background can be provided so we dont provide wrong review comment
> 
> Thanks for asking.
> 
> 1. **How GRPO is applied to TTS:** for each text prompt, the rollout policy samples a group of codec-token trajectories. Each trajectory is decoded into a waveform by `code2wav` and scored by an external audio reward model. The scores are converted into group-relative advantages, and the actor replays the sampled trajectories to compute codec-0 selected-token log-probabilities for the standard GRPO update. A closely related reference is [Group Relative Policy Optimization for Text-to-Speech with Large Language Models](https://arxiv.org/abs/2509.18798).
> 2. **What “16-codebook teacher-forced inputs” means:** Qwen3-TTS-12Hz produces 16 codec tokens per audio frame. Codec-0 is predicted by the Talker backbone, while codec-1 to codec-15 provide residual acoustic details. During training, the complete 16-codebook trajectory generated during rollout is fed back to the actor as fixed history so that it can recompute the probability of the sampled codec-0 tokens. “Teacher-forced” here means replaying rollout-generated tokens, not using ground-truth speech tokens. The architecture is described in the [Qwen3-TTS Technical Report](https://arxiv.org/abs/2601.15621). A close multi-codebook precedent for treating only the first codebook as the policy sequence is [SpeechAlign](https://arxiv.org/abs/2404.05600), which applies RL to the AR model generating the first of eight RVQ codebooks while a pretrained NAR model supplies the remaining seven.

Thanks for your explanation.
Please reflecting these references in the TTS grpo readme. To make sure the algo implemented here is paper supported.

verl_omni/workers/rollout/vllm_rollout/vllm_omni_async_server.py
zhtmike: ping @knlnguyen1802 @NancyFyong, should we land this PR before #444

NancyFyong: I still have a few tests to run; I expect to complete the merge today.

dongbo910220:
> Thanks for your explanation. Please reflecting these references in the TTS grpo readme. To make sure the algo implemented here is paper supported.

References updated. Thanks!

docs/start/http_scorer.md
zhtmike: need update date

examples/grpo_trainer/qwen3_tts/README.md
```diff
@@ -0,0 +1,161 @@
...
+decoding. Here, "teacher-forced" means that the actor replays the tokens sampled
+during rollout as fixed history; it does not mean that ground-truth speech
+tokens are used.
+
+[SpeechAlign](https://arxiv.org/abs/2404.05600) is a related multi-codebook
+speech-alignment precedent: its autoregressive model generates the first of
+eight RVQ codebooks and a pretrained non-autoregressive model supplies the
+remaining layers. It supports treating the first codebook as the optimized
+sequence while preserving residual acoustic codebooks, but it uses preference
+optimization rather than the GRPO objective implemented here.
```
zhtmike: I think it is a bit too detail, interested reader will read the reference if the algorithm is following the paper

examples/grpo_trainer/qwen3_tts/run_qwen3_tts_grpo.sh
zhtmike, line 1: I think we need a e2e smoke test for this, since it is a talker implementation, we should guard this type

examples/grpo_trainer/qwen3_tts/README.md
```diff
@@ -0,0 +1,161 @@
...
+The client retries only transient network, timeout, HTTP 408/429, and 5xx
+failures. Missing, malformed, or non-finite results stop the run instead of
+being converted to a valid zero reward.
+
+For SpeechJudge-BTRM, deploy the official
+[`AmphionTeam/SpeechJudge`](https://github.com/AmphionTeam/SpeechJudge) code and
+[`RMSnow/SpeechJudge-BTRM`](https://huggingface.co/RMSnow/SpeechJudge-BTRM)
+checkpoint in a separate environment, then expose its pointwise score through
+this protocol. Pin the SpeechJudge source revision and runtime versions in the
+service deployment. SpeechJudge-BTRM is licensed CC-BY-NC-4.0.
```
zhtmike: SpeechJudge is supported by vllm natively, https://github.com/AmphionTeam/SpeechJudge#inference-with-vllm. Can we use the native vllm reward server for this?

verl_omni/utils/reward_score/audio_http_scorer_client.py
zhtmike, line 1: if we use vllm as SpeechJudge, we don't need this file then

verl_omni/workers/engine/fsdp/omni_impl.py
```diff
@@ -43,6 +43,43 @@
...
+        finally:
+            self.engine_config.forward_only = True
+
+    def _build_fsdp_module(self, module):
+        parent_build = super()._build_fsdp_module
+        return self._run_with_manual_ref_offload(lambda: parent_build(module))
+
+    def to(self, device, model=True, optimizer=True, grad=True):
+        parent_to = super().to
+        return self._run_with_manual_ref_offload(lambda: parent_to(device, model=model, optimizer=optimizer, grad=grad))
```
zhtmike: what happens here, it is really necessary?

verl_omni/workers/engine/fsdp/omni_impl.py
```diff
@@ -192,6 +228,7 @@ def _build_module(self):
                 self.model_config.model_stage,
                 self.model_config.get("external_lib"),
             )
+            self.model_adapter_cls = adapter_cls
```
zhtmike: seems irrelevant change to your pr

verl_omni/workers/rollout/vllm_rollout/vllm_omni_async_server.py
zhtmike: I will postpone this part review after #444 lands

verl_omni/pipelines/model_base.py
```diff
@@ -724,6 +729,16 @@ def rollout_flags(cls, pipeline_mode="thinker_only") -> dict[int, dict]:
         """
         return {}
 
+    @classmethod
+    def weight_sync_stage_ids(cls, pipeline_mode="thinker_only") -> list[int] | None:
+        """Return stages that receive actor weights, or all stages by default."""
+        return None
+
+    @classmethod
+    def supports_cache_engine_sleep(cls, pipeline_mode="thinker_only") -> bool:
```
zhtmike: what is this hook suppose to do?

verl_omni/pipelines/model_base.py
```diff
@@ -778,3 +793,41 @@ def get_stage_engine_extras(cls, stage_id: int, pipeline_mode: str = "thinker_on
...
+
+    @classmethod
+    def postprocess_agent_loop_output(cls, output, *, tokenizer, response_length: int):
+        """Map an engine trajectory to the policy sequence used by the actor."""
+        return output
+
+    @classmethod
+    def combine_engine_outputs(cls, outputs: list, prompt: dict) -> tuple[Any, dict[str, Any]]:
+        """Select the policy output and collect architecture-specific fields."""
+        return (outputs[-1] if outputs else None), {}
```
zhtmike: consider minimize the hook introduced here. 

We add hook if it is really necessary.

verl_omni/pipelines/qwen3_tts/transformers_compat.py
```diff
@@ -0,0 +1,81 @@
...
+    """Expose the TF5 APIs expected while qwen-tts binds its imports.
+
+    qwen-tts modules retain the mask wrappers they import. The global
+    Transformers functions are restored immediately afterwards.
+    """
+    import transformers
+
+    if Version(transformers.__version__).major < 5:
+        yield
+        return
```
zhtmike: we don't support transformer < 5, drop fallback or compatibility change.

verl_omni/pipelines/qwen3_tts/transformers_compat.py
```diff
@@ -0,0 +1,81 @@
...
+
+    return wrapper
+
+
+def patch_qwen3_tts_config_defaults(config_cls) -> None:
+    """Restore the config default that qwen-tts expects from Transformers 4.x."""
+    talker_config_cls = getattr(config_cls, "sub_configs", {}).get("talker_config")
+    if talker_config_cls is not None and not hasattr(talker_config_cls, "pad_token_id"):
+        talker_config_cls.pad_token_id = None
+
```
zhtmike: move this patch to the necessary hooks, check qwen3-omni implementation.

verl_omni/pipelines/qwen3_tts/talker_training_adapter.py
```diff
@@ -0,0 +1,232 @@
...
+
+    patch_qwen3_tts_config_defaults(config_cls)
+    try:
+        AutoConfig.register(getattr(config_cls, "model_type", "qwen3_tts"), config_cls)
+    except ValueError:
+        pass
+    try:
+        AutoModelForMultimodalLM.register(config_cls, model_cls)
+    except ValueError:
+        pass
```
zhtmike: Drop the patch, fix it formally, if it is not from AutoModelForMultimodalLM, pick the right AutoModel class and read it insider the omni fsdp engine

verl_omni/pipelines/qwen3_tts/omni_rollout_adapter.py
```diff
@@ -0,0 +1,253 @@
...
+    completions = getattr(request_output, "outputs", None)
+    return completions[0] if completions else None
+
+
+def _materialize(value):
+    if isinstance(value, Mapping):
+        return {key: _materialize(item) for key, item in value.items()}
+    if isinstance(value, list):
+        return [_materialize(item) for item in value]
+    return value
```
zhtmike: it is not readable, what is this code doing?

docs/start/http_scorer.md
dongbo910220: Updated

examples/grpo_trainer/qwen3_tts/README.md
dongbo910220: Changed it to a shorter version.

examples/grpo_trainer/qwen3_tts/run_qwen3_tts_grpo.sh
dongbo910220: Added a two-GPU, two-step Qwen3-TTS Talker GRPO smoke E2E test

examples/grpo_trainer/qwen3_tts/README.md
dongbo910220: Thanks for pointing this out. I checked the official SpeechJudge paths. `main_grm_vllm.py` runs the pairwise generative GRM through `vllm.LLM.generate`, while the pointwise BTRM used by this recipe is implemented with `RewardModelInferencePipeline`, PEFT, and a scalar reward head. The GRM vLLM script is therefore not a drop-in pointwise reward server for GRPO candidates. I clarified this distinction in the README and kept the scorer behind the generic audio HTTP contract so either BTRM or another pointwise scorer can run in its compatible environment.

verl_omni/utils/reward_score/audio_http_scorer_client.py
dongbo910220: I kept this file because the native vLLM example is for the pairwise generative GRM, not the pointwise BTRM scalar head required here. The client remains scorer-agnostic and only defines a fail-closed waveform-to-scalar protocol; it contains no SpeechJudge-specific logic. The README now makes that boundary explicit.

verl_omni/workers/engine/fsdp/omni_impl.py
dongbo910220: You're right, this override was not necessary. Removed.

verl_omni/workers/engine/fsdp/omni_impl.py
dongbo910220: `self.model_adapter_cls` is used by the FSDP engine to reconstruct model-specific replay inputs in `prepare_model_inputs`. It is also now selected before `from_pretrained`, allowing Qwen3-TTS to provide its official model class without modifying the global AutoModel registry.

verl_omni/pipelines/model_base.py
dongbo910220: Removed

verl_omni/pipelines/model_base.py
dongbo910220: Agreed. I removed `get_output_modalities`; the server now derives retained modalities from the pipeline stage metadata. The remaining hooks cover Qwen3-TTS-specific prompt construction, dual-sampler candidate seeding, codec-0 policy replay mapping, and multi-stage output assembly, which cannot be inferred from the generic topology.

verl_omni/pipelines/qwen3_tts/transformers_compat.py
dongbo910220: Removed.

verl_omni/pipelines/qwen3_tts/transformers_compat.py
dongbo910220: Thanks. I scoped the compatibility context to the Qwen3-TTS config/model import hooks instead of leaving a persistent Transformers patch.

verl_omni/pipelines/qwen3_tts/talker_training_adapter.py
dongbo910220: Agreed. I removed the global `AutoConfig`/`AutoModelForMultimodalLM` registration. `OmniModelBase` now has optional config/model-class selectors, and `OmniFSDPEngine` owns `from_pretrained`, using the adapter-provided official Qwen3-TTS class only when required and preserving `AutoModelForMultimodalLM` as the default.

verl_omni/pipelines/qwen3_tts/omni_rollout_adapter.py
dongbo910220: Renamed it to `_copy_plain_containers` and documented its purpose: convert Ray/shared-memory `Mapping` and list shells into built-in containers required by the upstream processor's strict `dict` checks, while preserving tensor payloads.

examples/grpo_trainer/qwen3_tts/README.md
zhtmike: I see

verl_omni/utils/reward_score/audio_http_scorer_client.py
zhtmike: OK

verl_omni/workers/engine/fsdp/omni_impl.py
zhtmike: Good

tests/special_e2e/qwen3_tts_smoke_scorer.py
zhtmike, line 1: For e2e smoke test, better to  use a cpu dummy reward for this. Otherwise the test is too heavy. Consider change to cpu test for this, and drop this

We will add integration test later, but not due to lack of resources, we only have smoke test

verl_omni/pipelines/qwen3_tts/omni_rollout_adapter.py
zhtmike: I think if it is not this PR's scope. 

If we need a general fix for this, pleas open a new pr for `Copy Ray/shared-memory mapping containers without copying tensor payloads. `

verl_omni/pipelines/qwen3_tts/omni_rollout_adapter.py
```diff
@@ -0,0 +1,249 @@
...
+
+@lru_cache(maxsize=4)
+def _load_speaker_vector(path: str) -> list[float]:
+    return load_speaker_xvector(path).reshape(-1).tolist()
+
+
+def _completion(output):
+    request_output = getattr(output, "request_output", None) or output
+    completions = getattr(request_output, "outputs", None)
+    return completions[0] if completions else None
```
zhtmike: make it inline, drop fallback change according to https://github.com/verl-project/verl-omni/issues/388

verl_omni/pipelines/qwen3_tts/omni_rollout_adapter.py
```diff
@@ -0,0 +1,249 @@
...
+)
+from verl_omni.pipelines.qwen3_tts.talker_forward import (
+    TEXT_PROMPT_TRAILER_TOKENS,
+    build_assistant_text,
+    load_speaker_xvector,
+    require_auto_language,
+)
+
+_PIPELINE_ID = "qwen3_tts_rl"
+_SYNC_PROCESSOR = "verl_omni.pipelines.qwen3_tts.omni_rollout_adapter.talker2code2wav_token_only"
```
zhtmike: make it inline since `_SYNC_PROCESSOR` only appeared once

verl_omni/pipelines/qwen3_tts/rollout_utils.py
```diff
@@ -0,0 +1,147 @@
...
+# limitations under the License.
+"""Helpers for collecting Qwen3-TTS rollout outputs."""
+
+import hashlib
+from collections.abc import Mapping
+from typing import Any
+
+import torch
+
+_EVALUATION_SPLITS = {"gate", "official_test", "test", "val", "validation"}
```
zhtmike: make it inline

verl_omni/pipelines/qwen3_tts/rollout_utils.py
```diff
@@ -0,0 +1,147 @@
...
+    candidate = int(_scalar(session_id) or 0)
+    explicit_seed = validation_generation_seed(extra_info)
+    if explicit_seed is not None:
+        return (explicit_seed + candidate) % _MAX_SAMPLING_SEED
+
+    info = _extra_info_mapping(extra_info)
+    sample_id = _scalar(info.get("id", info.get("index", uid)))
+    step = "evaluation" if is_evaluation_split(info) else _scalar(global_steps)
+    payload = "\0".join(map(str, (int(base_seed), step, sample_id, candidate))).encode()
+    return int.from_bytes(hashlib.blake2b(payload, digest_size=8).digest(), "big") % _MAX_SAMPLING_SEED
```
zhtmike: we do need a seed for omni pipeline. But consider make it general. So consider drop this, add a global seed control for omni pipeline in new PR

verl_omni/pipelines/qwen3_tts/rollout_utils.py
```diff
@@ -0,0 +1,147 @@
...
+
+
+def with_rollout_generation_seed(sampling_params, extra_info, **seed_kwargs):
+    """Seed codec-0 and residual-codebook sampling without mutating input."""
+    seed = rollout_generation_seed(extra_info, **seed_kwargs)
+    seeded = dict(sampling_params)
+    residual_args = dict(seeded.get("extra_args") or {})
+    residual_args["tts_local_seed"] = seed
+    seeded.update(seed=seed, extra_args=residual_args)
+    return seeded
```
zhtmike: same

verl_omni/pipelines/qwen3_tts/rollout_utils.py
```diff
@@ -0,0 +1,147 @@
...
+        return accumulated
+    if accumulated is None:
+        return chunk
+    if chunk.shape[1:] != accumulated.shape[1:]:
+        raise RuntimeError(f"Rollout chunks changed shape from {tuple(accumulated.shape)} to {tuple(chunk.shape)}.")
+    if chunk.shape[0] >= accumulated.shape[0] and torch.equal(chunk[: accumulated.shape[0]], accumulated):
+        return chunk
+    if accumulated.shape[0] >= chunk.shape[0] and torch.equal(accumulated[: chunk.shape[0]], chunk):
+        return accumulated
+    return torch.cat((accumulated, chunk), dim=0)
```
zhtmike: non-readable. Fix it. Drop compatible and fallback change according to #388

verl_omni/pipelines/qwen3_tts/talker_forward.py
```diff
@@ -0,0 +1,242 @@
...
+class TalkerBatch:
+    input_ids: torch.Tensor
+    codec_ids: torch.Tensor
+    text_embedding_mask: torch.Tensor
+    codec_embedding_mask: torch.Tensor
+    codec_mask: torch.Tensor
+    attention_mask: torch.Tensor
+    codec_lens: list[int]
+    logit_start: list[int]
+    speaker_slots: list[int]
```
zhtmike, line 73: consider make it general. Not only Qwen3-TTS needs this, Qwen3-Omni Talker (or future model) also needs this. If this PR is too huge, consider design the interface first with a RFC, we add the interface as the first PR, this PR is the second.

verl_omni/pipelines/qwen3_tts/talker_training_adapter.py
```diff
@@ -0,0 +1,209 @@
...
+    require_auto_language,
+    tts_actor_logits,
+)
+from verl_omni.pipelines.qwen3_tts.transformers_compat import (
+    patch_qwen3_tts_config_defaults,
+    qwen3_tts_import_context,
+)
+
+logger = logging.getLogger(__name__)
+_PASSTHROUGH_TEMPLATE = "{% for message in messages %}{{ message['content'] }}{% endfor %}"
```
zhtmike: make inline all global variables if it is using only once.

verl_omni/pipelines/qwen3_tts/transformers_compat.py
```diff
@@ -0,0 +1,76 @@
...
+    def wrapper(*args, input_embeds=None, inputs_embeds=None, cache_position=None, **kwargs):
+        del cache_position
+        embeddings = inputs_embeds if inputs_embeds is not None else input_embeds
+        return original(*args, inputs_embeds=embeddings, **kwargs)
+
+    return wrapper
+
+
+def patch_qwen3_tts_config_defaults(config_cls) -> None:
+    """Restore the config default that qwen-tts expects from Transformers 4.x."""
```
zhtmike: As I said, we don't support Transformers 4.x., but you still add compaitbility code

verl_omni/workers/engine/fsdp/omni_impl.py
```diff
@@ -180,18 +198,13 @@ def _build_module(self):
         with init_context(), warnings.catch_warnings():
             warnings.simplefilter("ignore")
 
-            module = AutoModelForMultimodalLM.from_pretrained(
+            model_cls = adapter_cls.get_model_class() or AutoModelForMultimodalLM
+            module = model_cls.from_pretrained(
```
zhtmike: simply use `AutoModelForTextToWaveform.from_pretrained`, thanks!

zhtmike:
Ask you AI to bump the testing environment to Transformer 5.X, we don't support transformer 4.x here, so no need compatibility code, it it non-readable and hacky.

verl_omni/agent_loop/single_turn_agent_loop.py
```diff
@@ -24,9 +22,6 @@
 from verl_omni.agent_loop.diffusion_agent_loop import DiffusionAgentLoopOutput
 from verl_omni.pipelines.model_base import OmniRolloutPipelineBase
 
-logger = logging.getLogger(__file__)
-logger.setLevel(os.getenv("VERL_LOGGING_LEVEL", "WARN"))
```
knlnguyen1802, line 28: Why remove this one

tests/special_e2e/qwen3_tts_smoke_scorer.py
dongbo910220: Replaced the external smoke scorer with an in-process CPU dummy reward.

verl_omni/workers/rollout/vllm_rollout/vllm_omni_ar_strategy.py
```diff
@@ -67,29 +74,31 @@ def override_generation_config(self) -> dict[str, Any]:
         return vLLMHttpServer._get_override_generation_config(self.server)
 
     def worker_extension_cls(self, device_type: str) -> str:
-        return _WORKER_EXTENSION
+        return "verl_omni.workers.rollout.vllm_rollout.utils.vLLMOmniColocateWorkerExtension"
```
knlnguyen1802: Please revert this

verl_omni/workers/rollout/vllm_rollout/vllm_omni_ar_strategy.py
```diff
@@ -67,29 +74,31 @@ def override_generation_config(self) -> dict[str, Any]:
...
                 hf_overrides = engine_kwargs.get("hf_overrides", {})
                 if isinstance(hf_overrides, str):
                     hf_overrides = json.loads(hf_overrides)
                 hf_overrides.update(adapter_overrides)
                 engine_kwargs["hf_overrides"] = hf_overrides
 
-        stage_init_timeout = engine_kwargs.get("stage_init_timeout") or engine_kwargs.get("stage-init-timeout")
-        init_timeout = engine_kwargs.get("init_timeout") or engine_kwargs.get("init-timeout")
+        stage_init_timeout = engine_kwargs.get("stage_init_timeout")
+        init_timeout = engine_kwargs.get("init_timeout")
```
knlnguyen1802: Please check this again and confirm no regression

verl_omni/pipelines/qwen3_tts/omni_rollout_adapter.py
dongbo910220: Removed.

verl_omni/pipelines/qwen3_tts/omni_rollout_adapter.py
dongbo910220: Changed.

verl_omni/pipelines/qwen3_tts/omni_rollout_adapter.py
dongbo910220: Removed.

verl_omni/pipelines/qwen3_tts/rollout_utils.py
dongbo910220: Removed.

verl_omni/pipelines/qwen3_tts/rollout_utils.py
dongbo910220: Removed.

verl_omni/pipelines/qwen3_tts/rollout_utils.py
dongbo910220: Removed.

verl_omni/pipelines/qwen3_tts/rollout_utils.py
dongbo910220: Removed. The pinned engine returns final stage payloads, so the adapter now reads those outputs directly and fails closed on contract mismatches.

verl_omni/pipelines/qwen3_tts/talker_forward.py
dongbo910220: Merged PR #504. Qwen3-TTS now uses the shared Talker rollout and actor-input hooks.

verl_omni/pipelines/qwen3_tts/talker_training_adapter.py
dongbo910220: Inlined the single-use module globals.

verl_omni/pipelines/qwen3_tts/transformers_compat.py
dongbo910220: Removed.

verl_omni/workers/engine/fsdp/omni_impl.py
dongbo910220: Changed.

dongbo910220:
>  bump the testing environment to Transformer 5.X, we don't support transformer 4.x here, so no need compatibility code, it it non-readable and hacky.

Done. Removed the compatibility module.

verl_omni/agent_loop/single_turn_agent_loop.py
dongbo910220: Restored.

verl_omni/workers/rollout/vllm_rollout/vllm_omni_ar_strategy.py
dongbo910220: Reverted.

verl_omni/workers/rollout/vllm_rollout/vllm_omni_ar_strategy.py
dongbo910220: Restored.

verl_omni/pipelines/qwen3_tts/talker_training_adapter.py
```diff
@@ -0,0 +1,179 @@
...
+        from qwen_tts.core.models.configuration_qwen3_tts import Qwen3TTSConfig
+        from qwen_tts.core.models.modeling_qwen3_tts import Qwen3TTSForConditionalGeneration
+        from transformers import AutoConfig, AutoModelForTextToWaveform
+
+        AutoConfig.register("qwen3_tts", Qwen3TTSConfig, exist_ok=True)
+        AutoModelForTextToWaveform.register(
+            Qwen3TTSConfig,
+            Qwen3TTSForConditionalGeneration,
+            exist_ok=True,
+        )
```
zhtmike, line 91: I see why you need to register it here. So it is not in `transfomers`. Consider to watch https://github.com/huggingface/transformers/pull/44517 closely

verl_omni/workers/rollout/vllm_rollout/vllm_omni_ar_strategy.py
zhtmike, line 1: it has huge change in this file, might break current omni pipeline. Need to take care

zhtmike:
@AndyZhou952 pls take a review thx

verl_omni/pipelines/qwen3_tts/talker_training_adapter.py
dongbo910220: Thanks. We'll track huggingface/transformers#44517 and remove the local registration when available.

verl_omni/workers/rollout/vllm_rollout/vllm_omni_ar_strategy.py
dongbo910220: Yes, checked again and added a regression test for the existing Qwen3-Omni thinker-only path.

AndyZhou952:
**Verdict**

  Not mergeable. The idea is right and the codec-0 replay math is correct, but the PR fails gate zero on evidence, CI, reproducibility, and scope. Every finding below is new relative to zhtmike's and knlnguyen1802's threads. All of the author's "Removed/Restored" replies check out except one: the global AutoConfig registration was not removed, only
  moved into a hook that now runs for every omni model.

  **Blockers**

  1. The committed recipe cannot learn. The actor is materialized in bf16 with no fp32 master copy, and plain AdamW runs on those bf16 params at lr 2e-7. That step is below half a bf16 ULP for any weight above about 6e-5 in magnitude, so over 99% of parameters round back to their old value every step. The recipe was fp32 until commit 0598b1a switched it to bf16 mid-PR. The 500-step validation gain is too consistent to be noise, so it was almost certainly produced by the fp32 recipe that is no longer in the PR. The description never says which config produced the plot.

  2. Silent regression on Qwen3-Omni pipeline_mode=full. The rewritten AR strategy enters its multi-output branch whenever the upstream pipeline has two or more final_output stages, and the base combiner returns the last one. Qwen3-Omni's full pipeline has two such stages, so the policy output becomes the code2wav stage, which carries no token ids. Before the PR this config crashed loudly on a sampling-params length check. After it, training proceeds on empty responses. The thinker-only path is behaviour-identical on every axis we traced, so this is the concrete case behind zhtmike's warning.
  3. Dependency stack is not reproducible. The repo caps transformers at 5.14.1 while the pinned qwen-tts commit requires 5.15.1 or newer. The README and smoke script hide this  with --no-deps. The git pin is not in the tts extra or any pin file, and torchaudio is unpinned against vllm's torch.
  4. The GPU smoke is a vacuous gate. Its success check is grep -q "training/global_step.*2", which matches the step-1 log line. It also runs uv pip install inside the shared CI job, downloads the real 0.6B checkpoint instead of a tiny-random build, hard-requires exactly two GPUs, and its file names match no group pattern, so the selector falls open to all three groups.
  5. Process. Head is CONFLICTING, so re-labeling cannot start CI. Open PR #282 adds the same reward_manager/audio.py path. The description lists a deleted test file, claims an unreproducible test count, and lacks the AI-assistance and non-duplication statements CLAUDE.md requires.

**Notable mediums**

  - async_chunk defaults to true upstream and the Qwen3-TTS contract only holds when it is false. The example passes it, but nothing enforces it.
  - Four new base-class hooks each have exactly one implementer, after zhtmike asked to minimize hooks.
  - The training adapter hard-requires use_remove_padding=false and FSDP1 use_orig_params=true but asserts neither. Failures land as unrelated errors.
  - The AutoModel class is chosen by the string model_stage == "talker", which breaks a future Qwen3-Omni talker adapter.
  - The audio reward manager fabricates global_steps=0 and forwards it to the scorer, has a dead mono-downmix branch that averages a (T, 2) array over time, and reads only the streamed batch layout. The HTTP client is a third copy of the sibling clients with drifted kwarg names swallowed by **kwargs.
  - The claimed thinker-only regression test never calls the deploy-config writer it is supposed to protect. Most new tests assert mocks rather than the embedding construction or the freezing surface.

  **Verification caveat**

  The local box has no torch or vllm environment and the disk is too full to build one, so the CI-replica pytest run did not happen. What did run: py_compile, ruff check and format, the torch-free pre-commit sanity scripts, and the title checker, all clean. The training reviewer built a throwaway torch-cpu venv, ran the pure-torch pipeline test file, and independently transcribed the official Qwen3-TTS prefill and decode layout against talker_forward.py. The embeddings match numerically. Upstream claims were checked against the exact CI pins of vllm-omni, verl, and qwen-tts. The prioritized to-do list in the report tags each item as this PR, follow-up, or upstream, with exact test parameterizations.


**Review generated by Claude Code.** @zhtmike

dongbo910220:
@AndyZhou952 Thanks for the detailed review. I addressed the rollout, dependency, smoke, configuration, model-loading, reward-I/O, and test-coverage findings.

- Multi-final-output pipelines now require an explicit combiner, so Qwen3-Omni `full` fails closed instead of training on code2wav output; thinker-only deploy behavior is covered.
- Qwen3-TTS enforces non-chunked rollout and `use_remove_padding=false`; FSDP1 partial freezing enforces `use_orig_params=true`.
- AutoModel selection is adapter-owned rather than inferred from the `talker` stage name.
- Audio metadata/layout/shape handling and HTTP argument validation are fail-closed.
- Qwen3-TTS now has an exact source pin, a vLLM-aligned torchaudio pin, and a pinned ~31 MB 16-codebook tiny-random smoke with exact step-2 validation and CI selection coverage.
- The tests execute the deploy writer and exercise actual tiny Qwen3-TTS embedding construction and parameter freezing. The pushed head also passed the two-step smoke on two RTX 5090 GPUs.

I kept the selected-adapter AutoConfig hook because its base implementation is a no-op and the pinned Transformers version still lacks native Qwen3-TTS support. I also kept the four model-owned hooks because they represent distinct lifecycle contracts; moving them into shared-worker Qwen3-TTS conditionals would make the generic path model-specific.

The rebase/process item remains pending, so I am not claiming that part is resolved in this update.

dongbo910220:
@AndyZhou952  I updated the recipe to keep the actor and reference master parameters and the actor's AdamW state in FP32, while using BF16 for FSDP forward/backward computation and rollout. The learning rate now warms up for 10 steps to `1e-6`.

The curve below uses this precision and learning-rate configuration with `B=4`, `G=8`, KL `0.12`, SpeechJudge-BTRM, and the same complete 100-sample validation set every 20 steps. After 200 updates, validation Reward increased from `15.5760` to `19.0116` (paired mean `+3.4356`, 95% CI `[+2.1739, +4.7301]`), while CER decreased from `4.260%` to `3.887%`.

Qwen3-TTS FP32-master BF16-compute GRPO reward curves through step 200

zhtmike:
To make this PR merge smoothly, I suggest to separate the change of the public utilities (added hooks, change on vllm-omni async server, etc) into sub-PRs first, I will review them carefully. Merging such big PR is still a bit difficult.

dongbo910220:
> To make this PR merge smoothly, I suggest to separate the change of the public utilities (added hooks, change on vllm-omni async server, etc) into sub-PRs first, I will review them carefully. Merging such big PR is still a bit difficult.

That makes sense. I propose splitting the shared changes into two focused PRs:

1. Generic Omni model and multi-stage AR adapter hooks, including the stage-selective async-server RPC path and backward-compatibility tests.
2. Generic pointwise audio HTTP reward support.

After they merge, I will update #428 to contain only the Qwen3-TTS adapter, recipe, dependencies, model-specific documentation, and tests.

zhtmike:
Yes, and pls add a rfc for the public change, as additional info.

NancyFyong (changes requested):
Reviewed at head `1b2d302`. I validated the three CPU contract files locally in an env matching this PR's pins (`vllm 0.28.0`, `vllm-omni 0.28.0rc2.dev68+gded893462` == `.github/vllm_omni_pin.txt` `ded8934`, `transformers 5.14.1`):

```
pytest tests/pipelines/test_qwen3_tts_on_cpu.py \
       tests/pipelines/test_qwen3_tts_package_on_cpu.py \
       tests/workers/rollout/rollout_vllm/test_qwen3_tts_rollout_on_cpu.py
# 36 passed, 1 skipped   (skip = optional qwen_tts package absent)
```

The implementation itself looks solid and fail-closed: the actor/rollout codec-0 trajectory is guarded frame-by-frame (`tts_actor_logits`), `align_audio_codes` treats codec-0 policy tokens as an exact invariant and rejects ambiguous spans, `mask_codec0_logits` uses `-1e4` (not `-inf`) and validates the vocab/eos bounds, and the scope is deliberately narrowed to the validated `tts_language=Auto` non-streaming layout. It also touches no generic engine/strategy files — the shared omni hooks it uses already landed on `main` (e.g. #504) — so the blast radius is contained. No blocking code-correctness issues found.

Requesting changes for one accuracy item plus a couple of tracking follow-ups:

**1. (must fix) PR description no longer matches the diff.** The "What does this PR do?" section still lists "a generic `AudioRewardManager` and fail-closed JSON HTTP client", and the Test section lists `test_qwen3_tts_transformers_compat_on_cpu.py`, `tests/reward_loop/test_audio_reward_manager_on_cpu.py`, `tests/utils/reward_score/test_audio_http_scorer_client_on_cpu.py`, and `tests/workers/test_omni_fsdp_engine_on_cpu.py`. None of those are in this three-dot diff — `verl_omni/reward_loop/reward_manager/audio.py`, `verl_omni/utils/reward_score/audio_http_scorer_client.py`, and those tests already live on `main` from an earlier slice. Please trim the description/Test section to the actual `[3/N]` scope (Qwen3-TTS pipeline + recipe + smoke + the 3 CPU test files), and note the reward manager/HTTP client as already-merged, so the Test commands are reproducible against this PR.

**2. (follow-up) Re-request review on the server-touching thread.** The `vllm_omni_async_server.py` thread was left as "postpone review until #444 lands" — #444/#504 have since merged, so this is ready for another look.

**3. (nit) Two current threads look addressed but are still open:** the recipe e2e smoke (a 2-GPU/2-step Talker GRPO smoke was added) and `talker_forward.py` generality (now via the shared Talker hooks from #504). Worth resolving or confirming.

pyproject.toml
```diff
@@ -59,6 +59,16 @@ audio = [
     "qwen-omni-utils>=0.0.9",
     "audioread",
 ]
+# Install qwen-tts separately from the pinned upstream Transformers 5 support
+# revision. The released 0.1.1 source and metadata target Transformers 4.57.3.
+tts = [
```
zhtmike: let us change `tts` to `omni`

.github/actions/gpu-smoke-prepare/action.yml
```diff
@@ -30,7 +30,15 @@ runs:
...
-        uv pip install --system --break-system-packages "transformers[mistral-common]==5.14.1"
+        # vllm-omni currently resolves accelerate 1.12, while the pinned
+        # Transformers stack and this project require accelerate >=1.14.
+        uv pip install --system --break-system-packages \
+          "transformers[mistral-common]==5.14.1" "accelerate>=1.14.0"
+        # The pinned Qwen3-TTS TF5 source declares Transformers >=5.15.1, while
+        # this repository deliberately caps Transformers at 5.14.1. Install its
+        # exact tested source without letting that metadata replace the CI stack.
+        uv pip install --system --break-system-packages --no-deps \
+          "qwen-tts @ git+https://github.com/QwenLM/Qwen3-TTS.git@$(cat .github/qwen_tts_pin.txt)"
```
zhtmike: 1. revert the change
accelerate>=1.14.0 already in .toml

2. add qwen-tts in .toml instead.

.github/workflows/gpu_smoke.yml
```diff
@@ -14,6 +14,7 @@ on:
       - "tests/special_e2e/**"
       - "pyproject.toml"
       - .github/workflows/gpu_smoke.yml
+      - .github/qwen_tts_pin.txt
```
zhtmike, line 17: drop the pin. we will not track qwen-tts pin.

zhtmike (commented):
few comments

dongbo910220:
@NancyFyong Updated the PR description and Test section to match the current `[3/N]` diff. It now scopes this PR to the Qwen3-TTS pipeline, recipe, smoke, and three CPU contract files, and identifies the shared audio reward manager and HTTP client as already merged through #528. The previous server-side changes are no longer part of the current diff; Qwen3-TTS now consumes the shared hooks merged through #504 and #527.

pyproject.toml
dongbo910220: Changed the optional dependency group from `tts` to `omni`.

.github/actions/gpu-smoke-prepare/action.yml
dongbo910220: Moved `qwen-tts` into the `omni` extra and removed the redundant CI reinstall steps. The repository's uv overrides retain Transformers 5.14.1 and Accelerate 1.14.0; dependency resolution and the Qwen3-TTS CPU contracts pass with the released package.

.github/workflows/gpu_smoke.yml
dongbo910220: Removed

zhtmike (approved):
Looks good.

Are you interested to contribute qwen3-omni talker grpo as well?

dongbo910220:
@zhtmike The final-head GPU smoke exposed a compatibility issue: the released `qwen-tts==0.1.1` still uses the Transformers 4 form of `@check_model_inputs()` and cannot run with this repository's Transformers 5.14.1 stack. The upstream Transformers 5 fix in [QwenLM/Qwen3-TTS#360](https://github.com/QwenLM/Qwen3-TTS/pull/360) has not been released and requires Transformers 5.15.1 or newer, while verl-omni currently caps Transformers at 5.14.1.

I restored the previously validated Qwen-TTS source revision on top of the current head and installed it with `--no-deps`, keeping the repository's Transformers and Accelerate versions authoritative. I also updated CPU CI to install that exact source so the real package contract is no longer skipped. The full CPU suite, pre-commit, docs, and the two-GPU/two-step Qwen3-TTS GRPO smoke all passed in pre-push validation.

dongbo910220:
@zhtmike Thanks for the suggestion. I am interested in Qwen3-Omni Talker GRPO, but I currently do not have enough GPU capacity to validate the full 30B-A3B Talker path end to end, so I would rather not commit to it before I can test it properly. I would still be happy to help with smaller, well-scoped parts that fit my hardware. In the near term, I may focus on work I can validate thoroughly, including the HPSv3 reward micro-batching optimization in #449 and cleaning up my locally reproduced Qwen3-TTS Hindi LoRA/GRPO work into a contribution.

.github/actions/gpu-smoke-prepare/action.yml
```diff
@@ -30,7 +30,14 @@ runs:
...
         uv pip install --system --break-system-packages torchcodec librosa soundfile av audioread
-        uv pip install --system --break-system-packages "transformers[mistral-common]==5.14.1"
+        # Re-apply repository constraints after vllm-omni installs its runtime
+        # dependencies, which otherwise downgrade Accelerate to 1.12.
+        uv pip install --system --break-system-packages \
+          "transformers[mistral-common]==5.14.1" "accelerate>=1.14.0"
+        # The released qwen-tts package targets Transformers 4.57.3. Install the
+        # tested upstream Transformers 5 source without changing this repo's stack.
+        uv pip install --system --break-system-packages --no-deps --reinstall \
+          "qwen-tts @ git+https://github.com/QwenLM/Qwen3-TTS.git@$(cat .github/qwen_tts_pin.txt)"
```
zhtmike, line 36: let us revert this again, add `accelerate>=1.14.0` in `override-dependencies` in project.toml
@dongbo910220

dongbo910220: OK. Added.

NancyFyong (approved):

zhtmike:
All CI green. Wonderful work!

PR #449, opened 2026-08-25
[reward, perf] feat: batch HPSv3 reward inference

dongbo910220:
### What does this PR do?

HPSv3 reward scoring previously executed one synchronous model forward per reward request under a global lock. Concurrent rollout samples therefore reached the reward worker independently but still serialized into small HPSv3 forwards.

This change converts `compute_score_hpsv3` to the repository-supported asynchronous reward-function interface and collects currently ready requests through a bounded, event-loop-local queue. Image and sampled video frames are flattened across compatible requests, grouped by checkpoint, device, and frame cap, and split into frame-bounded HPSv3 micro-batches. Model outputs are then mapped back to their original requests before applying the existing per-request frame average and reward scale.

This change also fixes frame extraction for the current uint8 `CHW`/`TCHW` rollout response contract. Canonical `TCHW` videos are sampled along their leading time dimension, while channels-last inputs and legacy direct-call layouts retain best-effort compatibility.

### Test

```bash
python -m pytest -q \
  tests/utils/reward_score/test_hpsv3_micro_batching_on_cpu.py \
  -p no:cacheprovider
```

All tests passed .

###  Performance

Measured on one NVIDIA RTX PRO 6000 with HPSv3 reward micro-batching (`max_batch_size=4`):

| Metric | main | PR | Improvement |
| --- | ---: | ---: | ---: |
| Rollout and async-reward window, `timing_s/gen` | 2.776059 s | 1.711242 s | 38.36% less time / 1.6222x |
| Full RL step, `timing_s/step` | 6.603008 s | 5.578935 s | 15.51% less time / 1.1836x |

### Design & Code Changes

- Add a bounded per-event-loop request queue and a restartable single consumer.
- Gather requests that are ready after one event-loop scheduling turn without introducing a fixed batching delay.
- Run synchronous HPSv3 preprocessing and inference in an executor so the reward worker event loop remains responsive.
- Flatten sampled image/video frames across compatible requests and split them by a configurable frame cap.
- Preserve each frame's original request index and prompt, then reconstruct and average per-request scores in submission order.
- Group frames by checkpoint, device, and frame cap so incompatible requests are never mixed.
- Cache HPSv3 inferencers by checkpoint and device rather than silently reusing the first loaded configuration.
- Isolate request-level frame extraction errors and propagate model/micro-batch failures to all affected requests.
- Settle active and queued futures when the consumer is cancelled or stops unexpectedly.
- Fix canonical `TCHW` extraction to sample the time dimension while retaining best-effort compatibility for channels-last and legacy direct callers.

Co-authored-by: @Michael-Zzq


zhtmike:
Is this PR ready? how about the memory change

dongbo910220:
Thanks @zhtmike for checking. Yes, the implementation is ready for review.

For the memory change, I measured the baseline and `max_batch_size=4` path on one NVIDIA RTX PRO 6000 with the same loaded 16.58 GB HPSv3 checkpoint. 

- 16 concurrent images: peak allocated memory increased from 16.151 GiB to 16.654 GiB, or +0.504 GiB (+3.12%).
- 8 videos with 16 sampled frames: peak allocated memory increased from 16.350 GiB to 16.654 GiB, or +0.304 GiB (+1.86%).

This is the bounded memory cost of using the default frame micro-batch size of 4. In the 12-step E2E run, sampled GPU memory did not regress (23,647 MiB on main versus 22,705 MiB on the PR), but I treat that process-level sampling only as supporting evidence, not as a claim that this PR reduces memory.

zhtmike:
Agree with AI, better to have some proof with real training.

dongbo910220:
> Agree with AI, better to have some proof with real training.

I ran a real 12-step Wan2.2 DanceGRPO training validation with HPSv3 reward micro-batching enabled (`max_batch_size=4`) on 2× NVIDIA RTX PRO 6000 GPUs.

| Item | Result |
| --- | --- |
| Global training steps | 12/12 completed |
| Generated samples per step | 16 |
| Warm-step time, steps 2–12 | 74.96 s average |
| Warm slowest per-request HPSv3 `compute_score` time | 0.328 s average |
| Observed HPSv3 micro-batch size | Up to 4 |
| Runtime failures | No OOM, traceback, or runtime error |

The run exercised the complete RL training path: Wan rollout generation, asynchronous HPSv3 reward inference, old log-prob computation, advantage estimation, actor forward/backward/update, and rollout weight synchronization. Reward, actor loss, gradient norm, and PPO KL remained finite throughout the run.

The `compute_score` value above is the slowest individual reward request within each global training step; it is not the duration of the complete RL step.

zhtmike (approved):

PR #444, opened 2026-08-25
[1/N][rollout, tests] refactor: split vLLM-Omni generation strategies

NancyFyong:
## Summary

This is part 1 of the implementation proposed in #403 and implements the AR/diffusion strategy-separation portion of #377.

- select one generation strategy during server config initialization
- move AR-specific config, engine setup, request preprocessing, engine invocation, and output conversion into `_ARGenerateStrategy`
- move the corresponding diffusion behavior into `_DiffusionGenerateStrategy`
- keep the HTTP server responsible for shared engine lifecycle, LoRA state, sleep/wake, abort/drain, and RPC delegation
- add CPU characterization coverage for strategy selection and the existing AR/diffusion contracts
- update output/deploy-config tests to exercise the production strategy implementations

The public `generate(...)` RPC, `TokenOutput`, `DiffusionOutput`, pixel `uint8` contract, and lifecycle behavior are unchanged.

## Motivation

`vllm_omni_async_server.py` previously interleaved AR and diffusion behavior across initialization, launch, generation, and output processing. That made model-specific rollout changes grow the shared server and made AR and diffusion changes harder to review independently.

This PR establishes a private strategy boundary before the typed multimodal request/output work proposed in #403. Later parts can evolve diffusion I/O behind `_DiffusionGenerateStrategy` without changing the AR path or adding more mode branches to the shared server.

## Scope and non-goals

In scope:

- behavior-preserving strategy extraction
- preservation of newly added config conversion, quantization, seed, and text-encoder TP behavior from current `main`
- characterization tests around the extracted seams

Not in scope:

- typed `MediaInput` / `MediaOutput` contracts
- changes to sleep, wake, abort, or drain behavior
- removal of legacy diffusion request/output compatibility
- model-specific behavior changes

## Duplicate-work check

This does not duplicate an existing PR. Before opening it, I reviewed #377 and #403 and searched open PRs for references to both issues and for `vLLM-Omni generation strategy` / `AR diffusion strategy`; those searches returned no matching open PR.

## Test plan

- `python -m pytest -q -o 'python_files=*_on_cpu.py' --asyncio-mode=auto tests/`
  - **801 passed, 24 warnings**
- targeted strategy/output/deploy/MiniMax contract tests
  - **37 passed, 21 warnings**
- `pre-commit run --files <all changed files>`
  - **all hooks passed**, including Ruff, Ruff format, mypy, generated trainer config verification, test structure, naming, and compile-all
[...]

PR #408, opened 2026-08-19
[ci, worker, rollout, cfg] chore: upgrade verl pin to 0.9.0 latest (c4b389a) and adapt config and agent loop

SamitHuang:
### What does this PR do?

Upgrades the `verl` pin to commit `c4b389adadc58ce51cb2b63e70df497ca166d77f` (after v0.9.0, with fsdp2 fix) and cleanly adapts `verl-omni` to upstream changes:
1. **Remove temporary FusedMoE monkey-patches**: Deleted obsolete `.pth` lazy hooks and `tests/conftest.py` as upstream `verl` natively supports vLLM 0.27 MoE modular refactoring.
2. **Agent loop tokenization adaptation**: Updated `DiffusionSingleTurnAgentLoop` and `LTX2DiffusionSingleTurnAgentLoop` to use `ct_build_initial_tokens` (and pass `hf_model_type`), aligning with the new `AgentLoopBase` API.
3. **Model and Rollout config alignment**: Added `hf_config`, `seed` (default `42`), and `full_determinism` to `DiffusionModelConfig` / `DiffusionRolloutConfig` and reward configs.
4. **Weight transfer callback update**: Updated `vLLMOmniColocateWorkerExtension.update_weights_from_ipc` to handle variable-argument callbacks passed by `BucketedWeightReceiver.receive_weights`.
5. **Regenerated trainer configs**: Updated `_generated_*.yaml` reference configs via `scripts/generate_trainer_config.sh`.

### Checklist Before Starting

- [ ] Search for similar PRs. Paste at least one query link here: ...
- [ ] Format the PR title as `[{modules}] {type}: {description}` (This will be checked by the CI)
  - `{modules}` include `fsdp`, `vllm_omni`, `rollout`, `trainer`, `ci`, `training_utils`, `recipe`, `ray`, `worker`, `single_controller`, `misc`, `perf`, `model`, `algo`, `env`, `tool`, `ckpt`, `doc`, `data`, `cfg`, `reward`, `diffusion`, `omni`, `tests`, `docker`
  - If this PR involves multiple modules, separate them with `,` like `[diffusion, doc]`
  - `{type}` is in `feat`, `fix`, `refactor`, `chore`, `test`
  - If this PR breaks any API (CLI arguments, config, function signature, etc.), add `[BREAKING]` to the beginning of the title.
  - Example: `[BREAKING][diffusion, fsdp] feat: new rollout scheduler`

### Test

All GPU smoke test suites and pre-commit checks executed and passed:

[...]

PR #454, opened 2026-08-27
[diffusion, tests] fix: support MiniMax H3 continuous tokens

NancyFyong:
### What does this PR do?

Fix MiniMax H3 DiffusionNFT prompt construction with the pinned Verl
continuous-token API.

MiniMax H3 intentionally uses raw text token IDs rather than a chat template.
Its agent loop previously bypassed base initialization to avoid mutating a
shared Rust tokenizer while agent loops are constructed concurrently. After the
Verl pin upgrade in #408, the framework invokes the continuous-token entry
point, but the bypassed initialization left the required shared state absent.
This caused MiniMax H3 rollout to fail before generation.

This PR lets H3 inherit the current base agent-loop initialization, which does
not tokenize at construction time, and overrides only both tokenization entry
points to preserve the H3 raw-text contract. It also adds CPU coverage for
both entry points.

This unblocks the MiniMax H3 diffusion smoke path referenced by #444. It does
not overlap with #444's vLLM-Omni generation-strategy refactor.

### Checklist Before Starting

- [x] Search for similar PRs. Queries: [MiniMax H3 continuous token](https://github.com/verl-project/verl-omni/pulls?q=is%3Apr+is%3Aopen+%22MiniMax+H3%22+%22continuous+token%22) and [continuous_token_builder minimax](https://github.com/verl-project/verl-omni/pulls?q=is%3Apr+is%3Aopen+continuous_token_builder+minimax). No separate open PR implements this fix. #444 identifies the same pre-generation failure as out of scope for its strategy refactor.

### Test

```bash
TORCH_COMPILE_DISABLE=1 TORCHINDUCTOR_DISABLE=1 \
  python -m pytest -q -s --asyncio-mode=auto \
  tests/agent_loop/test_minimax_h3_agent_loop_on_cpu.py
# 7 passed
```

The same targeted CPU test passed in both the project CPU test environment and
the H3 GPU-runtime dependency environment.

```bash
SKIP=autogen-trainer-cfg pre-commit run --files \
  verl_omni/pipelines/minimax_h3_diffusion_nft/agent_loop.py \
  tests/agent_loop/test_minimax_h3_agent_loop_on_cpu.py
# all applicable hooks passed

git diff --check
# passed

PR_TITLE='[diffusion, tests] fix: support MiniMax H3 continuous tokens' \
  python tests/special_sanity/check_pr_title.py
# passed
```

`autogen-trainer-cfg` is intentionally skipped because this change does not
modify trainer config dataclasses or generated YAML inputs.

### API and Usage Example

No public API, configuration, or launch command changes. Existing MiniMax H3
recipes continue using the same agent-loop name and now work with the pinned
Verl continuous-token path.

### Design & Code Changes

1. Inherit the current base initializer so shared continuous-token state is
   available without duplicating Verl initialization. The initializer creates
   the builder but does not tokenize or render a chat template.
2. Keep H3's raw-text semantics by routing both
   `ct_build_initial_tokens()` and the legacy tokenization entry point through
   one helper using `add_special_tokens=False`.
3. Preserve the existing native-token request marker that prevents the H3
   rollout from accepting generic chat-template token IDs.
4. Test raw-text IDs and empty-text rejection through both entry points. The
   initializer test verifies that the tokenizer is not touched at construction.

- [x] Read the Contribute Guide and testing guide.
- [x] Run applicable pre-commit checks.
- [x] Documentation is not required because there is no public API or usage change.
- [x] Expand the existing CPU test covered by the CPU CI workflow.

### AI assistance and human accountability

AI assistance (pi) was used for investigation, implementation, and validation.
The human submitter reviewed every changed line and is responsible for this
contribution.


verl_omni/pipelines/minimax_h3_diffusion_nft/agent_loop.py
```diff
@@ -41,41 +40,29 @@ def __init__(
...
-        self.config = trainer_config.config
-        self.rollout_config = self.config.actor_rollout_ref.rollout
-        self.server_manager = server_manager
-        self.tokenizer = tokenizer
-        self.processor = processor
-        self.dataset_cls = dataset_cls
-        self.data_config = data_config.config
-        self.apply_chat_template_kwargs = self.data_config.get("apply_chat_template_kwargs", {})
-        self.mm_processor_kwargs = self.data_config.get("mm_processor_kwargs", {})
+        # Current verl initializes its Continuous Token builder here without
```
mengchengTang: [review] Could we avoid overriding `__init__` here? The parent class already initializes the shared AgentLoop state, including the Continuous Token builder. Keeping a subclass initializer increases the risk of drifting from future verl updates. H3-specific behavior can be handled by overriding only the token-building methods.

NancyFyong: Good idea, I'll fix it

NancyFyong: Thanks, agreed and updated in 944da53. I removed the H3 __init__ override entirely. H3 now inherits DiffusionSingleTurnAgentLoop initialization for the Continuous Token builder and extra-tokenizer state, while overriding only the two raw-text token-building entry points. The CPU test now verifies inherited initialization does not tokenize and provides Continuous Token state.

NancyFyong:
@zhtmike , we will review and fix it as quickly as possible; this bug appears in Minimax following the Verl upgrade.
Thank you!

zhtmike (approved):
will the verl update affect all diffusion models?

NancyFyong:
> will the verl update affect all diffusion models?

So far, we’ve only observed this with the Minimax model. What do you think about writing a dedicated `AgentLoop` base class for this type of model—one that doesn't require a template?

zhtmike:
> > will the verl update affect all diffusion models?
> 
> So far, we’ve only observed this with the Minimax model. What do you think about writing a dedicated `AgentLoop` base class for this type of model—one that doesn't require a template?

yeah sure

Issue #445, opened 2026-08-25
[RFC][tracking] Upstream patch debt: audited ledger of temporary compatibility patches


zhtmike:
## Motivation

verl-omni sits on top of verl / vllm-omni / diffusers / transformers / peft and carries temporary patches. Each has one of three fates: **drop after rebase/bump** (fixed upstream; our pin or version floor predates the fix), **drop now** (fix already inside our pins — a delete or, for P3, an in-pin rewrite), or **blocked on upstream** (needs a fix we can only suggest — see Tracker §3). Critically, some things that *look* like patches are **intended adaptations** — code that exists because verl-omni trains diffusion/omni models that upstream deliberately doesn't cover, and which would survive a perfect upstream (Part II). Mixing those into patch debt produces wrong drop plans, so this ledger separates them.

Current tally (round 3): **28 patch-debt line items** — **8 landed** (P1–P8, via the verl re-pin #450 and the vllm-omni-0.28 re-pin #497) · **20 open** (P9 now a *delete-now*, P10–P28) · **16 intended adaptations** (Part II, A1–A16) · 2 false positives removed in round 1. Round 3 verified the open items against exact pins with **vllm 0.28.0 and verl@pin as the reference standard** and classified every deviation: `[upstream bug]` (violates upstream's own contract) · `[upstream gap]` (capability absent) · `[intended deviation]` (deliberate domain design) · `[downstream bug/risk/gap]` (our code is wrong, fragile, or missing — see P26–P28).

**Round 3 (2026-09-07) — adversarial verification at exact pins. Headlines:**
- **P9 FLIPPED to delete-now:** the CI FusedMoE `.pth` alias is no longer load-bearing (verl pin contains [verl#7147](https://github.com/verl-project/verl/pull/7147)+[verl#7538](https://github.com/verl-project/verl/pull/7538); verl-omni has zero vllm-`FusedMoE` imports) and now actively *degrades* verl's dispatch (forces the legacy duck-typing branch) — delete `l3_nightly.yml:186-188`.
[...]

PR #464, opened 2026-08-27
[diffusion, rollout, trainer, data, recipe, tests] feat: add MiniMax H3 Ref2VA multi-reference DiffusionNFT

NancyFyong:
### What does this PR do?

Extend MiniMax‑H3 Ref2VA DiffusionNFT from the initial single‑reference milestone to **full multi‑reference conditioning**: up to 9 images + 3 reference videos + 3 audios per prompt (≤12 references total, ≥1 image‑or‑video), matching the FlowGRPO Ref2VA reference implementation. References are decoded by the native vLLM‑Omni `MiniMaxH3Pipeline` while the Agent‑Loop text token IDs are preserved at the tail, and the DiffusionNFT actor replays the full packed layout, scoring only the target video/audio streams.

This extends the existing draft **#464** and builds on the merged FL2VA DiffusionNFT (**#401**); the FlowGRPO Ref2VA branch (no open PR) is the reference implementation.

### Checklist Before Starting

- [x] Search for similar PRs:
  [is:pr Ref2VA](https://github.com/verl-project/verl-omni/pulls?q=is%3Apr+Ref2VA) ·
  [is:pr MiniMax H3](https://github.com/verl-project/verl-omni/pulls?q=is%3Apr+MiniMax+H3).
  Related: #401 (FL2VA DiffusionNFT, merged), #464 (this PR). No other open PR covers Ref2VA DiffusionNFT.
- [x] Title format `[{modules}] {type}: {description}`.

### Test

Affected CPU tests (`151 passed`):

```bash
python -m pytest \
  tests/utils/test_padding_on_cpu.py \
  tests/utils/test_minimax_h3_ref2va_data_process_on_cpu.py \
  tests/agent_loop/test_minimax_h3_agent_loop_on_cpu.py \
  tests/pipelines/test_minimax_h3_diffusion_nft_on_cpu.py \
  tests/pipelines/test_minimax_h3_ref2va_nft_on_cpu.py \
  tests/workers/config/test_diffusion_config_on_cpu.py \
  tests/workers/rollout/rollout_vllm/test_vllm_omni_output_on_cpu.py \
  tests/workers/rollout/rollout_vllm/test_vllm_omni_deploy_config_on_cpu.py \
  tests/workers/rollout/rollout_vllm/test_vllm_omni_lora_on_cpu.py
```

Repository-wide pre-commit (`590` tracked files; all 12 hooks passed):

```bash
mapfile -d '' FILES < <(git ls-files -z)
pre-commit run --files "${FILES[@]}"
```

Coverage: mixed-reference parquet serialization and limit validation, actor/rollout config propagation, fixed-size `(MAX_REF_BLOCKS=12, 5)` reference-block serialization + count, global Agent-Loop reference-row padding with fail-closed shape/count/limit checks, dense-to-jagged conversion, minibatch-local restoration, per-sample condition-row slicing, packed-layout replay parity, prompt-token preservation across image+video+audio references, and reference-media passthrough in the Agent Loop.

Real‑weight: the single‑reference Ref2VA path was validated end‑to‑end (reward curves below). The multi‑reference serialization / padding / replay run on the same shared rollout+actor path that the merged FL2VA (#401) real‑weight runs already exercise; the new code beyond it is covered by the CPU contracts above.

image

image

image

### API and Usage Example

```bash
# 1) Convert JSONL (rows carry images / videos / audios) into packed parquet.
python examples/diffusionnft_trainer/minimax_h3/prepare_ref2va_data.py \
  --input_dir /path/to/ref2va_jsonl \
  --output_dir /path/to/ref2va_parquet

# 2) Train. MODEL_PATH is the MiniMax-H3 repo root holding Ref2VA/ and
#    transformer_ref/ as siblings.
MODEL_PATH=/path/to/MiniMax-H3 \
DATA_DIR=/path/to/ref2va_parquet \
  bash examples/diffusionnft_trainer/minimax_h3/run_minimax_h3_ref2va_lora.sh
```

Prompt content is `"<image>" * n_img + "<video>" * n_vid + "<audio>" * n_aud + text`; `RLHFDataset` expands each marker into a structured multimodal content block. New typed pipeline config: `actor_rollout_ref.rollout.pipeline.reference_image_short_edge` (also settable via the recipe's `REF_IMAGE_SHORT_EDGE` env, a per‑request override).

### Design & Code Changes

High‑level: unlike the single‑reference milestone, references are **variable in count and mixed in modality**, so the request → transport → replay contract must carry a **variable‑length, cross‑worker‑paddable** reference layout. Most of that is contained in the H3 pipeline package; a few generic/base layers had to change, each justified below together with why FL2VA did not need it.

**H3‑specific (`verl_omni/pipelines/minimax_h3_diffusion_nft/`)**

- `common.py`: `serialize_ref_blocks` now returns a fixed‑size `(meta[MAX_REF_BLOCKS=12, 5], count)`; drop the single‑image gate and route `ref2va` through `_PromptTokenOverride` + native `encode_prompt`, so the native pipeline decodes every reference modality while the exact Agent‑Loop text token IDs are preserved at the tail. The LoRA target whitelist is unchanged.
- `vllm_omni_rollout_adapter.py` `diffuse()`: capture multi‑image/video conditions via `minimax_h3_imgvid_cond_noise_aug_rows` and standalone audio via `minimax_h3_audio_cond_noise_aug_rows`, then emit `condition_video/audio_row_count` and the fixed‑size `ref_block_meta`.
- `diffusers_training_adapter.py`: slice condition rows per sample by the emitted count (a `None` fallback preserves pre‑multi‑reference behavior for t2va/fl2va replay).
- `agent_loop.py`: `process_multi_modal_info` keeps reference media in upstream format (images raw, videos `{path, start_time_seconds}`, audios raw).

**Why the generic/base layers changed (and why FL2VA did not need them)**

1. `verl_omni/agent_loop/diffusion_agent_loop.py`, `verl_omni/workers/utils/padding.py`, and the DiffusionNFT FSDP engine — **carry variable reference rows without model-specific manager logic.** The Agent Loop pads `condition_video_rows` / `condition_audio_rows` to the global `max_prompt_embed_length`, emits validity masks, and fails closed on invalid shapes, mismatched counts, or overflow. The standard `AgentLoopManager` can then concatenate worker outputs without knowing model fields. Before actor training, `embeds_padding_2_no_padding` converts the globally padded rows to jagged TensorDict values; the DiffusionNFT engine restores only the minibatch-local maximum and validates each mask against its row count. **FL2VA did not require this path** because its condition-row count is normally uniform, while Ref2VA varies by reference count and modality.
2. `verl_omni/agent_loop/single_turn_agent_loop.py` + `verl_omni/workers/rollout/vllm_rollout/vllm_omni_diffusion_strategy.py` — **carry `mm_processor_kwargs` onto the diffusion prompt.** Ref2VA delegates prompt building to the native multimodal pipeline, and reference **video (`fps`)** / **audio (`sampling_rate`)** must reach that pipeline's processor. The base diffusion rollout strategy accepted `mm_processor_kwargs` but dropped it (only the AR strategy attached it), so the agent loop now passes `_get_mm_processor_kwargs(audios)` and `DiffusionStrategy.preprocess_input` attaches it, mirroring `ARStrategy`. **FL2VA did not need it**: a single condition image carries no `fps` / `sampling_rate`.
3. `verl_omni/workers/rollout/vllm_rollout/utils.py` — **`_move_diffusion_lora_stacks_to_device` after `add_lora`.** Ref2VA's long (12288) reference prompt embedding runs with text‑encoder TP + layerwise offload; on the diffusion‑worker online LoRA‑sync path, vLLM‑Omni's LoRA manager registers the adapter but leaves `lora_a_stacked` / `lora_b_stacked` off the worker GPU, so the next DiT forward would hit a CPU/GPU device mismatch. The helper places them on device, reusing the pre‑existing AR‑vs‑diffusion discriminator (`_get_standard_weight_model_and_config`). This is a downstream workaround for a vLLM‑Omni gap; ideally it is fixed upstream in the diffusion LoRA manager.

**Config / data / recipe / docs**: typed `pipeline.reference_image_short_edge` (+ regenerated `_generated_*.yaml`), the multi‑reference `prepare_ref2va_data.py` converter, `run_minimax_h3_ref2va_lora.sh`, and the recipe README.

- [x] Apply pre-commit checks (pass on the changed files).
- [x] Add / Update the documentation (recipe README).
- [x] Add unit / end-to-end tests (CPU contracts listed under Test).
- [x] Human submitter reviewed every changed line.


verl_omni/pipelines/minimax_h3_diffusion_nft/common.py
```diff
@@ -54,6 +65,26 @@
...
+        raise ValueError("REF_IMAGE_SHORT_EDGE must be a multiple of 32 between 256 and 2048.")
+
+    from vllm_omni.diffusion.models.minimax_h3.pipeline_minimax_h3 import _reference_image_shape
+
+    resize_globals = _reference_image_shape.__globals__
+    constant = "MINIMAX_H3_REFERENCE_IMAGE_SHORT_EDGE"
+    if constant not in resize_globals:
+        raise RuntimeError("vLLM-Omni no longer exposes the MiniMax H3 reference image size constant.")
+    resize_globals[constant] = short_edge
+    return short_edge
```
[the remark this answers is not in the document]
NancyFyong: Accepted and fixed in fa316a8. The override is now applied only inside a process-wide RLock around the native pipeline forward and restored in a finally block, so concurrent train/validation requests cannot observe each other's short-edge value. This preserves distinct REF_IMAGE_SHORT_EDGE and VAL_REF_IMAGE_SHORT_EDGE settings. Added regression coverage for restoration, request override behavior, and concurrent serialization; the PR CPU set is 55 passed and all pre-commit hooks pass.

verl_omni/agent_loop/diffusion_agent_loop.py
```diff
@@ -46,6 +51,42 @@ def _config_to_sampling_dict(config: Optional[BaseConfig]) -> dict:
     return {k: v for k, v in config.items() if not k.startswith("_")}
 
 
+# Reference rows vary per sample, so pad them to a common length before stacking.
+_VARIABLE_REFERENCE_FIELDS = ("condition_video_rows", "condition_audio_rows")
```
zhtmike: pls make it inline

verl_omni/agent_loop/diffusion_agent_loop.py
```diff
@@ -405,3 +455,29 @@ def _postprocess(
...
+                for worker, chunk in zip(self.agent_loop_workers, chunks, strict=True)
+            ]
+        )
+        _pad_reference_worker_outputs(outputs)
+        output = DataProto.concat(outputs)
+
+        metrics = [item.meta_info.pop("metrics") for item in outputs]
+        timing = self._performance_metrics(metrics, output)
+        output.meta_info = {"timing": timing, **outputs[0].meta_info}
+        return output
```
zhtmike: why doing padding this in loop manager, pls drop this, padding them in the agent loop.

loop manager does not responsible for padding or data structures

verl_omni/pipelines/minimax_h3_diffusion_nft/agent_loop.py
```diff
@@ -27,13 +27,39 @@
...
+                elif media_type == "video":
+                    value = item.get("video", item.get("video_url", item.get("path")))
+                    if value is not None:
+                        start = item.get("start_time_seconds")
+                        media["videos"].append(value if start is None else {"path": value, "start_time_seconds": start})
+                elif media_type == "audio":
+                    value = item.get("audio", item.get("audio_url"))
+                    if value is not None:
+                        media["audios"].append(value)
+        return {key: values for key, values in media.items() if values}
```
zhtmike, line 56: hard to read.

simply this, drop the fallback or compatibility code.

verl_omni/trainer/config/diffusion/rollout/diffusion_rollout.yaml
```diff
@@ -272,6 +275,9 @@ pipeline:
   # Distilled guidance scale for models with guidance_embeds; null to disable
   guidance_scale: null
 
+  # Reference-image short edge for compatible pipelines
+  reference_image_short_edge: null
```
zhtmike, line 279: should update doc for this.

verl_omni/workers/rollout/vllm_rollout/utils.py
```diff
@@ -160,6 +169,8 @@ def update_weights_from_ipc(
                 )
             t2 = time.perf_counter()
             self.add_lora(lora_request)
+            if self._get_standard_weight_model_and_config() is None:
+                self._move_diffusion_lora_stacks_to_device()
```
zhtmike, line 174: why need this?

verl_omni/agent_loop/diffusion_agent_loop.py
NancyFyong: Fixed in d04f7a9. Removed _VARIABLE_REFERENCE_FIELDS and inlined the two field names at both call sites.

verl_omni/pipelines/minimax_h3_diffusion_nft/agent_loop.py
NancyFyong: Fixed in d04f7a9. process_multi_modal_info now reads only the canonical image/video/audio keys produced by RLHFDataset; the image_url/video_url/path/audio_url compatibility fallbacks were removed. Video start_time_seconds is still preserved.

verl_omni/trainer/config/diffusion/rollout/diffusion_rollout.yaml
NancyFyong: Fixed in d04f7a9. Added reference_image_short_edge to the DiffusionPipelineConfig example and documented its MiniMax-H3 Ref2VA range plus the val_kwargs override in docs/examples/config.md.

verl_omni/agent_loop/diffusion_agent_loop.py
NancyFyong: Good point. We already pad the rows inside each agent-loop worker in `_postprocess`. The extra step in the manager is only there because different workers may still return different maximum row counts, and `DataProto.concat` will fail when it tries to `torch.cat` shapes such as `[B, 20, 96]` and `[B, 35, 96]`.

I agree that the manager should not know these model-specific fields. Simply removing it is not enough, though, because each worker does not know the other workers' maximum length. Do you have a preferred pattern for this case? We could pad to a fixed limit inside the agent loop, but that may waste a lot of memory for Ref2VA inputs.

verl_omni/workers/rollout/vllm_rollout/utils.py
NancyFyong: This is a workaround for the pinned vLLM-Omni revision `444485650`.

With CPU or layerwise offload enabled, `add_adapter()` can create `lora_a_stacked` and `lora_b_stacked` on CPU. They are stored as plain tensor tuples rather than registered buffers, so moving or waking the DiT does not move them to CUDA. The next LoRA matmul then fails because the activations are on CUDA while the LoRA tensors are still on CPU.

The regular AR vLLM path does not have this issue, so the workaround is limited to diffusion workers. Ideally this should be fixed in vLLM-Omni's `DiffusionLoRAManager`; for the currently pinned revision, we still need the device move after `add_lora()`.

verl_omni/agent_loop/diffusion_agent_loop.py
zhtmike: 1. Pad to a global constant in agent loop.
2. use `embeds_padding_2_no_padding` to trim the pad in the trainer.
3. use `_unpad_nested_embeds` to provide the minimal padding for each minibatch input before feeding into the training engine.

check qwen-image flowgrpo script of this work flow.

verl_omni/workers/rollout/vllm_rollout/utils.py
zhtmike: 1. why other diffusion scripts doesnt need this? they are cpu offloaded, as well.
2. if it is a workaround, better to mention it as TODO. And later provide the formal fix. Here or upstream.

zhtmike:
We can use tensordict to completely eliminate the padding in agentloop in future, only keep minimum padding before feeding into engine, but it is not this PR's scope.

verl_omni/workers/rollout/vllm_rollout/utils.py
NancyFyong: The difference is layerwise offload rather than model-level CPU offload. The Ref2VA recipe explicitly enables `enable_layerwise_offload=True`; when the rollout adapter is dynamically replaced after weight sync, those DiT layers are still CPU-resident, so `create_lora_weights()` derives CPU placement from the wrapped base layer. The LoRA stacks are tensor tuples, so the later per-layer prefetch does not move them with the module. The existing recipes using model-level `enable_cpu_offload` wake the whole active module before this path and have not needed this workaround.

I added `TODO(@NancyFyong)` in 65839be to track moving this into vLLM-Omni `DiffusionLoRAManager`. Once the pinned vLLM-Omni revision includes the formal placement fix, this downstream move can be removed.

verl_omni/agent_loop/diffusion_agent_loop.py
NancyFyong: I think agentloop will require refactoring once we have a larger number of models; for now, I will use an interim solution, and I will implement fast-fail checks for boundary conditions.

verl_omni/workers/rollout/vllm_rollout/utils.py
zhtmike: tracked in https://github.com/verl-project/verl-omni/issues/445

verl_omni/agent_loop/diffusion_agent_loop.py
NancyFyong: Updated in 5ad1924 following this workflow:

1. Removed `DiffusionAgentLoopManager` and restored the standard `AgentLoopManager`.
2. The agent-loop worker now pads `condition_video_rows` and `condition_audio_rows` to the global `max_prompt_embed_length` and emits validity masks.
3. `embeds_padding_2_no_padding` strips those rows into jagged TensorDict values in the trainer.
4. `NFTDiffusersFSDPEngine` restores only the minibatch-local maximum before model input.

The padding path now fails closed if the row tensor shape is invalid, the declared row count differs from the actual rows, the global limit is exceeded, or the restored mask disagrees with the row count. No rows are silently truncated. The affected CPU suite passes 109 tests, and all pre-commit hooks pass.

NancyFyong: fix done

zhtmike: I think we need to update doc for this change

NancyFyong: done

zhtmike:
pls fix AI review

verl_omni/workers/config/diffusion/rollout.py
```diff
@@ -66,6 +66,7 @@ class DiffusionPipelineConfig(BaseConfig):
     true_cfg_scale: float = 1.0
     max_sequence_length: int = 512
     guidance_scale: Optional[float] = None
+    reference_image_short_edge: Optional[int] = None
```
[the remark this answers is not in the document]
NancyFyong: Fixed in 25b0bef. Added `reference_image_short_edge: ${oc.select:actor_rollout_ref.rollout.pipeline.reference_image_short_edge,null}` to the actor-side `diffusion_model.yaml` pipeline mapping and regenerated both diffusion trainer configs. Hydra composition with a rollout value of 1024 now resolves the actor model pipeline value to 1024 as well.

examples/diffusionnft_trainer/minimax_h3/prepare_ref2va_data.py
```diff
@@ -0,0 +1,150 @@
...
+                if start < 0:
+                    raise ValueError(f"video start_time_seconds must be non-negative, got {start}.")
+        else:
+            raise ValueError("videos entries must be paths or objects containing a path.")
+        videos.append({"video": str(path), "start_time_seconds": start})
+        sources.append(str(path))
+    return videos, sources
+
+
+def _convert_split(input_dir: Path, split: str, max_samples: int) -> pd.DataFrame:
```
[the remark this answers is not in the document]
NancyFyong: Fixed in 25b0bef. Added `tests/utils/test_minimax_h3_ref2va_data_process_on_cpu.py`, covering singular/plural normalization, image/video/audio ordering, relative-to-absolute path resolution, parquet serialization, per-modality limits, and the total 12-reference limit. The focused converter suite passes 5 tests; the broader affected suite passes 151 tests.

zhtmike:
Seems flacky CI, LGTM

zhtmike (approved):

Issue #97, opened 2026-05-20
[RFC] Q3 Road Map (welcome contributions)


SamitHuang:
We are open to collect feature, model, architecture requests for Q3 development roadmap. Welcome your feedbacks and comments.

### Training Stability

* Determinstic RL training 🙋
  - [ ] **vLLM-Omni determinstic rollout**  @MmMaiIIi  @KaisennHu 
  - [ ] vLLM determinstic reward compute (VLM)    @KaisennHu 
    - [ ] Agent Loop Worker https://github.com/verl-project/verl/issues/6570 https://github.com/verl-project/verl/pull/6572
    - [ ] Reward Loop Worker
  - [ ] Actor and E2E determinstic training  @KaisennHu 

* Rollout-Train consistency
  - [ ]  Support diffusers rollouter, via diffusion-backend in vllm-omni or train-side engine https://github.com/vllm-project/vllm-omni/pull/2724 https://github.com/vllm-project/vllm-omni/pull/3096   🙋 
  - [ ]  Improve VeOmni/DiffusersFSDP trainer and vLLM-Omni rollout consistency (attention kernel, fused ops, mixed precision config etc)   @cr-gao 
  - [x]  Attention kernel alignment: support `kernel` package backend in vllm-omni  https://github.com/vllm-project/vllm-omni/pull/4926
  - [ ]  0-diff train rollout for true on-policy RL (reference: https://github.com/verl-project/verl/pull/7688)

### Efficiency
* System optimizations
  - [ ] Rollout: 
    - [x] prompt embedding cache @knlnguyen1802 https://github.com/vllm-project/vllm-omni/pull/2962
    - [x] support sample-level batching & scheduling in vllm-omni and validate performace   #172    https://github.com/vllm-project/vllm-omni/pull/4079
  - [x] Trainer: 
    - [x] integrate VeOmni trainer backend  https://github.com/verl-project/verl-omni/issues/92
  - [x] **VeRL v1 trainer adaptation**
  - [ ] Improve performance profiling and resolve bottleneck  #219 
  - [ ] Investigate and improve larger-scale deployment performance
  - [x] Upstream version rebase
    - [x] verl 0.9.0
    - [x] vllm-omni 0.24 or later

* Easy-to-Use
  - [ ] Save lora-only weight during training for inference usage  @guozhihao-224 
  - [x] Omni-series model support refactor #232  @zhtmike 

### CI tests and Usability

- [ ] Standard convergence test for diffusion models
[...]

PR #479, opened 2026-08-29
[rollout] feat: prototype train-side diffusers SDE sampler

OnePunchMonk:
### What does this PR do?

Prototype for the Q3 roadmap item "Support diffusers rollouter ... via train-side engine" (#97).

Adds `verl_omni/workers/rollout/diffusers_train_rollout.py::sde_denoise_loop`, a device/engine-agnostic port of the SD3 FlowGRPO rollout adapter's SDE denoising loop (`FlowMatchSDEDiscreteScheduler`) that runs directly against the training transformer in-process, with no vLLM-Omni server involved.

Scope note: this covers only the sampling core shared by the SD3-family rollout adapters -- not a drop-in `AsyncRollout` implementation. Request batching, async server plumbing, and rollout config schema wiring are left as follow-up once the sampling core itself is trusted, so it is intentionally **not** registered in `_ROLLOUT_REGISTRY` yet.

## Duplicate-work check

- `gh pr list --repo verl-project/verl-omni --state all --search "train-side rollout"` -- no other PR (open, merged, or closed) targets this roadmap item.
- `gh pr list --repo verl-project/verl-omni --state all --search "diffusers rollout"` -- no overlap; matches are unrelated diffusion-rollout work (attention backends, video export, MiniMax H3, etc.).
- Issue #97 (the Q3 roadmap tracking issue) has no assignee or prior claim comment on this specific line item.

## Checklist Before Starting

- [x] Search for similar PRs. Paste at least one query link here: https://github.com/verl-project/verl-omni/pulls?q=train-side+rollout

### Test

CPU unit tests (new, wired into CI via `tests/**/*_on_cpu.py` auto-discovery): `tests/workers/rollout/test_diffusers_train_rollout_on_cpu.py`. float32, tiny offline SD3 checkpoint (random weights, built the same way as `tests/special_e2e/build_sd3_tiny_random.py`). Validates `sde_denoise_loop` against an independently-written reference loop that batches CFG the way production SD3 pipelines actually do (single concatenated forward instead of two separate forwards), for both `do_cfg=True` and `do_cfg=False`, plus a sanity check that CFG actually changes the output.

Run locally (Python 3.12, CPU-only venv: `torch==2.4.0 diffusers==0.31.0 transformers==4.46.0`), with `verl_omni`'s and `verl_omni.workers.rollout`'s package-init eagerly importing `verl`/`vllm_omni` stubbed out the same way prior CPU-test PRs have worked around this pre-existing, unrelated environment limitation:

```
$ python3 -m pytest -v tests/workers/rollout/test_diffusers_train_rollout_on_cpu.py
tests/workers/rollout/test_diffusers_train_rollout_on_cpu.py::test_sde_denoise_loop_matches_concatenated_cfg_reference[True] PASSED
tests/workers/rollout/test_diffusers_train_rollout_on_cpu.py::test_sde_denoise_loop_matches_concatenated_cfg_reference[False] PASSED
tests/workers/rollout/test_diffusers_train_rollout_on_cpu.py::test_do_cfg_changes_output PASSED
3 passed in 3.92s
```

**GPU run (Modal, T4, fp16)**: see comment below with the parity-test results against the same reference-loop methodology at fp16 on real GPU hardware.

**vLLM-Omni diffusers-backend adapter cross-check**: see the other comment below -- separately verified the merged upstream `DiffusersAdapterPipeline` alternative also works end to end against the same tiny checkpoint, for context on the two options named in #97.

## API and Usage Example

`sde_denoise_loop` is a plain function, not yet wired into any config-driven entrypoint (see scope note above). Direct usage:

```python
from verl_omni.workers.rollout.diffusers_train_rollout import sde_denoise_loop

result = sde_denoise_loop(
    transformer,             # plain nn.Module or FSDP-wrapped SD3Transformer2DModel
    scheduler,               # a fresh FlowMatchSDEDiscreteScheduler, timesteps already set
    prompt_embeds=prompt_embeds,
    pooled_prompt_embeds=pooled_prompt_embeds,
    negative_prompt_embeds=negative_prompt_embeds,       # None to skip CFG
    negative_pooled_prompt_embeds=negative_pooled_prompt_embeds,
    latents=initial_latents,
    timesteps=scheduler.timesteps,
    do_cfg=True,
    guidance_scale=3.5,
    noise_level=0.7,
    sde_window_range=(0, num_inference_steps),
    sde_type="sde",
    generator=torch.Generator(device=device).manual_seed(seed),
    logprobs=True,
)
# result.final_latents, result.all_latents, result.all_timesteps, result.log_probs
```

`TrainSideDiffusionOutput`'s field names (`all_latents` / `all_timesteps` / `log_probs`) intentionally mirror `DiffusionOutput.extra_fields` from the existing vLLM-Omni rollout path, so downstream FlowGRPO log-prob recomputation code needs no branching on which backend produced the trajectory.

## Design & Code Changes

- `predict_noise_with_cfg`: runs the transformer once (no CFG) or twice with two separate forwards (CFG) and combines via the standard `uncond + guidance_scale * (cond - uncond)` formula. Verified equivalent to the single-concatenated-batch forward real SD3 pipelines use (see the parity tests) -- two separate forwards is simpler to read and doesn't require batch-dim bookkeeping, at the cost of one extra kernel-launch round trip per step.
- `sde_denoise_loop`: line-for-line structurally matches `StableDiffusion3PipelineWithLogProb.diffuse` (`verl_omni/pipelines/sd3_flow_grpo/vllm_omni_rollout_adapter.py`) so the two stay easy to diff against each other as either evolves, but takes plain tensors instead of `OmniDiffusionRequest`/`DiffusionRequestBatch` -- there is no request/batch abstraction on the train side to translate through.
- Deliberately not registered in `_ROLLOUT_REGISTRY` (see module docstring) until request batching, async server plumbing, and config schema wiring exist -- an incomplete registry entry would silently break rollout backend selection for every other model.

## Checklist Before Submitting

- [x] Apply [pre-commit checks](https://github.com/verl-project/verl-omni/blob/main/CONTRIBUTING.md#code-linting-and-formatting): `pre-commit install && pre-commit run --all-files --show-diff-on-failure --color=always`. All hooks pass except `autogen-trainer-cfg`, which fails on `ModuleNotFoundError: No module named 'omegaconf'` in this local env -- a pre-existing, unrelated environment gap (this PR touches no config files); confirmed the same failure exists on `main` before this diff.
- [x] Add / Update the documentation. Added a module-docstring paragraph explaining why this path exists separately from vLLM-Omni's merged `DiffusersAdapterPipeline` black-box adapter (see Design & Code Changes) -- there is no other user-facing doc surface yet since this prototype has no config-driven entrypoint.
- [x] Add unit or end-to-end test(s) to [the CI workflow](https://github.com/verl-project/verl-omni/tree/main/.github/workflows) to cover all the code. `tests/workers/rollout/test_diffusers_train_rollout_on_cpu.py` is auto-discovered by `cpu_unit_tests.yml`'s `tests/**/*_on_cpu.py` pattern.

- [x] AI assistance (Claude Code) was used for this change.
- [x] A human submitter (onepunchmonk) has reviewed every changed line.

### DIFF
--- a/tests/workers/rollout/test_diffusers_train_rollout_on_cpu.py
+++ b/tests/workers/rollout/test_diffusers_train_rollout_on_cpu.py
@@ -1,0 +1,325 @@
+# Copyright 2026 Bytedance Ltd. and/or its affiliates
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+"""CPU parity test for the train-side diffusers rollouter prototype.
+
+Validates ``sde_denoise_loop`` against an independently-written reference
+loop that batches CFG the way production SD3 pipelines actually do (a single
+concatenated forward instead of two separate forwards). Same
+``FlowMatchSDEDiscreteScheduler``, same tiny offline SD3 checkpoint (see
+``tests/special_e2e/build_sd3_tiny_random.py``), same seeds.
+
+This is the CPU-runnable, CI-wired counterpart of the GPU parity check
+posted as a comment on the PR that adds this file (a real T4 GPU run on
+Modal). float32 throughout, since fp16 on CPU is not well supported by all
+of the ops this loop touches and isn't the point of this test -- the fp16
+run on GPU covers that regime separately.
+"""
+
+import os
+import sys
+import types
+
+import pytest
+import torch
+
+sys.path.insert(0, os.path.join(os.path.dirname(__file__), "..", "..", "special_e2e"))
+
+diffusers = pytest.importorskip("diffusers")
+
+
+def _stub_heavy_parent_packages() -> None:
+    """Let ``verl_omni.workers.rollout.diffusers_train_rollout`` import without
+    the full install (``verl``, ``vllm_omni``) this module doesn't need.
+
+    ``verl_omni/__init__.py`` and ``verl_omni/workers/rollout/__init__.py``
+    eagerly import ``verl``-dependent submodules at package-import time (see
+    ``tests/utils/test_stable_diffusion_3_flops_on_cpu.py``-style PRs for the
+    same pre-existing environment limitation, unrelated to this change).
+    Pre-registering empty namespace packages -- pointed at the real source
+    tree via ``__path__`` -- skips only those heavy ``__init__``s while still
+    resolving submodule imports (this file, and
+    ``verl_omni.pipelines.schedulers``) to the real, unmodified source.
+    """
+    repo_root = os.path.abspath(os.path.join(os.path.dirname(__file__), "..", "..", ".."))
+    for dotted_path in (
+        "verl_omni",
+        "verl_omni.pipelines",
+        "verl_omni.workers",
+        "verl_omni.workers.rollout",
+    ):
+        if dotted_path in sys.modules:
+            continue
+        module = types.ModuleType(dotted_path)
+        module.__path__ = [os.path.join(repo_root, *dotted_path.split("."))]
+        sys.modules[dotted_path] = module
+
+
+_stub_heavy_parent_packages()
+
+
+def _build_pipeline_and_scheduler():
+    import build_sd3_tiny_random
+    from diffusers import StableDiffusion3Pipeline
+
+    from verl_omni.pipelines.schedulers import FlowMatchSDEDiscreteScheduler
+
+    # The repo's helper assumes a T5Tokenizer(vocab=...) API from a
+    # transformers version/fork not pinned by this test; substitute a real
+    # sentencepiece-backed T5Tokenizer trained on the same tiny corpus.
+    def _real_t5_tokenizer(*, vocab_size: int = 32, model_max_length: int = 512):
+        import tempfile
+
+        import sentencepiece as spm
+        from transformers import T5Tokenizer
+
+        with tempfile.TemporaryDirectory() as tmp_dir:
+            corpus_path = os.path.join(tmp_dir, "corpus.txt")
+            with open(corpus_path, "w") as f:
+                f.write("a red circle on a white background\n")
+                f.write("a blue square on a black background\n")
+                f.write("a green triangle next to an orange rectangle\n")
+            model_prefix = os.path.join(tmp_dir, "spm")
+            spm.SentencePieceTrainer.train(
+                input=corpus_path,
+                model_prefix=model_prefix,
+                vocab_size=vocab_size,
+                model_type="unigram",
+                pad_id=0,
+                unk_id=2,
+                eos_id=1,
+                bos_id=-1,
+                pad_piece="<pad>",
+                unk_piece="<unk>",
+                eos_piece="</s>",
+            )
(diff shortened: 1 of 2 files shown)


OnePunchMonk:
## GPU parity test (Modal, T4)

Validated `sde_denoise_loop` against an independently-written reference loop that batches CFG the way production SD3 pipelines actually do (single concatenated forward instead of two separate forwards). Same `FlowMatchSDEDiscreteScheduler`, same tiny offline SD3 checkpoint (random weights, Hub-free, built the same way as `tests/special_e2e/build_sd3_tiny_random.py`), same seeds, 6 denoising steps.

| Check | Result |
|---|---|
| Log-probs (2 implementations) | exact match, diff = 0.0 |
| Latent trajectory / final latents | max abs diff 0.0042 (fp16 batching-order noise, expected) |
| CFG demonstrably changes output | true |
| Output finite | true |

Log-prob agreement being bit-exact is the strongest signal here: it means the scheduler math that FlowGRPO training actually consumes is identical regardless of how CFG is batched. The small latent diff is expected fp16 matmul-accumulation-order noise between two-separate-forwards vs. one-concatenated-forward, not a correctness issue.

Not run against the real vLLM-Omni SD3 rollout adapter directly since `vllm_omni` isn't installable in this environment -- the reference loop above is the closest available oracle, since the SD3 rollout adapter's own `diffuse()` method is this same scheduler math over the same transformer.

OnePunchMonk:
Related to the train-side rollouter item in the Q3 roadmap: #97 (see comment there linking back to this PR).

OnePunchMonk:
## Option 1 (vLLM-Omni diffusion-backend) also verified working, for context

While validating this train-side prototype, I also tested the other alternative listed in #97 -- vLLM-Omni's own generic diffusers-backend adapter (`vllm_omni.diffusion.models.diffusers_adapter.DiffusersAdapterPipeline`, from vllm-project/vllm-omni#2724, merged upstream). Ran it end-to-end on a CPU-only container (no GPU needed) against the same tiny offline SD3.5 checkpoint:

- Installed `vllm==0.27.0` + `vllm-omni@444485650` (the exact commit this repo's `.github/vllm_omni_pin.txt` pins)
- `DiffusersAdapterPipeline(od_config=...)` + `.load_weights()` -> loaded the tiny SD3.5 pipeline (`Weights loaded OK: StableDiffusion3Pipeline`)
- Built a real `OmniDiffusionRequest` wrapped in a `DiffusionRequestBatch`, called `adapter.forward(...)` -> ran a genuine 4-step diffusers denoising loop and returned a real `DiffusionOutput` (PIL image list)

So both alternatives named in the roadmap item are now confirmed as real, working paths, not just this PR's prototype:

- **Option 1** (vLLM-Omni diffusion-backend): already implemented and merged upstream, works as a black-box adapter today (`--diffusion-load-format diffusers`), but is explicitly a "black-box" per its own docs -- no CFG parallel, no sequence parallel, no TeaCache/Cache-DiT, no step-wise/continuous batching, and no per-step log-prob/trajectory data, so it can't feed FlowGRPO-style training as-is.
- **Option 2** (this PR, train-side engine): narrower in scope today (sampling core only, not wired into `_ROLLOUT_REGISTRY`), but designed from the start to produce the per-step log-probs and trajectory data FlowGRPO/RL training needs.

Noting this so the maintainer has both data points when deciding which direction to invest in for #97.

SamitHuang:
Thanks for the contribution. I think we need a RFC issue to disucss the design of train-side rollouter, since we may need to refactor the rollout interfaces to support both train-sider rollouter and vllm-omni rollouter.

Btw, E2E validation for train-rollout consistency test and model convergence test (like SD3.5/QwenImage) are also essential for this PR.

OnePunchMonk:
@SamitHuang opened the rfc issue. please let me know your thoughts on the design. thank you !

OnePunchMonk:
marking this draft, until the design is finalised in #484

PR #427, opened 2026-08-23
[worker, trainer, cfg, tests, doc] feat: multi-teacher routing and standalone teacher pool for diffusion OPD

cr-gao:
### What does this PR do?

Brings verl's multi-teacher on-policy distillation design (`DistillationConfig`, `AsyncTeacherLLMServerManager`, `teacher_pool`) to the diffusion OPD runtime merged in #300 and #325.

- **Config.** `distillation.teacher_models` accepts several named entries, each with a routing `key` and, for a standalone pool, a `world_size`. A single-teacher config is unchanged and resolves to the key `default`. `distillation.teacher_key` (default `data_source`) names the batch column used for routing. `distillation.n_gpus_per_node` and `nnodes` define an optional `teacher_pool`; `nnodes: 0` (the default) keeps the teachers colocated with the actor, exactly as before.
- **Worker.** `ActorRolloutRefWorker` builds one frozen `TrainingWorker` per routing key (`self.teachers`), accepts `role="teacher"` for standalone groups, and `infer_teacher_batch` picks the teacher from a `teacher_key` carried in the TensorDict metadata. The call is `blocking=False`.
- **Trainer.** The new `DiffusionTeacherManager` is the diffusion counterpart of verl's `AsyncTeacherLLMServerManager`. It resolves routing keys with verl's exact error messages, sends each key's rows to its teacher group (padded to the group's DP size × its infer micro batch size so every rank's shard splits into micro-batches, all groups in flight before the first `.get()`), and scatters `teacher_prev_sample_mean` back in the original row order. The loss side is untouched.
- **Placement.** With `nnodes > 0`, `main_diffusion` registers `teacher_pool` / `Role.TeacherModel`, and the trainer splits the pool into one sub-pool per teacher (`split_resource_pool`, by `world_size`), each with its own worker group, so the teachers score concurrently. Teacher scoring remains a serial stage of the step, so a standalone pool buys memory isolation and distance from the rollout engine's sleep/wake cycle, not speed. Note: the default `trainer.ray_master_port_range` hands every spawned worker group the same rendezvous port, so multi-worker-group runs currently need `trainer.ray_master_port_range=null` (the standalone smoke sets it); the port-range fix, split out of this PR per review, makes the default work by slicing the range per group.

Not in scope, left for follow-ups: per-teacher `lora_adapter_path` / `guidance_scale`, a pure-distillation fast path, and the v1 trainer.

### Checklist Before Starting

[...]

PR #482, opened 2026-08-29
[trainer, cfg, tests] fix: give each ray worker group its own rendezvous port slice

cr-gao:
### What does this PR do?
`diffusion_trainer.yaml` defaults `trainer.ray_master_port_range` to `[20000, 30000]`, and verl's `get_master_addr_port` returns the first *currently free* port of the range at worker-group **spawn** time — the port is only really bound later, in `init_model`'s `init_process_group`. Both the v0 and v1 trainers spawn every worker group before any `init_model`, so with two or more worker groups every group gets port 20000 and the run dies with `EADDRINUSE`. #427's standalone `teacher_pool` is the first config that actually creates a second worker group here (the standalone reward model only reserves a `reward_pool` — its servers are `RolloutReplica`s, not a worker group), which is why the default never fired before.

Fix, keeping both properties:

- `diffusion_trainer.yaml`: `ray_master_port_range: null` by default (ephemeral ports, as upstream); the range is only for isolating concurrent training jobs on one machine.
- v0 (`ray_diffusion_trainer.py`) and v1 (`trainer_base.py`): when a range **is** configured, `worker_group_port_ranges` slices it into disjoint per-group sub-ranges, so CI's per-job `RAY_MASTER_PORT_RANGE` injection keeps working with any number of worker groups.
- Drops the temporary `trainer.ray_master_port_range=null` pin the standalone teacher smoke carried since #427 — with the default now `null` it is a no-op and its comment no longer describes the default.

### Checklist Before Starting

- [x] Duplicate check: open-PR searches for `port` / `master_port` / `EADDRINUSE` return no overlapping PR; this commit originally sat in #427, where the reviewer asked for it to be a separate PR.
- [x] PR title format `[trainer, cfg, tests] fix: ...`

### Test

CPU (`python -m pytest … -q`, verl 0.10.0.dev):

```
tests/trainer/diffusion/test_distillation_trainer_wiring_on_cpu.py   (TestWorkerGroupPortRanges: default null; null → no per-group range; slicing disjoint & in-range; range smaller than group count raises)
tests/trainer/diffusion/test_worker_batch_projection_on_cpu.py
→ 25 passed; full tests/trainer/diffusion → 229 passed
pre-commit run → all hooks pass (autogen-trainer-cfg skipped, known-broken locally)
```

GPU (4× RTX PRO 6000 Blackwell 96GB), `SMOKE=standalone NUM_GPUS=3` teacher smoke = 3 worker groups (actor + 2 standalone teachers):

- **Repro on current `main` (`783ea1f`)**: `trainer.ray_master_port_range=[20000,30000]` → dies at rendezvous with `DistNetworkError … port: 20000 … EADDRINUSE` (exit 1).
- **This branch, same command** → passes; `distill_kl_loss` bit-identical to the same smoke without a range.
- **This branch, tight range `[20000,20010]`** (3 groups slice 10 ports) → passes, bit-identical.
- **This branch, default (`null`, pin removed from the smoke)** → passes, bit-identical.
- **This branch, single worker group (`SMOKE=distill NUM_GPUS=1`) with `[20000,30000]`** → passes (whole range for the one group).

### Note on AI assistance

AI assistance (Claude) was used for this change. I reviewed every changed line and ran the tests above myself.

### DIFF
--- a/tests/special_e2e/run_diffusion_teacher_smoke.sh
+++ b/tests/special_e2e/run_diffusion_teacher_smoke.sh
@@ -119,8 +119,6 @@ case "${SMOKE}" in
             "${two_teachers[@]}"
             distillation.n_gpus_per_node=2
             distillation.nnodes=1
-            # The default fixed port range hands every worker group the same rendezvous port.
-            trainer.ray_master_port_range=null
         )
         objective=("${pure_distill[@]}")
         ;;
--- a/tests/trainer/diffusion/test_distillation_trainer_wiring_on_cpu.py
+++ b/tests/trainer/diffusion/test_distillation_trainer_wiring_on_cpu.py
@@ -111,3 +111,32 @@ class TestTeacherPoolRegistration:
         cfg, runner = self._runner(["distillation.n_gpus_per_node=0", "distillation.nnodes=1"])
         with pytest.raises(ValueError, match="config.distillation.n_gpus_per_node must be greater than 0"):
             runner.init_resource_pool_mgr(cfg)
+
+
+class TestWorkerGroupPortRanges:
+    def test_default_has_no_port_range(self):
+        assert compose_cfg([]).trainer.ray_master_port_range is None
+
+    def test_null_range_gives_no_range_per_group(self):
+        from verl_omni.trainer.diffusion.diffusion_trainer_utils import worker_group_port_ranges
+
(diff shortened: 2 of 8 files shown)


zhtmike:
thanks for the follow up.

zhtmike (approved):

Issue #207, opened 2026-06-24
[RFC] Design and Implementation Plan for HunyuanImage 3.0 Support


SamitHuang:
### 1. Motivation
`HunyuanImage 3.0` is an 80B native multimodal model (13B active parameters) that unifies multimodal understanding and generation within an autoregressive (AR) and diffusion framework. It supports multiple generation modes including `t2t`, `i2t`, `t2i`, `it2i`, and a novel `t2ti` mode (think-recaption chain: AR chain-of-thought followed by diffusion generation). 

Integrating `HunyuanImage 3.0` into `verl-omni` is a critical step for scaling our RL training capabilities to state-of-the-art MoE-based unified multimodal models. It will allow us to perform end-to-end RL (e.g., GRPO for the AR reasoning phase and Flow-GRPO for the DiT phase) on a single unified architecture. 

### 2. Current Problem
Currently, `verl-omni` supports single-modality or decoupled architectures (like Qwen-Image for diffusion and Qwen3/Qwen-VL for AR). Introducing `HunyuanImage 3.0` presents new architectural and infrastructural challenges:
1. **Unified Two-Engine Pipeline (AR + DiT)**: The model is neither purely a diffusion model nor purely an AR model. It requires a unified trainer and rollout loop that can orchestrate both the AR engine (for CoT reasoning/captioning) and the DiT engine (for pixel generation) within the same RL step.
2. **Scale & MoE Architecture**: At 80B total parameters (MoE), the model exceeds the memory limits of single-GPU and potentially single-node setups without aggressive offloading. We must ensure FSDP/Megatron (for training) and `vllm-omni` (for rollout) can handle MoE routing and distributed parallelism efficiently.
3. **Sequential Modality Generation (`t2ti`)**: Supporting the `t2ti` mode requires custom rollout orchestration where the output tokens of the AR phase directly condition the diffusion phase before the final reward is calculated.

### 3. Proposed Design & Changes
Extending `verl-omni` to feature a **Unified Model Trainer and Rollout mechanism**.

1. **Unified Model Pipeline**:
   - Create a new pipeline directory (e.g., `verl_omni/pipelines/unified_model/`) that orchestrates both AR and DiT RL components.
   - For `t2ti` rollouts, the agent will query the `vllm-omni` AR engine to generate the CoT text, and immediately pass the generated text state into the `vllm-omni` DiT engine to generate the image. The joint trajectory will be evaluated by the reward function.

2. **MoE and Multi-Node Support**:
[...]

PR #483, opened 2026-08-30
[rollout] fix: remap co-located ZMQ handle rank via CUDA_VISIBLE_DEVICES in multi-stage deploy

Xenoryn:
### What does this PR do?

With co-located training, LoRA weights bypass Ray serialization: each actor rank pushes its LoRA tensors directly to the vLLM worker on the same physical GPU over a per-rank ZMQ IPC socket. The socket path looks like:

```
ipc:///tmp/rl-colocate-zmq-{job_id}-replica-{replica_rank}-rank-{rank}.sock
```

**Where it breaks.** In a multi-stage deploy (e.g. HunyuanImage-3 pins the AR stage to GPUs 0-3 and the DiT stage to 4-7), vllm-omni narrows `CUDA_VISIBLE_DEVICES` for each stage's processes. The DiT workers therefore see a **stage-local** `local_rank` of 0..3, while the co-located actor ranks are the **node-local** 4..7. The two sides derive different socket names, with these consequences:

- actors 4..7 find no listener, so the send hangs or fails;
- the `rank-0` name is derived by two workers at once (the AR worker on GPU 0 and the DiT worker on GPU 4), so the second `bind` fails with an address-in-use error.

**The fix.** `_get_zmq_handle()` now resolves the rank through the worker's own `CUDA_VISIBLE_DEVICES`: `entries[local_rank]` recovers the node-local physical GPU index so both sides agree. For example, in a DiT stage process CVD is `"4,5,6,7"`, so the stage-local ranks 0..3 map back to 4..7.

Three cases fall back to the raw `local_rank`, keeping existing behavior: no CVD, rank beyond the CVD list, and non-numeric entries (ROCm device UUIDs / MIG slice IDs, which cannot be resolved).

**Zero impact on single-stage deployments**: with the whole engine visible, `entries[local_rank] == local_rank` — an identity mapping.

No API, config, or documentation changes.

Context (not a dependency): this is a prerequisite fix for multi-stage recipes such as the HunyuanImage-3 FlowGRPO work in RFC #207.

### Checklist Before Starting

- [x] Search for similar PRs. Paste at least one query link here: ...

### Test

CPU unit tests (5 new cases appended to `tests/workers/rollout/rollout_vllm/test_zmq_utils.py`):

```bash
python3 -m pytest tests/workers/rollout/rollout_vllm/test_zmq_utils.py -q
# 8 passed, 1 failed (all 5 new cases pass; the failure is the pre-existing one explained below)
```

| Case | Input | Expected |
| --- | --- | --- |
| stage-local → node-local | common multi-stage case: `CVD="4, 5, 6, 7"` + local_rank=2 (also covers the strip) | handle contains `rank-6` |
| identity mapping | single-stage: `CVD="0,1,2,3,4,5,6,7"` + local_rank=5 | handle contains `rank-5` |
| fallback 1 | CVD absent | raw rank kept |
| fallback 2 | rank beyond the CVD list: `CVD="0,1"` + local_rank=3 | raw rank kept |
| fallback 3 | non-numeric entries: `GPU-1234,GPU-5678` (ROCm UUID / MIG) | raw rank kept |

For the record: the failure (`test_worker_receivers_resolve_shared_update_id_from_their_rank_local_handles`, added in #349) is pre-existing and unrelated to this PR. It exercises the full-weight branch of `update_weights_from_ipc` and monkey-patches `_get_zmq_handle` away, so none of the code changed here runs — the failing lines are byte-identical on `main`.

### API and Usage Example

N/A

### Design & Code Changes

N/A

> [!IMPORTANT]

---

**AI assistance disclosure** (required by the contribution policy):

- AI assistance (Claude Code) was used for this change.
- The human submitter has reviewed every changed line and ran the tests listed above.


knlnguyen1802 (changes requested):
﻿The multi-stage mismatch diagnosis looks right, but remapping with `int(CVD[local_rank])` as the ZMQ rank conflates physical GPU id with the actor's node-local rank (`rollout_rank % RAY_LOCAL_WORLD_SIZE`).

That works for CVD=`0..7` + DiT=`4..7`, and regresses single-stage replicas whose server CVD is e.g. `4,5,6,7` (actor still uses 0..3). The HTTP server builds CVD from Ray physical accelerator ids, so offset lists are a real co-located path.

## Blocking

- **CVD absolute ints ≠ actor node-local rank when Ray CVD is offset** — `verl_omni/workers/rollout/vllm_rollout/utils.py` `_get_zmq_handle`
  Remap must yield the same node-local index the actor uses, not the raw CVD integer. Prefer `VERL_ZMQ_BASE_TRAINER_RANK` (stage start index within the replica), or preserve replica CVD and compute `replica_entries.index(stage_entries[local_rank])`. Add a unit test with CVD=`4,5,6,7` + `local_rank=0` expecting `rank-0`, not `rank-4`.

- **New tests never run in L1 CI** — `tests/workers/rollout/rollout_vllm/test_zmq_utils.py`
  CPU workflow only collects `*_on_cpu.py`. Rename (or move cases) so the regression test above is actually executed.

pls fix the remap to stay on node-local ranks and wire the tests into L1.

Xenoryn:
> The multi-stage mismatch diagnosis looks right, but remapping with `int(CVD[local_rank])` as the ZMQ rank conflates physical GPU id with the actor's node-local rank (`rollout_rank % RAY_LOCAL_WORLD_SIZE`).
> 
> That works for CVD=`0..7` + DiT=`4..7`, and regresses single-stage replicas whose server CVD is e.g. `4,5,6,7` (actor still uses 0..3). The HTTP server builds CVD from Ray physical accelerator ids, so offset lists are a real co-located path.
> 
> ## Blocking
> * **CVD absolute ints ≠ actor node-local rank when Ray CVD is offset** — `verl_omni/workers/rollout/vllm_rollout/utils.py` `_get_zmq_handle`
>   Remap must yield the same node-local index the actor uses, not the raw CVD integer. Prefer `VERL_ZMQ_BASE_TRAINER_RANK` (stage start index within the replica), or preserve replica CVD and compute `replica_entries.index(stage_entries[local_rank])`. Add a unit test with CVD=`4,5,6,7` + `local_rank=0` expecting `rank-0`, not `rank-4`.
> * **New tests never run in L1 CI** — `tests/workers/rollout/rollout_vllm/test_zmq_utils.py`
>   CPU workflow only collects `*_on_cpu.py`. Rename (or move cases) so the regression test above is actually executed.
> 
> pls fix the remap to stay on node-local ranks and wire the tests into L1.

@knlnguyen1802 Thanks! Both blocking issues addressed:                                                                                                                                                                                                           

- **Remap** now uses the replica CVD approach you suggested: the server stashes the replica-level device list (`VERL_ZMQ_BASE_VISIBLE_DEVICES`) before stage narrowing, and `_get_zmq_handle` computes  `replica_entries.index(stage_entries[local_rank])` — the raw physical id is never used as the rank. Added the requested regression (CVD=`4,5,6,7`, `local_rank=0` → `rank-0`), plus offset-replica and UUID cases. 
- **Tests** renamed to `test_zmq_utils_on_cpu.py` so L1 collects them: 13 passed.                                                                                                                                                         

Side note: renaming surfaced a pre-existing red test on main (the #349 fixture missing `model_runner.pipeline.load_weights` after the vllm-omni 0.26 contract change); fixed the fixture in the same move so the collected file is green — unrelated to this remap.

knlnguyen1802:
@Xenoryn Could you resolve conflict ?

Xenoryn:
> @Xenoryn Could you resolve conflict ?

@knlnguyen1802 Rebase done, please check again.

knlnguyen1802 (approved):

PR #478, opened 2026-08-29
[2/N][diffusion, rollout, tests] refactor: declare diffusion media outputs

NancyFyong:
## Summary

Part of the [#403 RFC](https://github.com/verl-project/verl-omni/issues/403), stacked on the merged PR1 generation-strategy split (#444). This is **PR2 — the typed diffusion output vertical slice**.

It moves the model-specific joint video/audio conventions out of the shared diffusion rollout strategy and into **adapter-owned declarations**, and makes **every** diffusion pipeline declare the modality it produces.

- Before: `DiffusionStrategy.process_output` hard-coded "audio lives at tuple position 1" **and** `extra_fields.setdefault("audio_sample_rate", 32000)` — a MiniMax-specific constant in shared code. Consumers infer image-vs-video from tensor rank (`is_video = outputs.ndim == 5`).
- After: adapters declare a `DiffusionIOSpec`; the strategy resolves the active pipeline's spec and takes the audio sample rate from it (or from runtime rollout metadata). **The 32 kHz literal is gone from the strategy.**

### Changes
- **New** `verl_omni/pipelines/rollout_media.py`: `MediaSpec` + `DiffusionIOSpec` (CPU-importable typed contracts).
- **Every registered diffusion pipeline declares `diffusion_io_spec`:**
  - `primary=image`: BAGEL, Boogu-Image, Qwen-Image (`flow_grpo`/`diffusion_nft`/`dpo`/`edit`; `dual_grpo`/`mix_grpo` inherit), SD3.
  - `primary=video`: Wan2.2 (single video).
  - `primary=video` + auxiliary `audio`: MiniMax H3 (`flow_grpo` 32000, `diffusion_nft` 32000), LTX-2 (24000; runtime vocoder rate takes precedence).
- `vllm_omni_diffusion_strategy.py`: `_diffusion_io_spec()` resolves the adapter-declared spec; audio sample rate is spec-driven; the model-specific 32 kHz default is removed.
- CPU characterization + completeness tests.

### Exit criterion (met)
Adding another combination of existing output modalities now changes an **adapter and its tests**, not the server or the strategy.

### Deferred (not in this PR)
- Typed `DiffusionOutput.primary/auxiliary` fields + `MediaOutput` container → a later PR, once a consumer exists (avoids landing unused code).
- Consumer migration to read the declared modality (and dropping `ndim == 5` inference in `ray_diffusion_trainer.py`) → PR4.
- Request-side `ImageGenerationRequest` generalization → PR3.

[...]

PR #488, opened 2026-08-31
[diffusion] fix: preserve Qwen-Image metadata in V1 TransferQueue

Sky-Trigger:
### What does this PR do?

This PR preserves condition-aware Qwen-Image `img_shapes` metadata across the
V1 TransferQueue trajectory boundary.

The Qwen-Image-Edit rollout adapter returns `img_shapes` as structured,
non-tensor metadata. The TransferQueue writer previously forwarded tensor
fields and selected reward metadata, but dropped this list before actor
training. The training adapter therefore could not reliably restore the target
and condition image shapes required by Qwen-Image 2D RoPE.

The fix adds `img_shapes` to a narrow allowlist and carries it through the
existing `extra_fields` envelope. The existing TransferQueue reader already
unpacks that envelope into `DataProto.non_tensor_batch`, so no wire-format or
public API change is required. Arbitrary rollout metadata remains filtered.

### Checklist Before Starting

- [x] Searched for similar PRs: [open PRs matching `TransferQueue img_shapes`](https://github.com/verl-project/verl-omni/pulls?q=is%3Apr+is%3Aopen+TransferQueue+img_shapes).
  - No matching open PR was found.
  - [PR #478](https://github.com/verl-project/verl-omni/pull/478) declares diffusion media modalities at the rollout adapter/strategy boundary. It does not modify the TransferQueue writer, the non-tensor `extra_fields` envelope, or `img_shapes` restoration, so it does not overlap with this fix.
- [x] PR title follows `[{modules}] {type}: {description}`:
  `[diffusion] fix: preserve Qwen-Image metadata in V1 TransferQueue`.

### Test

Focused CPU tests were run with `/data/8T/wangmb/LLM-Infra/myenv`, the local
`verl` and `vllm` source checkouts, and the `vllm-omni` checkout pinned by the
current `verl-omni` main branch (`444485650b19b792403b12976f1b0eeb2ac1451c`):

```bash
PYTHONPATH=<this-verl-omni-checkout>:/data/8T/wangmb/LLM-Infra/verl:/data/8T/wangmb/LLM-Infra/vllm:<pinned-vllm-omni-checkout> \
    /data/8T/wangmb/LLM-Infra/myenv/bin/python -m pytest -q \
    tests/pipelines/test_qwen_image_edit_on_cpu.py
```

Result:

```text
28 passed, 29 warnings in 11.48s
```

Additional checks:

```text
python3 -m py_compile <three changed Python files>  # passed
git diff --check                                    # passed
```

Full `pre-commit` and Ruff were not run because those tools are not installed
in the available environment.

### API and Usage Example

No public API or configuration changes.

Qwen-Image-Edit V1 rollouts continue to emit `img_shapes` through the existing
adapter output. The transport now preserves it automatically for actor
training.

### Design & Code Changes

- Add `img_shapes` to an explicit non-tensor trajectory metadata allowlist in
  `DiffusionAgentLoopWorkerTQ`.
- Store allowlisted metadata inside the existing per-row `extra_fields`
  envelope without forwarding unrelated values.
- Verify that the TQ writer preserves `img_shapes` and continues to reject
  unrelated metadata.
- Verify that `diffusion_tq_batch_to_dataproto()` restores the per-sample
  metadata into `DataProto.non_tensor_batch` and its TensorDict view.
- Strengthen the Qwen-Image-Edit condition-injection test by checking that the
  packed hidden-state length matches the target-plus-condition `img_shapes`.

### DIFF
--- a/verl_omni/agent_loop/diffusion_agent_loop_tq.py
+++ b/verl_omni/agent_loop/diffusion_agent_loop_tq.py
@@ -35,6 +35,9 @@ from verl_omni.agent_loop.utils import _derive_rollout_seed
 logger = logging.getLogger(__name__)
 logger.setLevel(os.getenv("VERL_LOGGING_LEVEL", "INFO"))
 
+# Structured model metadata is carried inside the existing non-tensor envelope.
+_NON_TENSOR_TRAJECTORY_FIELDS = frozenset({"img_shapes"})
+
 
 @ray.remote
 class DiffusionAgentLoopWorkerTQ(DiffusionAgentLoopWorker):
@@ -234,6 +237,9 @@ class DiffusionAgentLoopWorkerTQ(DiffusionAgentLoopWorker):
 
         reward_extra_info = extra.get("reward_extra_info")
         extra_fields_out: dict[str, Any] = {}
+        for metadata_key in _NON_TENSOR_TRAJECTORY_FIELDS:
+            if metadata_key in extra:
+                extra_fields_out[metadata_key] = extra[metadata_key]
         if reward_extra_info is not None:
             extra_fields_out["reward_extra_info"] = reward_extra_info
         # Track the rollout model version this trajectory was generated against.
--- a/verl_omni/agent_loop/diffusion_agent_loop_tq.py
+++ b/verl_omni/agent_loop/diffusion_agent_loop_tq.py
@@ -35,9 +35,6 @@ from verl_omni.agent_loop.utils import _derive_rollout_seed
 logger = logging.getLogger(__name__)
 logger.setLevel(os.getenv("VERL_LOGGING_LEVEL", "INFO"))
 
-# Structured model metadata is carried inside the existing non-tensor envelope.
-_NON_TENSOR_TRAJECTORY_FIELDS = frozenset({"img_shapes"})
-
 
 @ray.remote
 class DiffusionAgentLoopWorkerTQ(DiffusionAgentLoopWorker):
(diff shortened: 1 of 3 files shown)


zhtmike:
@knlnguyen1802

verl_omni/agent_loop/diffusion_agent_loop_tq.py
```diff
@@ -35,6 +35,9 @@
 logger = logging.getLogger(__name__)
 logger.setLevel(os.getenv("VERL_LOGGING_LEVEL", "INFO"))
 
+# Structured model metadata is carried inside the existing non-tensor envelope.
+_NON_TENSOR_TRAJECTORY_FIELDS = frozenset({"img_shapes"})
```
zhtmike: Make it inline if it is only used once.

zhtmike (approved):

verl_omni/agent_loop/diffusion_agent_loop_tq.py
```diff
@@ -234,6 +234,8 @@ async def _write_trajectory_to_tq(
 
         reward_extra_info = extra.get("reward_extra_info")
         extra_fields_out: dict[str, Any] = {}
+        if "img_shapes" in extra:
+            extra_fields_out["img_shapes"] = extra["img_shapes"]
```
zhtmike, line 238: I think it should be a formal fix, e.g., assign all `extra` to `extra_fields_out` later? @Sky-Trigger 

cc @knlnguyen1802

Sky-Trigger: ok，pls wait a minute

knlnguyen1802 (approved):

verl_omni/agent_loop/diffusion_agent_loop_tq.py
knlnguyen1802: > I think it should be a formal fix, e.g., assign all `extra` to `extra_fields_out` later? @Sky-Trigger
> 
> cc @knlnguyen1802

I'll do formal fix after migrate all to v1, seem there is some field separately depend on model

zhtmike: good

Sky-Trigger: Ok, then I won’t make a general fix for this for now.

PR #493, opened 2026-08-31
[trainer, recipe, tests, doc] feat: multi-teacher OPD on the v1 sync diffusion trainer

cr-gao:
### What does this PR do?

Ports the merged diffusion OPD teacher runtime (#325/#427) onto the v1 sync trainer, closing the "OPD still lives on v0" gap tracked in #389. Adds the v1 MOPD recipe, a `V1=1` switch on the teacher e2e smoke, and CPU wiring tests mirroring the v0 suite. `separate_async` + distillation raises `NotImplementedError` until benched (planned follow-up: teacher scheduling in separate_async).

Also bundled: the OPD example recipes now override `actor_rollout_ref.ref.fsdp_config.model_dtype=bfloat16` — teachers derive their engine config from `ref`, whose fp32 default streams 2× parameter bytes per forward under the forward-only CPU offload policy while compute is bf16 either way (FSDP mixed precision). Measured: teacher stage 7.76s → 3.13s standalone (−60%, together with the micro-batch default), 4.35s → ~2.3s colocated; `distill_kl` trajectories under fp32 vs bf16 teachers agree per step over 10 steps (median 0.25%, max 0.85%).

### Why this is not duplicating an existing PR

#389 lists diffusion OPD on v1 as an open gap; `gh pr list --state open --search "v1 distill"` and `--search "OPD"` show no open PR on this path. Claimed on #389 before starting.

### Tests

- CPU (RED verified before implementation): `pytest tests/trainer/diffusion/test_v1_distillation_wiring_on_cpu.py` → 7 passed; regression across v1 + distillation suites → 65 passed.
- E2E smoke matrix (4× RTX PRO 6000): `V1=1 SMOKE={distill,coexistence,mopd,standalone}` + `V1=0 SMOKE=mopd` control → 5/5 PASS; v1 colocated-vs-standalone routing gives identical `distill_kl` to 4 decimals (the #427 equivalence, reproduced on v1).
- Real-model parity, SD3.5-M two-teacher recipe, 100 steps + val each: loss and val-reward curves overlap (final val OCR 0.8505 v0 vs 0.8469 v1, PickScore 0.8015 vs 0.7995); per-stage timings at parity (teacher 2.38 vs 2.33s, update 5.14 vs 4.91s).

v0 vs v1 curves

AI assistance (Claude Code) was used for this change; the submitting human has reviewed every changed line and run the tests above.

Issue #389, opened 2026-08-14
[RFC] Tracking v1 upgrade for verl-omni v0.3.0: remaining recipe coverage


knlnguyen1802:
### Feature request

This is the **v0.3.0 dashboard** for migrating remaining verl-omni models and algorithms onto the **verl v1 trainer** (TransferQueue + ReplayBuffer), and for proving each migrated path is **performance-competitive** and **behavior-compatible** with the current v0 recipes.

It does **not** redesign the v1 trainer. Architecture RFCs already exist:

| Area | RFC | Status |
|------|-----|--------|
| Diffusion v1 trainer (`sync` / `separate_async`) | #249 | Done in #296, #315 |
| Omni model base + v1 sync trainer | #232 | Done in #253, #258, #312 |
| Omni v1 `colocate_async` / `separate_async` | #320 | `separate_async` done in #380; `colocate_async` still in progress (#337); related pin/lifecycle: #491, landed in #497 |
| Diffusion OPD one-step-off teacher scheduling | #498 | Done in #495 |

**Please attach the corresponding PR in the tables below** when you start or land a (model, algorithm, mode) path. Keep this issue as the single dashboard. Only PRs against `verl-project/verl-omni` count.

Cross-reference (vllm-omni rollout, not a verl-omni v1 trainer blocker): [vllm-project/vllm-omni#778](https://github.com/vllm-project/vllm-omni/issues/778)

### Motivation

v1 is the long-term trainer control plane (aligned with upstream verl): TransferQueue, ReplayBuffer, agent-loop workers, and `sync` / `colocate_async` / `separate_async` modes.

The **architecture is in**. For **verl-omni v0.3.0**, the remaining work is **flagship recipe coverage on v1 `sync`**, not new trainer architecture. Several production recipes still launch the legacy v0 trainer (`main_diffusion`). Until those v0.3.0 models have a v1 recipe, we cannot treat v1 as the release default.

v1 is still **opt-in**. Diffusion still defaults to `trainer.use_v1=false`.

### Dashboard (updated 2026-09-11)

**v0.3.0 focus:** land remaining flagship diffusion recipes on v1 `sync`. Omni async trainers and extra algorithm ports do **not** block this list.

## Done

Merged on `verl-project/verl-omni`. These paths have a first-class v1 recipe or trainer.

| Stack | Path | Mode | Recipe / entrypoint | PR |
|-------|------|------|---------------------|----|
[...]

PR #495, opened 2026-08-31
[trainer, tests, doc] feat: async teacher scheduling for diffusion OPD on v1 separate_async

cr-gao:
### What does this PR do?

Async teacher scheduling for diffusion OPD on the v1 `separate_async` trainer, implementing RFC #498 (#493 is merged, so this PR carries only this work — 6 commits):

1. **Inline enablement**: drop the sync-only distillation gate; teachers run inline in the separate_async training phase; the teacher e2e smoke gains `V1_MODE=separate_async`.
2. **One-step-off scheduler** (`distillation.scheduler=one_step_off`): teacher scoring of batch k overlaps the actor update on batch k−1. `DiffusionTeacherManager` splits into dispatch/collect, the trainer keeps a one-batch pipeline, and the residual wait surfaces as `timing_s/wait_prev_teacher`. Valid only with standalone teacher pools on separate_async; requires `num_warmup_batches >= 2` (the pipeline fill consumes one batch of generation lead — with less, gen wait balloons 2.9s → 14.1s, measured).
3. The inline schedule value is named `inline` (not `sync`) to avoid colliding with `trainer_mode=sync`.
4. From review: docs for the scheduler knob, and two separate_async teacher smokes (inline and one_step_off) in the diffusion GPU smoke group.

### Measured

All runs SD3.5-M, 64 samples/step, 384², same seeds; `num_warmup_batches=2` on both schedulers and no evictions on either side, so every comparison is single-variable.

**Shipping recipe geometry** — dual-teacher MOPD (OCR + Aes on a 2-GPU pool, `parameter_sync_step=2`, 100-step runs): step 22.1s → **19.0s (−13.8%)**, teacher stage 3.6s → 0.02s residual wait. Behavior parity over the 100-step curve pairs — PickScore pointwise deltas ≤ 0.004, distill_kl pointwise median diff 2.4% with equal endpoints, val OCR within val noise across seeds:

parity curves

**Operating-point analysis** — the saving equals the teacher stage's share of the step, which scales with rows per teacher GPU. A single standalone teacher scoring the full batch on one GPU is the upper end (12-step runs):

| | inline | one_step_off |
|---|---|---|
| GS=1.0 step | 20.2s | **16.9s (−16.7%)** |
| GS=4.5 step (official operating point) | 37.6s | **28.7s (−23.7%)** |
| teacher stage → residual wait | 4.9s / 9.4s | 0.01s / 0.02s |

The inline baseline runs with `max_off_policy_threshold=3`; at the default 2 it evicts its oldest batch every other step and reads 21.9s / 38.0s. Expect roughly −14% to −24% end-to-end depending on how loaded the teacher pool is; the shipping dual-teacher recipe measures −13.8%.

Off-policy semantics: the teacher is frozen, so delayed scoring adds no teacher staleness. With the same generation lead both schedulers update on the same batch at the same step and report the same `trajectory_staleness`; one_step_off only samples it one step earlier. `max_off_policy_threshold` is applied at sampling time, so at the same threshold the oldest batch one_step_off can update on is one step older than under inline (RFC #498).

### Known observation (pre-existing, disclosed for reviewer context)

While validating this PR we observed that separate_async training can trail the v1 sync trainer's val-OCR curve on the same recipe at one seed (endpoint 0.747 vs 0.847, seed 42), while at another seed the early curve recovered into the sync trainer's envelope — so the effect is at least partly seed/data-order, with single-seed evidence on each side. It affects **both schedulers in this PR equally** (the single-variable parity pairs isolate the scheduler from it) and predates this PR; candidates under separate investigation are the TransferQueue batch-mixture order vs the v0 dataloader and the warm-up generation lead. Should read this PR as: OPD on separate_async inherits separate_async's training dynamics, whatever they turn out to be.

### Why this is not duplicating an existing PR

Continues the #389-tracked OPD-on-v1 line from #493 per RFC #498; no open PR touches v1 diffusion distillation.

### Tests

- CPU (TDD throughout, RED verified before each implementation step): distillation/v1 suites all green (120 passed on the rebased head).
- E2E on 5× RTX PRO 6000: the two new smoke-group cases as wired (4 GPUs, `wait_prev_teacher` 0.003s under one_step_off); 12-step sanity pairs and 100-step curve pairs at both geometries (numbers above); colocate-RM geometry check (a 10s blocking reward stage leaves the teacher fully hidden).
- The colocate-RM check surfaced a pre-existing v1 bug (colocated reward output replacing the batch); fixed in #513.

AI assistance (Claude Code) was used for this change; the submitting human has reviewed every changed line and run the tests above. All numbers were additionally re-verified by an independent adversarial review pass against the raw logs.


docs/algo/diffusion_opd.md
```diff
@@ -172,8 +172,7 @@ teacher stage time without changing the scores.
 
 Supported scope: the policy-gradient trainer with online sampling and
 FSDP/FSDP2 engines, on both the legacy trainer (`main_diffusion`) and the v1
-sync trainer (`main_diffusion_v1` with `trainer.v1.trainer_mode=sync`; the
-v1 `separate_async` mode does not support distillation yet). Unsupported
+trainer (`main_diffusion_v1`, `sync` and `separate_async` modes). Unsupported
```
knlnguyen1802, line 206: pls also document the new knob here, not only that `separate_async` is now allowed:

- `distillation.scheduler=inline` (default) vs `one_step_off`
- `one_step_off` needs standalone teachers + `num_warmup_batches >= 2` (YAML default is 1, so a bare override fails at startup)
- residual wait is `timing_s/wait_prev_teacher`; `trajectory_staleness` is still sample-time (effective age = measured + 1)

YAML comments are not enough for a user-facing scheduler.

tests/special_e2e/run_diffusion_teacher_smoke.sh
```diff
@@ -135,7 +142,19 @@ entrypoint=verl_omni.trainer.main_diffusion
...
+    v1_flags=(trainer.use_v1=true "trainer.v1.trainer_mode=${V1_MODE}")
+    if [[ "${V1_MODE}" == "separate_async" ]]; then
+        v1_flags+=(
+            actor_rollout_ref.rollout.mode=async
+            actor_rollout_ref.rollout.calculate_log_probs=True
+            actor_rollout_ref.rollout.nnodes=1
+            actor_rollout_ref.rollout.n_gpus_per_node=1
+            actor_rollout_ref.rollout.checkpoint_engine.backend=nccl
+            "actor_rollout_ref.actor.ppo_mini_batch_size=${train_batch_size}"
+            trainer.v1.separate_async.num_warmup_batches=0
```
knlnguyen1802: smoke script accepts `V1_MODE=separate_async` but `run_gpu_smoke_diffusion_e2e.sh` still only runs v1 sync. pls add something like:

```bash
run_test 11 "Diffusion OPD v1 separate_async standalone teachers e2e" \
    env ... V1=1 V1_MODE=separate_async SMOKE=standalone \
```

4-GPU layout already fits (1+1+2). A `one_step_off` + `num_warmup_batches=2` case would actually hit `wait_prev_teacher`; this script currently forces warmup=0, which that scheduler rejects.

verl_omni/trainer/diffusion/v1/trainer_base.py
```diff
@@ -306,25 +306,53 @@ def step(self, metrics: dict, timing_raw: dict) -> KVBatchMeta:
...
-        data = diffusion_tq_batch_to_dataproto(batch_meta, pad_token_id=self.tokenizer.pad_token_id or 0)
+        if data is None:
+            data = diffusion_tq_batch_to_dataproto(batch_meta, pad_token_id=self.tokenizer.pad_token_id or 0)
 
         # [OPTIONAL] colocated reward model
         if self.reward_loop_manager.reward_loop_worker_handles is None and self.use_rm:
             with marked_timer("reward", timing_raw, color="yellow"):
                 self.checkpoint_manager.sleep_replicas()
-                data = self._compute_reward_colocate(data)
+                data = data.union(self._compute_reward_colocate(data))
```
knlnguyen1802, line 355: this `union` is the #513 cherry-pick. Since #513 is merged, can drop this

knlnguyen1802 (commented):
A few nits on docs, smoke wiring, and the #513 cherry-pick. Scheduler + fail-closed validation look good.

docs/algo/diffusion_opd.md
cr-gao: Added a distillation.scheduler section with both values and the four startup checks, a one-step-off usage example, and timing_s/wait_prev_teacher under Metrics. Also I re-measured the staleness, pls review the conclusion: with the same lead and no evictions both schedulers train the same batch at the same step and read the same trajectory_staleness. The earlier 0-vs-1 came from max_off_policy_threshold=2 evicting inline's oldest batch. The only interaction is that the threshold applies at sampling, so one_step_off's oldest trainable batch is one step older at the same threshold. Docs and the PR body now say that.

verl_omni/trainer/diffusion/v1/trainer_base.py
cr-gao: Rebased and dropped that commit.

tests/special_e2e/run_diffusion_teacher_smoke.sh
cr-gao: Added test 11 separate_async inline and test 12 one-step-off. 12 should raise warmup to 2. Also both takes about 2 mins which raise the diffusion group by 5mins, let me know if you'd rather keep only 12.

zhtmike (approved):

zhtmike:
All CI green. Thanks for the work.

Issue #492, opened 2026-08-31
[RFC]: Bump vLLM-Omni Pin to `ded89346` and Properly Fix the Async Rollout Server Lifecycle


zhtmike:
Status: **ready to implement (1A) after step 0** — #244 hold acknowledged + #445 tracker updated (§2). 1A = frozen-SHA bump + compat + the both-hook resume bridge (the sync-path hang is latent in 1A and live from 1B's delegation — §5.3; the bridge ships now so 1B can never land a hang). 1B's lifecycle design is locked (abort-then-pause, wrap-four, drain design, ACK validation, `level=1`, keyword `tags=`, no server self-resume) and lands immediately after 1A — as a tight pair, since the pin-file push auto-fires all smoke groups including diffusion separate-async (§7). All citations verified against the frozen SHA, `verl@fefb080`, and live GitHub state.
Audience: the implementation PR. This is the design, not a roadmap.
Umbrella: #491 (PR 1 of 3; PRs 2–3 are the omni async trainers and are **out of scope here**).
Related: [#433](https://github.com/verl-project/verl-omni/issues/433), [#377](https://github.com/verl-project/verl-omni/issues/377), [#388](https://github.com/verl-project/verl-omni/issues/388), [#445](https://github.com/verl-project/verl-omni/issues/445) (P8/P9/P10 apply; P7 already done), [#403](https://github.com/verl-project/verl-omni/issues/403) + open #478/#480/#481 (seam freeze, §2), [#460](https://github.com/verl-project/verl-omni/pull/460) (level-1 constraint), [#363](https://github.com/verl-project/verl-omni/pull/363) (previous bump), [#452](https://github.com/verl-project/verl-omni/pull/452) (sequenced after [vllm-omni#6645](https://github.com/vllm-project/vllm-omni/pull/6645), open non-draft).

---

## 1. Goal

Bump `vllm-omni` from `444485650b19b792403b12976f1b0eeb2ac1451c` (v0.27-era, 218 commits behind) to **frozen snapshot** `ded8934626aaad1a3e816c3a1d9d742efc012d93` (2026-08-31 01:10Z — *not* "`main`", which has later commits; re-check the pin the morning the PR opens), and replace the async-server workaround stack with a proper lifecycle fix that follows verl's **trainer-visible contract** — not verl's Python calls, because `AsyncOmni` is not `AsyncLLM` (§5).

Exit criteria (all):

[...]

Issue #491, opened 2026-08-31
[RFC]: Restore Async Omni Training via the vLLM-Omni Pin Bump (Colocate & Separate Async)


zhtmike:
Status: **PR 1 landed — ready to implement PR 2 / PR 3** (updated 2026-09-03): 1A + 1B landed together as [#497](https://github.com/verl-project/verl-omni/pull/497) (frozen pin `ded89346`, merged 2026-09-03; sub-RFC #492 closed "Done in #497"); 1C executed/waived (§8). Remaining: PR 2 (`omni_colocate_async`, rebase #337) and PR 3 (`omni_separate_async`, rebase #380) — corrected scope, lifecycle, and gates are tracked in the updated [#320](https://github.com/verl-project/verl-omni/issues/320) (2026-09-03); coordinate resume ownership with [#517](https://github.com/verl-project/verl-omni/issues/517) (§3.2). Shipped deviations from the sub-RFC are recorded in §3.2.
Audience: maintainers sequencing the PR stack. Implementation detail lives in the sub-RFC; this document tracks scope, order, and gates.
Sub-RFC: #492 (PR 1 — pin bump + async-server proper fix; closed 2026-09-03 as done in #497 — the planned 1A→1B pair landed as a single PR).
[...]

PR #486, opened 2026-08-30
[1/N][diffusion, rollout] feat: support HunyuanImage-3.0 Flow-GRPO training

Xenoryn:
Close https://github.com/verl-project/verl-omni/issues/207

### What does this PR do?

Brings HunyuanImage-3.0-Instruct into verl-omni as a pipeline package. It is an 80B unified AR+DiT MoE transformer whose checkpoint is in transformers (non-diffusers) format, and whose rollout is a native two-stage vllm-omni pipeline (AR stage samples think/recaption text → DiT stage produces the image). This PR delivers four pieces:

1. **Training-side model definition** `HunyuanImage3ForTraining` — FSDP2-compatible with layer-level sharding, generation backbone ported from Tencent's official `modeling_hunyuan_image_3.py`;
2. **FlowGRPO training adapter** — the text condition at training time matches the rollout's sampled condition token-for-token (see below), CFG via batch doubling aligned with vllm-omni;
3. **Rollout adapter + agent loop** — both sides construct the same scheduler so the sigma schedule matches to the bit (prerequisite for an unbiased importance-sampling ratio); the agent loop builds the official AR prompt client-side through `prompt_utils` (the checkpoint ships no chat template);
4. **Two small shared-code changes** — the vllm-omni async server forwards AR sampling metadata through a whitelist; throughput metrics gain token counts. Both are backward compatible (details below).

**Scope boundary (stated up front):**
- This PR alone cannot run end-to-end training. The AR segment only samples during rollout and receives no gradient here; joint training lands in [2/N].
- End-to-end additionally requires: [the [1/N] multi-stage weight-sync ZMQ rank fix](https://github.com/verl-project/verl-omni/pull/483), two small upstream vllm-omni fixes [(#6768 sampling metadata](https://github.com/vllm-project/vllm-omni/issues/6768), [#6770](https://github.com/vllm-project/vllm-omni/issues/6770)), and the config / data-preprocessing scripts from the follow-up recipe PR.
- At the code level this PR depends on none of the above commits / upstream patches: it references nothing unmerged, degrades gracefully when metadata is absent, is self-contained for review, and does not affect existing models.

### Checklist Before Starting

- [x] Search for similar PRs. Paste at least one query link here: ...
- [x] Format the PR title as `[{modules}] {type}: {description}` (This will be checked by the CI)
[...]

PR #485, opened 2026-08-30
[diffusion, rollout, trainer, data, cfg, recipe, tests] feat: add MiniMax H3 Ref2VA FlowGRPO

NancyFyong:
### What does this PR do?

Implement the **FlowGRPO portion of #458** for MiniMax H3 Ref2VA, including multiple reference images, reference videos with or without soundtracks, and standalone reference audio.

The implementation:

- preserves the Agent Loop's original prompt token IDs and delegates official Ref2VA presentation/media preprocessing to vLLM-Omni without decode/re-tokenize;
- keeps all reference video/image rows fixed at timestep `0.999` and reference-audio rows fixed at `1.0`;
- stores only generated target video/audio states and target transition density in FlowGRPO trajectories;
- transports fixed condition rows and compact `ref_block_meta` once, then reconstructs the official packed layout during Actor replay;
- reuses the merged Ref2VA Agent Loop transport and variable reference-row padding while retaining explicit row counts;
- routes Ref2VA full-weight and LoRA synchronization to `transformers_ref`, reuses the merged LoRA placement fix, and uses the pinned vLLM-Omni resident audio-VAE path under rollout offload;
- reuses the merged multimodal converter through a recipe symlink and adds the Ref2VA launcher, documentation, CPU parity/contracts, and configurable train/validation reference-image short edges.

This PR is intentionally limited to `flow_grpo`; `flow_dppo`, `grpo_guard`, and the TransferQueue-specific variable-reference path are not supported.

### Relationship to existing work and non-duplication

- Implements the policy-gradient FlowGRPO milestone from #458.
- #448 implements **FL2VA FlowGRPO** and leaves Ref2VA out of scope.
- #464 is merged and provides the shared full-multimodal Ref2VA DiffusionNFT transport, layout, padding, and runtime foundations. This PR reuses those foundations and adds target-only stochastic FlowGRPO replay.
- Open-PR searches found no other Ref2VA FlowGRPO implementation:
  - https://github.com/verl-project/verl-omni/pulls?q=is%3Apr+is%3Aopen+%22Ref2VA+FlowGRPO%22
  - https://github.com/verl-project/verl-omni/pulls?q=is%3Apr+is%3Aopen+458+in%3Abody

This branch is updated through current `main` (including #464) and keeps only the FlowGRPO-specific delta.

### Checklist Before Starting

- [x] Search for similar PRs. Queries and scope comparison are listed above.
- [x] Format the PR title as `[{modules}] {type}: {description}`.

### Test

[...]

PR #497, opened 2026-08-31
[BREAKING][vllm_omni, rollout, trainer, ci, docker, diffusion, model, tests] fix: bump vllm-omni pin to 0.28 and fix the async rollout server lifecycle

zhtmike:
### What does this PR do?

Implements #492 (umbrella #491): bump `vllm-omni` to the frozen snapshot `ded89346` (vLLM 0.28) and replace the async-server workaround stack with the engine's real lifecycle — abort-then-pause, ACK-validated sleep/wake, and a rollout-side admission resume. Design and rationale live in the RFC; two logical parts: (1) pin bump + compat + bridge, (2) lifecycle rework.

Deviations from the RFC, on purpose:

- Python floor `>=3.11` + CI 3.12 (the pin imports `enum.StrEnum`), kernels `==0.16.0` everywhere plus a `[tool.uv] override-dependencies` entry so the standalone git-pin install cannot downgrade it (#417), transformers floor `>=5.13.0` (below 5.13 the `hub_kernels` mapping breaks `import vllm` with kernels 0.16.0), diffusers floor-only `>=0.40.0` (vllm-omni pins `==0.40.0`), `audioread` declared in the `[audio]` extra (qwen-omni-utils imports it undeclared; the failure silently emptied AVQA to 0 samples), #388 B4 probes reverted as out of scope (follow-up under #388).
- `abort_request` aborts one id without pausing the engine; any dict ACK fails `_validate_acks`; `wait_for_requests_to_drain` stays the upstream stub (the engine's `_admitting` gate is the sleep-time safety). `abort_request` also ignores `reset_prefix_cache` — the engine's cache resets are upstream stubs and the real clear rides `pause_generation`, which a per-request abort must not trigger; revisit when the async trainers add a caller.
- Known limit: multi-stage AR abort (thinker+talker) is broken **upstream** — the engine drops its stage-0 abort fallback in the consume loop. Single-stage/thinker-only rollouts (the registered smoke and the follow-up async trainers #337/#380) are correct; full multi-stage needs the vllm-omni fix + pin bump (TODO at the abort site).
- The `fault_tolerance_config` strip dropped as dead code in 1A is load-bearing at the 0.28 pin — `asdict()` serializes vLLM's default `FaultToleranceConfig` instance into a dict that `__post_init__` reads as an explicit `--fault-tolerance-config`, killing every rollout server at startup. Restored with a drop-condition TODO and pinned by a CPU contract test.
- The frontend multimodal cache is cleared through the renderer after each ACK-validated sleep **and after a successful abort pause with `reset_prefix_cache`**: `EngineCore.sleep(level>=1)` and `pause_generation(clear_cache=True)` wipe the engine-side cache, but the engine's own `AsyncOmni.reset_mm_cache()` no-ops on a missing attribute, so hash-only follow-up requests finished empty with `FinishReason.ERROR` (silent zero-length training responses). This mirrors vLLM's `AsyncLLM` (same desync fixed upstream in [vLLM#43001](https://github.com/vllm-project/vllm/pull/43001); verl disabled the cache on old vllm in [verl#6932](https://github.com/verl-project/verl/pull/6932)) and is droppable once vllm-omni fixes the lookup (TODO at the sites).
- Admission is resumed after every ACK-validated server-side `wake_up` (weight sync, `release/resume_kv_cache`): `AsyncOmni.sleep()` sets an admission hold its own `wake_up()` keeps, so without this only `omni_sync` (trainer bridge) could generate after sleep/wake. This restores the upstream contract — vLLM's `AsyncLLM` holds nothing across sleep/wake and verl's sync trainers never resume there — so every AR trainer works unmodified. The abort path still holds until the trainer's explicit `resume_generation_replicas`; the `omni_sync` bridge stays as an init-time safety net; diffusion engines are unaffected (no-op without AR stages). Engine-side fix tracked by TODO (long).
- diffusers 0.40 made `FlowMatchEulerDiscreteScheduler.set_timesteps` ignore an explicit `timesteps=` argument (stores `sigmas * 1000`), which broke the BAGEL replay's exact-value timestep lookup (the ci-e2e-diffusion IndexError). `setup_bagel_sigmas` restores the normalized grid from the stored sigmas, pinned by a CPU contract test. Only BAGEL passes `timesteps=`; all other pipelines use the standard path and were never affected.
- Install docs note the omni actor's `flash_attention_2` default (verl's LLM practice) requires the local `flash-attn` package, pointing at verl's install docs.

#445: P4/P5/P6 ✅, P8 ✅ ([vllm-omni#4092](https://github.com/vllm-project/vllm-omni/pull/4092) verified in-pinned), P9 kept; fault-tol re-ledgered as a live patch (blocked on upstream). #337's server hunk superseded; #403 seams untouched.

### Checklist Before Starting

- [x] Searched for similar PRs (`492 in:body`): only held draft #244 (P4/P5/P6 carried here per #445); nothing else touches the pin or the lifecycle.
- [x] Title validated against `tests/special_sanity/check_pr_title.py`.

### Test

```
# python 3.12, vllm==0.28.0+cpu, vllm-omni@ded89346, verl@fefb080, diffusers==0.40.0
TORCH_COMPILE_DISABLE=1 TORCHINDUCTOR_DISABLE=1 pytest -o python_files="*_on_cpu.py" -q --asyncio-mode=auto tests/
pytest -q tests/special_sanity
pre-commit → all Passed
```

CPU tests: the omni_sync resume-bridge ordering; the server lifecycle matrix (abort-before-pause, non-empty partial tokens, ACK validation, enqueue-then-raise, timeouts, per-request abort), including real-`AsyncOmni` sleep/wake contracts, the frontend mm-cache clear on sleep and on abort pause, and the wake-time admission resume (skipped on failed wake ACKs); the BAGEL scheduler grid contract (fails on diffusers 0.40 without the restore); and the `run_server` engine-args contract. Focused re-runs after the review fixes: bagel contract 6/6 (5/6 fail pre-fix), lifecycle + bridge 31/31, `tests/trainer/omni` + `tests/workers/rollout` 107/107.

GPU (AVQA GSPO, thinker LoRA): 240-step run with zero empty responses, training quality and rollout↔trainer logprob correlation at v0.2.0 parity, 10–12% higher throughput ([run](https://wandb.ai/verl-omni/gspo/runs/ta9g0ysq) vs [v0.2.0 baseline](https://wandb.ai/verl-omni/gspo_demo/runs/kzzrq9pr)). The colocated per-step weight sync never touches the frontend sleep/wake, so the review fixes are inert on this flow (same for SD3.5 / Qwen-Image: standard scheduler path, no abort-time behavioral change). Merge-blocking per RFC §6: `ci-e2e-omni` + `ci-e2e-diffusion` green; CUDA-free server PID after wake; #6442 prefix signature with `enable_prefix_caching=True`. FA3 engagement is 1C.

### Design & Code Changes

35 files, +1077/−442; latest `main` merged (#464, #494, #513, #514 in). e2e smoke scripts no longer install packages; the prepare action and pyproject bounds own the CI environment. Rebase owners: #428, #354, #485, #470, #375, #486.

> AI assistance (ZCode) was used for this change; every changed line was reviewed and all listed tests were run by the submitter.


knlnguyen1802 (commented):
two nits vs #492 — not blocking:

- `pyproject.toml` is still `diffusers>=0.40.0`. RFC §3 (and the 1A commit text) said `>=0.40.0,<0.41`. vllm-omni pins `==0.40.0`, so this is fine today; a later `pip` upgrade can walk off the pin. pls cap it or say the RFC line was dropped on purpose.
- #388 B4 is not in the tree. `_get_worker_extension_cls` and `npu_utils._is_npu_platform` still `except Exception`. the 1B commit claimed the startup probe re-raises and the runtime NPU probe narrows to `ImportError`. pls land that or drop the claim.

verl_omni/workers/rollout/vllm_rollout/vllm_omni_async_server.py
```diff
@@ -207,63 +199,71 @@ def _resolve_sleep_level(self) -> int:
...
         if self.node_rank != 0:
             return
-        await self.engine.collective_rpc(
-            "wake_up", kwargs={"tags": tags if tags is not None else self._get_wake_up_tags()}
-        )
-        self._invalidate_lora_request_cache()
+        if self.rollout_mode == RolloutMode.STANDALONE:
+            logger.info("skip wake_up in standalone mode")
+            return
+        await self._delegated_wake(tags)
```
knlnguyen1802: About the name `_delegated_wake` and `_delegate_sleep` I think it is not reasonable. Prefer to have api call directly here and  post process with validate the ack message and validate lora

verl_omni/trainer/omni/ray_omni_trainer.py
```diff
@@ -72,6 +72,16 @@ def _init_tokenizer(self):
         self.tokenizer = model_config.tokenizer
         self.processor = model_config.processor
 
+    # AsyncOmni.sleep() sets an admission hold that wake_up() does not clear;
+    # without this bridge the first generate() of init and of every step blocks.
```
knlnguyen1802: Could you mark this as TODO for me to check and fix on rollout side

zhtmike:
> two nits vs #492 — not blocking:
> 
> * `pyproject.toml` is still `diffusers>=0.40.0`. RFC §3 (and the 1A commit text) said `>=0.40.0,<0.41`. vllm-omni pins `==0.40.0`, so this is fine today; a later `pip` upgrade can walk off the pin. pls cap it or say the RFC line was dropped on purpose.

There is no special reason to cap `diffusers`, so I dropped during the human review.

> * #388 B4 is not in the tree. `_get_worker_extension_cls` and `npu_utils._is_npu_platform` still `except Exception`. the 1B commit claimed the startup probe re-raises and the runtime NPU probe narrows to `ImportError`. pls land that or drop the claim.

#388 B4 is a false alarm.

verl_omni/trainer/omni/ray_omni_trainer.py
zhtmike: done

verl_omni/workers/rollout/vllm_rollout/vllm_omni_async_server.py
zhtmike: done. inlined it.

knlnguyen1802:
cc: @NancyFyong  for review change related to Minimax H3

verl_omni/workers/rollout/vllm_rollout/vllm_omni_async_server.py
```diff
@@ -347,90 +342,48 @@ async def wait_for_requests_to_drain(self):
...
-                await self.clear_kv_cache()
-                logger.info("Prefix cache reset after abort")
+        aborted = False
+        try:
+            if request_ids:
+                # Never abort with an empty list — a silent no-op engine-side.
+                # TODO (mike): multi-stage AR abort is broken upstream — the engine's
+                # abort fallback terminal is stage_id=0 and the consume loop breaks on
+                # finished non-final messages, so generate() exits empty. Single-stage /
+                # thinker-only is correct here; needs a vllm-omni fix + pin bump.
```
zhtmike: @knlnguyen1802 

cc @dongbo910220 for this

verl_omni/workers/rollout/vllm_rollout/vllm_omni_async_server.py
```diff
@@ -347,90 +342,48 @@ async def wait_for_requests_to_drain(self):
...
-                    logger.info(
-                        "abort_all_requests: draining %d in-flight request(s) (%.1fs elapsed)",
-                        in_flight,
-                        time.monotonic() - drain_start,
-                    )
-                    last_count = in_flight
-                await asyncio.sleep(drain_poll_interval_s)
-
-            if drained:
-                if reset_prefix_cache:
```
AndyZhou952, line 392: **`verl_omni/workers/rollout/vllm_rollout/vllm_omni_async_server.py:426`**
`reset_prefix_cache` silently no-ops here; base verl always clears cache on abort for RL-correctness — stale-weight KV reuse risk once this gets a caller.

- line 430's `super()` branch never fires for AsyncOmni (no `output_processor` attr) → the native branch (438-446) always runs, and it ignores the flag.
- not a quick fix: `reset_prefix_cache()`/`reset_encoder_cache()` are TODO stubs upstream; the only real clear-cache path (`pause_generation`) would pause the whole engine as a side effect of one abort.
- no caller yet, but the description says one's coming (follow-up async trainers) — suggest name the gap in the description now rather than leave it silent

**Remark:  review by an agent**

zhtmike: > **`verl_omni/workers/rollout/vllm_rollout/vllm_omni_async_server.py:426`** `reset_prefix_cache` silently no-ops here; base verl always clears cache on abort for RL-correctness — stale-weight KV reuse risk once this gets a caller.
> 
> * line 430's `super()` branch never fires for AsyncOmni (no `output_processor` attr) → the native branch (438-446) always runs, and it ignores the flag.
> * not a quick fix: `reset_prefix_cache()`/`reset_encoder_cache()` are TODO stubs upstream; the only real clear-cache path (`pause_generation`) would pause the whole engine as a side effect of one abort.
> * no caller yet, but the description says one's coming (follow-up async trainers) — suggest name the gap in the description now rather than leave it silent
> 
> **Remark: review by an agent**

Right — and `abort_all_requests` (the only path with callers today) does honor the flag via pause_generation(clear_cache=…). The per-request path leaves it a no-op since the engine's resets are stubs and pause is engine-wide; gap now named in the description, to be revisited with the async-trainer callers.

**Remark: answer by an agent**

verl_omni/workers/rollout/vllm_rollout/vllm_omni_async_server.py
```diff
@@ -347,90 +342,48 @@ async def wait_for_requests_to_drain(self):
...
+        try:
+            if request_ids:
+                # Never abort with an empty list — a silent no-op engine-side.
+                # TODO (mike): multi-stage AR abort is broken upstream — the engine's
+                # abort fallback terminal is stage_id=0 and the consume loop breaks on
+                # finished non-final messages, so generate() exits empty. Single-stage /
+                # thinker-only is correct here; needs a vllm-omni fix + pin bump.
+                await asyncio.wait_for(
+                    engine.abort(request_ids), timeout=float(os.getenv("VERL_OMNI_ABORT_ACK_TIMEOUT_S", "120"))
+                )
```
knlnguyen1802: this is not important actually, abort with empty list will return immediately compare with do nothing in this fix.
They are literally the same
The problem mention is different things and yet, that is problem of multi stage abort. Noted it thanks

zhtmike: done, dropped the incorrect guard.

zhtmike:
~Fresh-env install issue from GPU verification of this PR (`docs/start/install.md`, py3.12): `uv pip install "vllm-omni @ git+…"` silently downgrades `kernels` to 0.14.1 (vllm-omni pins `==0.14.1`), recreating the #417 FA3-to-native degradation this PR's `==0.16.0` pin is meant to prevent — same for `accelerate` (`==1.12.0` vs `>=1.14.0`), since `override-dependencies` doesn't apply to that separate install step. Fix idea: add `kernels==0.16.0` to `override-dependencies`, or `--no-deps` the git pin, or fail a post-install check. With kernels re-pinned, everything imports cleanly and is ready for the GPU runs.~

Fixed

zhtmike:
~The transformers floor `>=5.10.1` is incompatible with kernels 0.16.0: a fresh resolve can pick 5.10.4, whose `integrations/hub_kernels.py` builds `LayerRepository(repo_id=..., layer_name=...)` with neither `revision` nor `version`, while kernels 0.16.0 requires exactly one — and since `_KERNEL_MAPPING` is constructed at module import, `import vllm` dies with `ValueError: Either a revision or a version must be specified.` Transformers 5.14.1 (inside the existing `<=5.14.1` cap) passes `version=`, so raising the floor to the first release that does fixes it.~

Fixed

zhtmike:
~`asdict()` in `vLLMOmniHttpServer.run_server` silently enables vLLM fault tolerance and kills every rollout server at startup: `EngineArgs.fault_tolerance_config` defaults to a `FaultToleranceConfig` instance (not None), the `engine_args = asdict(...)` round-trip serializes that default into a plain dict, and vllm's `__post_init__` treats any dict as an explicit `--fault-tolerance-config`, auto-setting `enable_fault_tolerance=True` — engine creation then fails with `Fault tolerance requires external load balancer mode` under verl's internal LB, before training starts. One-line fix after the `asdict` call: `if not engine_args.get("enable_fault_tolerance"): engine_args.pop("fault_tolerance_config", None)`. (This is likely the resurrection of the "dead fault_tolerance_config strip" dropped in 3832721.)~

Fixed

zhtmike:
~`qwen-omni-utils` doesn't declare `audioread`, so the AVQA dataset silently collapses to 0 samples on a fresh env: `qwen_omni_utils/v2_5/audio_process.py` does `import audioread` at import time but the package doesn't depend on it, so after `pip install -e ".[audio]"` the lazy `from qwen_omni_utils import process_mm_info` inside `QwenOmniRLHFDataset._process_multi_modal_info` raises per-row during prompt filtering, and training dies much later with the misleading `num_samples should be a positive integer value, but got num_samples=0`. Adding `audioread` (and `av`, same situation) to the `[audio]` extra makes it impossible.~

Fixed

zhtmike:
~The omni trainer hard-requires local `flash-attn` but nothing installs it on this stack: `verl_omni/workers/config/omni/model.py:170` defaults `attn_implementation` to `flash_attention_2`, so actor construction fails with `ImportError: ... the package for FlashAttention2 doesn't seem to be installed` — yet the flash-attn release matrix stops at `cu13torch2.9` (no torch 2.13 wheel), and neither `pyproject.toml` nor `docker/Dockerfile.cuda` covers it; I had to build 2.8.3.post1 from source (conda `cuda-nvcc`/`cuda-cudart-dev` 13.3, `TORCH_CUDA_ARCH_LIST=9.0`, 10 min). Either the install docs need a source-build note for the vllm 0.28 stack, or the omni actor should default to the hub-kernels backend like the diffusion trainer.~

Fixed

zhtmike:
~One zero-length-response sample per epoch batch crashes verl's `single_turn_agent` twice, and the omni trainer inherits that default. One AVQA prompt group (1/128, all 16 rollouts) deterministically produces a `response_length=0` sample at step 3 — notably the engine itself returns tokens (an instrumented `process_output` zero-token warning never fires), so the empty response is created in verl's continuous-token merge/align step — and verl then (1) raises `IndexError: index -1 is out of bounds` in `AgentLoopOutput.as_dict` (`rm_scores[-1] = reward_score` on an empty response_mask), and after guarding that, (2) raises `KeyError: 'rollout_log_probs'` because `SingleTurnAgentLoop.run` collapses the legit empty `log_probs` list to `None` via truthiness (`[] if output.log_probs else None`, `... if response_logprobs else None`), the unset field drops the batch column in `list_of_dict_to_tensordict`, and debug-metrics KeyErrors. Both are verl@fefb080 bugs worth fixing upstream (`is not None` checks + a numel guard); locally I registered an `omni_ar_single_turn_agent` loop (same flow, `is not None` checks, tolerant `as_dict`) and made `omni_trainer.yaml` default to it. There is no warning or error for the emptying itself — the only loud signatures are the two downstream crashes above; otherwise the phenomenon is fully silent and shows up only in step metrics as `response_length/min: 0.0` and `response/aborted_ratio > 0` (verl computes `aborted_mask = (response_length == 0)`, so "aborted" there literally means empty), with zero engine-side abort/preemption logs and the engine never returning an empty token list. It is also frequent, not a one-off: 113 of the first 155 steps had at least one empty response, up to 16% of rollouts in a step — my reference 0.2-stack run of the same recipe never showed a min below 12 across 348 steps, so why the CT merge empties these responses on the new stack is worth an upstream look.~

Fixed.

What happens:
image

SamitHuang:
SD3.5 validation results:
image
image
image

verl_omni/workers/rollout/vllm_rollout/vllm_omni_async_server.py
```diff
@@ -229,6 +237,7 @@ async def sleep(self):
             return
         acks = await self.engine.sleep(level=self._resolve_sleep_level())
         self._validate_acks("sleep", acks)
+        await self._reset_frontend_mm_cache()
```
knlnguyen1802, line 241: This is not reasonable,  frontend mm cache is the part of convert raw image into pixel (normally uint8 format) only and keep cache it. It has nothing related to model weight and dont need to reset after sleep

zhtmike: The cache content is indeed weight-independent — nothing in it goes stale. The reset is needed because `EngineCore.sleep(level>=1)` wipes the **engine-side** copy of the multimodal cache on its own (`sleep` → `pause_scheduler(clear_cache=True)` → `_reset_caches()`), while the frontend cache keeps acting as the sender's record of what the engine still holds: after each sleep, follow-up requests that reference an already-seen audio/image by hash only are finished **empty** with `FinishReason.ERROR` — the engine's `Multi-modal cache miss … cache drift` warning, 7,508 of them on the failing AVQA run (≈1.7% of samples, exactly the observed zero-length responses). vLLM proper clears the frontend cache on this same path (`AsyncLLM.pause_generation(clear_cache=True)` awaits `renderer.clear_mm_cache_async()` before the engine-side clear), and vllm-omni's `AsyncOmni.sleep` tries too, but its `reset_mm_cache()` looks the cache up as `input_processor.mm_processor_cache`, an attribute that doesn't exist — so it no-ops. This line just restores vLLM's contract on the frozen pin and is droppable once the engine fixes that lookup; an A/B with `mm_processor_cache_gb=0` should make the drift warnings and zero-length responses vanish if you want hardware proof.

knlnguyen1802: Get the idea, but it is better to call it in another place, putting it here is quite misleading

zhtmike: Agreed the placement isn't ideal — it's a pin-frozen workaround for the engine-side no-op (`AsyncOmni.reset_mm_cache` looks the cache up on `input_processor.mm_processor_cache`, which doesn't exist, so the pre-sleep clear in `AsyncOmni.sleep` silently does nothing). The natural place is upstream: fix that lookup to clear through `renderer.clear_mm_cache_async()`, matching `AsyncLLM.pause_generation`, and verl-omni carries nothing. Happy to drop this line if you can land the engine fix and we re-pin, or point me at the placement you'd prefer meanwhile.

zhtmike:
螢幕截圖 2026-09-03 上午9 38 43
螢幕截圖 2026-09-03 上午9 38 56

**GPU regression check: this PR vs the v0.2.0 stack (AVQA / qwen3-omni thinker LoRA GSPO, same recipe).** The rerun on the full PR stack ([`ta9g0ysq`](https://wandb.ai/verl-omni/gspo/runs/ta9g0ysq), 240 steps — vllm-omni 0.28 pin + lifecycle rework + mm-cache fix) shows **zero empty responses** (`response/aborted_ratio` 0.0 at every step), versus the pre-fix run on the same stack ([`81mq92l2`](https://wandb.ai/verl-omni/gspo/runs/81mq92l2)) where 7,508 cache-drift engine warnings silently emptied about 1.7% of samples. Correctness beyond that bug matches v0.2.0 ([`kzzrq9pr`](https://wandb.ai/verl-omni/gspo_demo/runs/kzzrq9pr), 348 steps): rollout-vs-trainer token-probability correlation is identical (0.995–0.999 in both), and prompt/response truncation ratios are byte-for-byte the same. Training quality is within single-seed noise — score 0.74 → 0.95 on the same trajectory (0.93 at step 157 in both), same response-length shortening (270 → 140) and entropy decay (1.35 → 1.20). Performance is strictly better: identical `timing_s/gen`, weight sync 17–20 s vs 25–28 s (the ACK-validated sleep/wake lifecycle is cheaper than 0.2.0's RPC sync), and 10–12% higher end-to-end throughput; the lifecycle path (sleep → sync → wake → resume bridge) ran clean on all 240 steps with no restarts. Diffusion-side changes are covered by the merge-blocking `ci-e2e-diffusion` gate rather than this run. No regressions found.

SamitHuang:
QwenImage:
image

image


image

verl_omni/pipelines/bagel_flow_grpo/common.py
```diff
@@ -70,5 +70,7 @@ def setup_bagel_sigmas(
         scheduler.set_timesteps(sigmas=sigmas, timesteps=sigmas, device=device)
     else:
         scheduler.set_timesteps(sigmas=sigmas, timesteps=sigmas)
+    # diffusers >= 0.40 ignores `timesteps=` and stores sigmas * 1000; the replay matches timesteps by exact value.
+    scheduler.timesteps = torch.tensor(sigmas, dtype=torch.float32, device=scheduler.timesteps.device)
```
SamitHuang: The CI `IndexError` is real, but this assignment is untested. `setup_bagel_sigmas` still has no CPU test that drives `index_for_timestep` / `sample_previous_step` after 0.40’s `set_timesteps` (which ignores `timesteps=` and writes `sigmas * num_train_timesteps`). Add a contract test that records a [0, 1] timestep from this grid and asserts the replay lookup hits, and that a 1000-scale value does not. Prefer `scheduler.timesteps = scheduler.sigmas[:-1]` so the lookup grid stays tied to the tensor `set_timesteps` actually stored after the shift formula, not the pre-call Python list.

verl_omni/workers/rollout/vllm_rollout/vllm_omni_async_server.py
```diff
@@ -347,90 +359,46 @@ async def wait_for_requests_to_drain(self):
...
+            # TODO (mike): multi-stage AR abort is broken upstream — the engine's
+            # abort fallback terminal is stage_id=0 and the consume loop breaks on
+            # finished non-final messages, so generate() exits empty. Single-stage /
+            # thinker-only is correct here; needs a vllm-omni fix + pin bump.
+            await asyncio.wait_for(
+                engine.abort(request_ids), timeout=float(os.getenv("VERL_OMNI_ABORT_ACK_TIMEOUT_S", "120"))
+            )
+            aborted = True
+            # Pause even with nothing to abort: holds admission until resume_generation.
+            await engine.pause_generation(
```
SamitHuang, line 392: `pause_generation(clear_cache=True)` already wipes the engine-side mm cache, and the pinned `AsyncOmni.reset_mm_cache()` still no-ops (wrong attribute). The renderer clear only runs after `sleep` / `release_kv_cache`. Any `generate()` that follows `abort_all_requests` without a sleep — the async trainer PRs this abort path is for — hits the same hash-only `FinishReason.ERROR` / empty-response failure this PR just fixed on the weight-sync path. Call `_reset_frontend_mm_cache()` after a successful pause here too.

verl_omni/trainer/omni/ray_omni_trainer.py
```diff
@@ -72,6 +72,17 @@ def _init_tokenizer(self):
...
+    # TODO (long): check and fix the resume bridge on the rollout side.
+    def on_init_end(self):
+        super().on_init_end()
+        self.checkpoint_manager.resume_generation_replicas()
+
+    def on_step_end(self):
+        super().on_step_end()
+        self.checkpoint_manager.resume_generation_replicas()
```
SamitHuang, line 84: This only fires for `omni_sync`. After AR `sleep()`, `wake_up()` leaves `_hold_admission_until_resume` set, so any trainer that sleeps replicas and then generates without this hook hangs on the first `generate()` of the next step. `OmniDirectPreferenceRayTrainer.update_weights` and the async trainer PRs (#337/#380) are not covered; #517 is the right ownership, but landing the pin without a rollout-side resume means those paths stay broken. Either resume from the replica sleep/wake path or document that only `omni_sync` is supported on this pin.

SamitHuang (changes requested):
Pin bump + abort-then-pause + ACK-validated sleep/wake is the right engine contract, and the AVQA empty-response diagnosis matches the pinned `AsyncOmni.reset_mm_cache()` looking up a missing attribute.

Three gaps before merge:
1. Bagel 0.40 timestep restore has no CPU `index_for_timestep` contract test (the CI IndexError can regress silently).
2. `abort_all_requests` → `pause_generation(clear_cache=True)` wipes the engine-side mm cache without the renderer clear, so the same hash-only empty-response path is still live for abort-then-generate (async trainers).
3. Admission resume is still only on `OmniPPOTrainerSync` (#517); any other AR trainer that sleeps replicas will hang on the next generate.

verl_omni/pipelines/bagel_flow_grpo/common.py
zhtmike: done, added a regression test, using `scheduler.timesteps = scheduler.sigmas[:-1]` instead.

verl_omni/workers/rollout/vllm_rollout/vllm_omni_async_server.py
zhtmike: done, add a call `_reset_frontend_mm_cache()` under `abort_all_requests`, with a TODO statement.

verl_omni/trainer/omni/ray_omni_trainer.py
zhtmike: done. added `resume_generation()` at the right caller.

SamitHuang (approved):
LGTM

zhtmike:
all CI green.

PR #456, opened 2026-08-27
[cfg, recipe, tests] feat: add Qwen3-Omni Thinker DAPO phase one

WenzheWang:
### What does this PR do?

Adds Phase 1 of #446: a Qwen3-Omni Thinker DAPO-without-dynamic-sampling baseline on the V1 omni trainer.

- Makes `OmniAlgoConfig` inherit upstream `AlgoConfig`, so the existing `filter_groups`, KL-control, and rollout-correction fields survive dataclass conversion.
- Adds a Thinker-only GPU LoRA AVQA launcher with vanilla token-level clipping (`0.2` / `0.28`), `token-mean`, GRPO advantages, no KL, and the AVQA `choice_reward`.
- Runs validation before training and every 10 steps with one uniform full-AVQA protocol: all 1,911 validation examples, greedy decoding, and `n=1`.
- Adds CPU launcher and config contract tests, including the enabled group-filtering Hydra conversion path.
- Documents that dynamic sampling and the overlong buffer are intentionally deferred.

The DAPO-specific GPU smoke was removed after review because the existing Qwen3-Omni GSPO smoke already exercises the same overall pipeline. The short-run performance assets and the extra reward-manager/loss warning were also removed after maintainer review. This is part of #446 and does not close the RFC.

### Checklist Before Starting

- [x] Searched for similar PRs: [open PRs mentioning Qwen3-Omni Thinker DAPO](https://github.com/verl-project/verl-omni/pulls?q=is%3Apr+is%3Aopen+%22Qwen3-Omni+Thinker+DAPO%22). No overlapping implementation was found; #446 is assigned to this author.
- [x] PR title follows `[{modules}] {type}: {description}`.

### Test

Current head: `bc0bee6ff8145c28b48a8dc2eba27f0f0d31fe2f`.
GitHub PR base snapshot: `main` at `783ea1f4198fb7c08b28dc1864aabd57f3ce6fcd`.

- Exact final tree `dea85fe228cde0246a53310f4df36e47068e764f`: 12 targeted CPU tests passed in CPU-only Slurm job `324519`. The suite covered the public DAPO launcher contract, AVQA `choice_reward`, shared config validation, and `OmniAlgoConfig` Hydra conversion.
- The same exact tree passed all 12 `pre-commit run --all-files` hooks in job `324519`.
- `git diff --check`: passed.
- Sphinx HTML build: completed. It reported the existing repository-wide 138 warnings and no new failure.
[...]

Issue #446, opened 2026-08-25
[RFC] Support DAPO training for Qwen3-Omni Thinker


SamitHuang:
### 1. Motivation

Qwen3-Omni Thinker is already a first-class omni policy in `verl-omni`. The README currently lists **DPO** and **GSPO** as the supported algorithms. [DAPO](https://arxiv.org/abs/2503.14476) (Decoupled Clip and Dynamic sAmpling Policy Optimization) is the production-scale GRPO-family recipe in upstream `verl` for long-CoT / math RL: clip-higher, token-level policy gradient, overlong reward shaping, and dynamic sampling.

There is no Thinker DAPO recipe today. The GSPO V1 launcher (`examples/gspo_trainer/qwen3_omni/run_qwen3_omni_thinker_gspo_lora_v1.sh`) is a sequence-level GSPO setup (`policy_loss.loss_mode=gspo`, `clip_ratio_{low,high}=3e-4/4e-4`, `loss_agg_mode=seq-mean-token-mean`, `reward.reward_manager.name=naive`). The legacy GSPO YAML sets `reward.reward_manager.name=dapo`, which only selects the **DAPO reward manager** (overlong buffer / math parsing). That is not the DAPO **algorithm**.

Supporting DAPO on Thinker is useful because:

1. Users already ask for it as the default long-horizon RL recipe on Qwen3-class models.
2. The Thinker adapter is algorithm-agnostic: DAPO should be a config + recipe + trainer-loop gap, not a new model adapter.
3. It gives a token-level GRPO baseline next to GSPO, which is needed to compare sequence-level vs token-level clipping on multimodal long responses.

This RFC proposes **Thinker-only DAPO** on the existing V1 omni path (`verl_omni.trainer.main_omni`, `OmniPPOTrainerSync`, vLLM-Omni AR rollout). It does not add a new trainer class.

### 2. Current baseline and gap

| Piece | Status |
| --- | --- |
| `OmniModelBase` / Qwen3-Omni Thinker training adapter | Available. Algorithm selection is `actor.policy_loss.loss_mode` + `algorithm.adv_estimator`. |
| V1 omni trainer (`omni_sync` → `PPOTrainerSync`) | Available. GSPO LoRA GPU and full-param NPU recipes exist. |
| vLLM-Omni AR rollout + LoRA weight sync | Available (`output_mode=ar`, `pipeline_name=qwen3_omni_moe`). |
[...]

PR #502, opened 2026-08-31
[cfg, recipe, tests] feat: wire DAPO overlong reward buffer on V1 omni loop

OnePunchMonk:
## Summary

Implements Phase 2 of #446 (Qwen3-Omni Thinker DAPO): wires the overlong
response-length penalty into the V1 omni reward path.

**Stacked on #456 (Phase 1) — not yet merged.** This branch is rebased onto
`wenzhe/qwen3-omni-thinker-dapo` at `bc0bee6` (current #456 head), so the diff
against `main` includes #456's commits. Please review only the two commits on
top (`feat: wire DAPO overlong reward buffer on V1 omni loop (#446 Phase 2)`,
`chore: trim redundant comments`) until #456 lands; I'll rebase and shrink the
diff once it merges.

**2026-09-04 update:** rebased onto #456's latest head (it had moved since
this PR was opened — the Phase-1 GPU smoke script was removed there per
review feedback, since the existing GSPO smoke covers the same pipeline).
Phase 2 restores that smoke script under its own scope, since it's this PR's
own e2e coverage for the overlong buffer, not a reintroduction of the removed
Phase-1 smoke. CPU tests pass post-rebase; the GPU e2e run below predates the
rebase and should be re-run before merge — I don't have further Modal/GPU
access in this pass.

## What's actually wired

`verl`'s registered `dapo` reward manager
(`verl.experimental.reward_loop.reward_manager.dapo.DAPORewardManager`) already
reads `config.reward.reward_kwargs.overlong_buffer_cfg` and
`config.reward.reward_kwargs.max_resp_len` — but `verl_omni`'s `reward.yaml`
never declared a `reward_kwargs` key, so any
`++reward.reward_kwargs.overlong_buffer_cfg.enable=true` override on the CLI
was silently swallowed rather than reaching the manager. This PR:

- Adds `reward.reward_kwargs` (`overlong_buffer_cfg.{enable,len,penalty_factor,log}`,
  `max_resp_len`) to `verl_omni/trainer/config/reward/reward.yaml`, disabled by
  default, and regenerates the four `_generated_*.yaml` trainer configs via
  `scripts/generate_trainer_config.sh`.
- Enables overlong shaping in the DAPO tiny-random smoke test
  (`tests/special_e2e/run_dapo_qwen3_omni_thinker_lora_v1_smoke.sh`), sized
  against its own `data.max_response_length=512`. The naive-manager AVQA
  example (`examples/dapo_trainer/qwen3_omni/run_qwen3_omni_thinker_dapo_lora_v1.sh`)
  is untouched and documented as such — overlong shaping is a no-op under
  `reward.reward_manager.name=naive`.
- Adds `tests/utils/test_dapo_overlong_reward_on_cpu.py`, which drives
  `DAPORewardManager` directly (no Hydra/trainer bring-up) and asserts the
  overlong penalty actually changes the reward on a truncated dummy response
  — this is Phase 2's exit criterion from the issue.
- Splits the shared launcher-contract assertions in
  `tests/utils/test_qwen3_omni_dapo_launcher_on_cpu.py` so "no overlong
  config" only applies to the naive-manager example, and adds
  overlong-specific assertions for the dapo-manager smoke test.

**Not in scope**: dynamic sampling (Phase 3), and any reward-loop metrics
plumbing beyond what `DAPORewardManager` already logs into `reward_extra_info`
(`overlong`, `overlong_reward`).

## Test plan

CPU (all green locally):
```bash
python3 -m pytest tests/utils/test_dapo_overlong_reward_on_cpu.py tests/utils/test_qwen3_omni_dapo_launcher_on_cpu.py -q
```

GPU e2e (2-GPU tiny-random smoke, exercises the new `reward.reward_kwargs.*`
overrides end-to-end through the real V1 trainer loop):
```bash
NUM_GPUS=2 TOTAL_TRAIN_STEPS=2 bash tests/special_e2e/run_dapo_qwen3_omni_thinker_lora_v1_smoke.sh
```
Ran this on a Modal 2×A10G box against an improvised environment (CUDA
12.4.1 base + the exact pins from `.github/actions/gpu-smoke-prepare/action.yml`),
since I don't have access to CI's private `cu130` registry image.

First two attempts hit environment issues unrelated to this PR: a missing
apt package, then a genuine upstream bug in `flashinfer-python==0.6.16.post3`
(pulled in transitively via `vllm==0.27.0`, no floor pinned anywhere in this
repo) that crashes any `TP>=2` GPU run at import time. Filed as #505, fixed
in #506 (floors `flashinfer-python>=0.6.16.post4` in the `gpu` extra) — see
those for the full root-cause writeup.

**With `flashinfer-python==0.6.16.post4` pinned (an environment-only override,
nothing in this repo changed), the smoke test passes:**
```
RETURNCODE 0
Qwen3-Omni Thinker DAPO+LoRA e2e V1 smoke without dynamic sampling passed.
Training Progress: 100%|██████████| 2/2 [01:54<00:00, ...]
```
Per-sample DAPO reward scores varied across the batch (`-1.0`, `-0.3438`,
`-0.0625`), consistent with the overlong-buffer penalty grading responses
approaching/at `max_resp_len=512` (buffer `len=128`, so penalty onset at
~384 tokens, full -1.0 at 512).

One caveat I'm not glossing over: I could not find an explicit
`step:N - reward/overlong...` console-logger line in the captured output —
grepped the full stdout+stderr for `step:` and got zero matches for *any*
metric, not just overlong, so this looks like a console-logging/capture gap
in this V1 path rather than evidence the wiring is broken (the reward
computation itself clearly executed through the DAPO path, per the reward
scores above). Worth a maintainer's eyes on whether metrics are expected to
print to console here or only to the configured tracker backend.

## AI usage disclosure

Claude Code was used to trace the config-plumbing gap, implement this change,
and write the tests described above. A human submitter (onepunchmonk) has
reviewed every changed line and will review the Modal GPU run before taking
this out of draft.


OnePunchMonk:
Marking this reviewable once #456 (Phase 1) lands. This branch is stacked on `wenzhe/qwen3-omni-thinker-dapo`, so the diff against `main` currently includes #456's commits — please hold off on a full review until that merges. Once it does, I'll rebase onto `main` and the diff will shrink to just this PR's Phase 2 commit.

Everything else in this PR is otherwise ready: CPU tests pass, GPU e2e smoke passes (see the PR description for the flashinfer environment detour — filed as #505/#506, unrelated to this PR's code), and the config-plumbing/reward-manager wiring is verified against upstream `verl`.

OnePunchMonk:
## GPU smoke re-run after rebase onto #456's latest head

Re-ran the Phase 2 GPU e2e smoke on Modal (2×A10G, CUDA 12.4.1 base + the same
pins from `.github/actions/gpu-smoke-prepare/action.yml`, with
`flashinfer-python>=0.6.16.post4` floored per #506) against the current
rebased head (`7dc3bd9`, rebased onto #456 at `bc0bee6`), so this supersedes
the earlier Modal run that predated the rebase.

```bash
NUM_GPUS=2 TOTAL_TRAIN_STEPS=2 bash tests/special_e2e/run_dapo_qwen3_omni_thinker_lora_v1_smoke.sh
```

```
RETURNCODE 0
Training Progress: 100%|██████████| 2/2 [00:57<00:00, 28.99s/it]
Qwen3-Omni Thinker DAPO+LoRA e2e V1 smoke without dynamic sampling passed.
```

Per-sample DAPO reward scores again varied across the batch (`-1.0`, `-1.0`,
`-0.3438`, `-0.0703`), consistent with the overlong-buffer penalty grading
responses at/near `max_resp_len=512` (buffer `len=128`, penalty onset at
~384 tokens) — same pattern as the pre-rebase run, on the code as it stands
now.

CPU tests also still pass on this head:
```bash
python3 -m pytest tests/utils/test_dapo_overlong_reward_on_cpu.py tests/utils/test_qwen3_omni_dapo_launcher_on_cpu.py -q
# 3 passed
```

Marking this ready for review now that it's rebased onto #456's latest and
both CPU and GPU coverage are green on the current head.

OnePunchMonk:
Hi @zhtmike, marking this as ready for review since #456 is now merged. Thank you!

tests/special_e2e/run_dapo_qwen3_omni_thinker_lora_v1_smoke.sh
```diff
@@ -0,0 +1,118 @@
+#!/usr/bin/env bash
```
wtomin, line 1: I know you plan to support dynamic sampling in #540. Why don't you move the smoke test to #540, so that you can enable both overlong reward buffer and dynamic sampling in the same test script?

Besides, to make this test take effect, you also need to add it to `tests/gpu_smoke/run_gpu_smoke_omni_e2e.sh`. If you have special pin version of third-party libraries, like `flashinfer-python==0.6.16.post3`, you may need to update `.github/actions/gpu-smoke-prepare/action.yml`.

examples/dapo_trainer/README.md
```diff
@@ -19,6 +19,15 @@ policy loss with asymmetric clipping, GRPO advantages, and no KL penalty. The
 registered naive reward manager calls the AVQA `choice_reward`; the reward
 manager name alone does not select the optimization algorithm.
 
+**Phase 2 (#446): overlong reward buffer.** Overlong shaping is wired through
```
wtomin, line 22: I think there is no need to mention about phase 2 or #446. Please simply give instructions on how to enable this overlong reward buffer feature, and what effect is to expected.

wtomin:
@OnePunchMonk  Thank you! Could you please provide the validation reward just like what is in #456 ? I am particularly interested in how the validation reward gets improved when overlong reward buffer is newly supported. 

@WenzheWang  Please also give your valuable comments on this PR.

WenzheWang (changes requested):
Thanks for wiring the overlong buffer configuration. I found one reproducible test failure and a gap in coverage of the configuration path introduced here:

1. **Fix the shared launcher contract assertion.** The new smoke test calls `_assert_dapo_without_dynamic_sampling_contract(smoke)`, but that helper still asserts that `overlong_buffer_cfg` is absent. The smoke intentionally enables it, so the new test fails before reaching its positive assertions. Running `python -m pytest tests/utils/test_qwen3_omni_dapo_launcher_on_cpu.py -q` against this head produced **1 failed, 1 passed**. Please remove the no-overlong restriction from the shared helper and keep it in the naive-example test, which already checks it separately.

2. **Cover the actual configuration integration.** The reward test constructs a handwritten `DictConfig` and directly instantiates `DAPORewardManager`, bypassing the changed YAML and registered factory. Please add a CPU regression test that composes `omni_trainer` with non-default `reward.reward_kwargs.*` overrides, constructs through `load_reward_manager`, and checks the forwarded settings and enabled/disabled penalty results. This would catch a missing or misnested YAML field that the current test cannot detect.

The existing tokenizer/offline-test comment also remains applicable. I will avoid duplicating the ongoing smoke-placement and README discussions.

tests/utils/test_dapo_overlong_reward_on_cpu.py
```diff
@@ -0,0 +1,82 @@
...
+                        "len": OVERLONG_BUFFER_LEN,
+                        "penalty_factor": OVERLONG_PENALTY_FACTOR,
+                        "log": True,
+                    },
+                    "max_resp_len": MAX_RESP_LEN,
+                }
+            }
+        }
+    )
+    tokenizer = AutoTokenizer.from_pretrained("gpt2")
```
wtomin, line 49: I suggest to use local dummy tokenizer, instead of downloading it from huggingface. Sometimes the connection fail, which may cause unit test fail too.

OnePunchMonk:
Thanks for the feedback. I don't have a 4x A800 (or equivalent) setup on hand to run the full AVQA e2e comparable to #456's evidence.

@WenzheWang, would you (or anyone with access to the cluster #456 ran on) be able to run this one? Ideally: resume from the same #456 seed/checkpoint (step 50 or wherever's cheapest) with `reward.reward_kwargs.overlong_buffer_cfg.enable=true` now on, and continue far enough (~150-235 steps, matching #456's horizon) to get a real before/after validation-reward delta rather than an early-noise read.

Alternatively, is there any compute assistance available for contributors (project-provided cluster time, credits, etc.) so I could run this myself? Happy to do it directly if there's a way to get access.

@zhtmike @SamitHuang, if there's cluster time available on your side that'd also work, happy to hand off the exact launch command and configs either way. Let me know what's easiest.

zhtmike:
Pls pick the task which you can handle with resources next time, thanks. It is not our duty to verify your PR.

OnePunchMonk:
Understood, thanks, apologies for the back and forth on this. Closing it out.

Issue #503, opened 2026-08-31
[RFC] Shared Talker RL rollout-to-replay interface


dongbo910220:
### Feature request

This RFC proposes a small, model-neutral rollout-to-replay contract for the autoregressive Talker stage of an omni model. A rollout adapter maps model-native engine output into the policy sequence used by RL while retaining the exact trajectory and conditioning needed for actor replay; a training adapter converts that retained payload into architecture-specific forward inputs.

First interface PR: #504.

The proposed lifecycle is:

```text
vLLM-Omni model-native output
    -> rollout adapter
    -> AgentLoopOutput policy fields + extra_fields[model_defined_key]
    -> AgentLoopWorker batches model_defined_key into the top-level non-tensor batch
    -> actor micro_batch[model_defined_key]
    -> training adapter
    -> architecture-specific actor forward inputs
```

The shared transport reuses existing `AgentLoopOutput` and `AgentLoopWorker` behavior:

- `response_ids` is the sampled policy-token sequence optimized by RL.
- `response_mask` and optional `response_logprobs` align one-to-one with `response_ids`.
- Each concrete adapter stores its per-sample trajectory and conditioning payload under a model-defined, namespaced key in `extra_fields`.
- `AgentLoopWorker` collects each `extra_fields` key into the top-level non-tensor batch, and `DataProto.to_tensordict()` exposes it to the actor as `micro_batch[model_defined_key]`.
- The training adapter validates and unwraps its own payload, including any `NonTensorStack` representation, and constructs the exact actor forward inputs.
- Missing fields, invalid shapes, excessive policy length, and policy-alignment mismatches fail closed.

For example, a concrete adapter may store `output.extra_fields["my_talker_replay"] = {"trajectory": ..., "conditioning": ...}` during rollout and consume the batched payload from `micro_batch["my_talker_replay"]` during actor replay. The outer key name and nested payload schema remain model-owned.

The interface deliberately does not define a shared Talker trajectory dataclass or encode Qwen3-TTS-specific assumptions such as 16 codebooks, fixed speaker/text slots, codec-0 offsets, or an external speaker embedding. Concrete adapters own their key name, trajectory schema, validation, device transfer, dtype conversion, and padding or stacking policy.

### Motivation

Without this boundary, each Talker integration must either teach the generic trainer about its codebook and conditioning layout or copy a private rollout-to-actor transport. Both options couple shared RL infrastructure to one architecture and make rollout/actor trajectory mismatches difficult to detect.

[...]

PR #504, opened 2026-08-31
[model, rollout] feat: add generic Talker RL interface

dongbo910220:
### What does this PR do?

This PR implements the model-neutral rollout-to-replay contract proposed in RFC #503 and follows the interface-first split requested during review of #428. Rollout adapters map model-native trajectories into aligned `AgentLoopOutput` policy fields, while training adapters turn model-owned replay payloads into the exact actor inputs required to replay the sampled policy trajectory.

The shared API deliberately does not encode Qwen3-TTS assumptions such as 16 codebooks, fixed speaker/text slots, codec-0 offsets, or an external speaker embedding. Each concrete adapter owns a model-defined, namespaced replay key and its nested schema. Qwen3-TTS remains the follow-up concrete implementation in #428; Qwen3-Omni and future Talkers can retain different trajectory and conditioning payloads behind the same lifecycle hooks.

The transport follows verl's existing batching behavior: a per-sample payload stored as `AgentLoopOutput.extra_fields[model_defined_key]` is collected by `AgentLoopWorker` into the top-level non-tensor batch and reaches the actor as `micro_batch[model_defined_key]`.

### API and usage example

A rollout adapter for an omni model's autoregressive Talker stage maps its engine-native output into the policy sequence while retaining model-owned replay data under its own namespaced key:

```python
class MyRolloutAdapter(OmniRolloutPipelineBase):
    @classmethod
    def postprocess_agent_loop_output(cls, output, *, tokenizer, response_length):
        replay = output.extra_fields["my_talker_replay"]
        output.response_ids = replay["policy_tokens"][:response_length]
        output.response_mask = [1] * len(output.response_ids)
        return output
```

After `AgentLoopWorker` batches the samples, the training adapter reads that model-defined key from the top level of `micro_batch`, validates it, and adds the exact replay inputs after standard language-model input preparation:

```python
from verl.utils import tensordict_utils as tu


class MyTrainingAdapter(OmniModelBase):
    @classmethod
    def prepare_model_inputs(cls, model_inputs, micro_batch, model_config):
        replay_payloads = tu.get(micro_batch, "my_talker_replay")
        if replay_payloads is None:
            raise ValueError("Missing my_talker_replay payload required for exact actor replay.")
        trajectories = [payload["trajectory"] for payload in replay_payloads]
        return {**model_inputs, "trajectories": trajectories}
```

Select the Talker loop and the registered rollout adapter through configuration:

```text
actor_rollout_ref.rollout.agent.default_agent_loop=omni_single_turn_agent
+actor_rollout_ref.rollout.engine_kwargs.vllm_omni.pipeline_name=my_talker
```

### Design and code changes

- Add `OmniRolloutPipelineBase.postprocess_agent_loop_output` as an optional boundary for mapping model-native rollout output to aligned policy IDs, masks, and optional rollout log-probabilities while retaining architecture-specific replay data under a model-defined key in `extra_fields`.
- Add `OmniModelBase.prepare_model_inputs` and call it from `OmniFSDPEngine` after standard language-model inputs have been prepared.
- Register `omni_single_turn_agent` for an omni model's autoregressive Talker stage, resolve its rollout adapter from the required `engine_kwargs.vllm_omni.pipeline_name`, and fail closed on unknown adapters, malformed output types, excessive policy length, or token-alignment mismatches.
- Cover two synthetic Talker contracts with different trajectory and conditioning schemas, and verify the real `AgentLoopWorker -> DataProto -> TensorDict` transport exposes the namespaced replay key at the top level of the actor micro-batch.
- Document how an omni model's autoregressive Talker stage opts into the hooks without imposing a shared key name, codebook count, conditioning source, or nested payload schema.
- Leave the default hook implementations as no-ops so existing text-token policy paths remain unchanged.

### Tests

all passed.

### DIFF
--- a/verl_omni/agent_loop/single_turn_agent_loop.py
+++ b/verl_omni/agent_loop/single_turn_agent_loop.py
@@ -16,14 +16,73 @@ import os
 from typing import Any
 from uuid import uuid4
 
-from verl.experimental.agent_loop.agent_loop import AgentLoopBase, register
+from verl.experimental.agent_loop.agent_loop import AgentLoopBase, AgentLoopOutput, register
+from verl.experimental.agent_loop.single_turn_agent_loop import SingleTurnAgentLoop
 from verl.utils.profiler import simple_timer
 from verl.utils.tokenizer.chat_template import apply_chat_template as _apply_chat_template
 
 from verl_omni.agent_loop.diffusion_agent_loop import DiffusionAgentLoopOutput
+from verl_omni.pipelines.model_base import OmniRolloutPipelineBase
 
 logger = logging.getLogger(__file__)
 logger.setLevel(os.getenv("VERL_LOGGING_LEVEL", "WARN"))
+
+
+@register("omni_single_turn_agent")
+class OmniSingleTurnAgentLoop(SingleTurnAgentLoop):
+    """Single-turn loop for autoregressive omni stages with custom policy tokens."""
+
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        self.rollout_adapter = self._resolve_rollout_adapter(self.rollout_config)
+
+    @staticmethod
+    def _resolve_rollout_adapter(rollout_config):
+        try:
+            pipeline_name = rollout_config.engine_kwargs["vllm_omni"]["pipeline_name"]
+        except (KeyError, TypeError) as error:
+            raise ValueError("omni_single_turn_agent requires engine_kwargs.vllm_omni.pipeline_name.") from error
+
+        adapter = OmniRolloutPipelineBase.get_class(pipeline_name)
+        if adapter is None:
+            raise ValueError(
+                "omni_single_turn_agent requires a registered "
+                f"engine_kwargs.vllm_omni.pipeline_name, got {pipeline_name!r}."
+            )
+        return adapter
+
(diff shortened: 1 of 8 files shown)


verl_omni/agent_loop/single_turn_agent_loop.py
```diff
@@ -16,16 +16,75 @@
...
 
 from verl_omni.agent_loop.diffusion_agent_loop import DiffusionAgentLoopOutput
+from verl_omni.pipelines.model_base import OmniRolloutPipelineBase
 
 logger = logging.getLogger(__file__)
 logger.setLevel(os.getenv("VERL_LOGGING_LEVEL", "WARN"))
 
 
+@register("omni_single_turn_agent")
+class OmniSingleTurnAgentLoop(SingleTurnAgentLoop):
```
zhtmike, line 32: Better to mention this is for talker part of the omni model only, here and in the doc

zhtmike:
Looks good, minor suggestion

zhtmike:
nice! are you interested to give a sharing on this and future plan in today's meeting?

dongbo910220:
> nice! are you interested to give a sharing on this and future plan in today's meeting?

Yes, happy to.

verl_omni/agent_loop/single_turn_agent_loop.py
dongbo910220: Updated

zhtmike (approved):

PR #513, opened 2026-09-02
[trainer, tests] fix: union colocated reward output into the v1 diffusion trainer batch

cr-gao:
### What does this PR do?

The v1 diffusion trainer assigns the reward-only DataProto returned by `_compute_reward_colocate` over the whole batch, in both the train step and `_validate`. The trajectory tensors are dropped and the next stage crashes (`KeyError: 'all_timesteps'` in old_log_prob; `prompts` in validate). The v0 trainer unions the reward output into the batch; this PR does the same in v1. The path was never exercised before: every v1 recipe uses a standalone reward pool, so the colocated branch (`reward_model.enable=true` without `enable_resource_pool`) first ran during OPD scheduler validation on #495's branch.

### Why this is not duplicating an existing PR

No open PR touches the v1 colocated reward path; the bug dates to the v1 port (#296).

### Tests

- CPU (RED first on main): `pytest tests/trainer/diffusion/test_v1_colocate_reward_on_cpu.py` — a stubbed reward step asserts the batch after the reward stage keeps both the trajectory fields and `rm_scores`; fails on main, passes with the fix. v1 suites regression: 93 passed on this head (independently re-verified from a clean archive; main + this test = 1 failed as expected).
- E2E (5× RTX PRO 6000): 12-step SD3.5-M dual-teacher separate_async run with colocated GenRM — crashed at step 1 on main's code path, 12/12 PASS with the fix (blocking reward stage ~10.2s/step, training healthy).

AI assistance (Claude Code) was used for this change; the submitting human has reviewed every changed line and run the tests above.

### DIFF
--- a/tests/trainer/diffusion/test_v1_colocate_reward_on_cpu.py
+++ b/tests/trainer/diffusion/test_v1_colocate_reward_on_cpu.py
@@ -1,0 +1,68 @@
+# Copyright 2026 Bytedance Ltd. and/or its affiliates
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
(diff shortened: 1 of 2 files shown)


knlnguyen1802:
Actually I plan to fix it here https://github.com/verl-project/verl-omni/pull/489

knlnguyen1802:
But I think can move forward this first to avoid other blocking

knlnguyen1802 (approved):

cr-gao:
> Actually I plan to fix it here #489

Oh I didn't see... yeah this bug is on a hot path:)

Issue #432, opened 2026-08-24
[RFC] Multi-reward-model serving with vLLM and vLLM-Omni deployments


Sky-Trigger:
## Summary

This RFC proposes a reward-model management layer in `verl-omni` that can coordinate one or more engine-backed reward models.

The design follows the existing `verl` reward-model lifecycle instead of introducing a new inference runtime. Reward models remain hosted by vLLM or vLLM-Omni server processes, while `verl-omni` manages their endpoints, lifecycle, parallel configuration, request dispatch, and score aggregation.

The interface should be small and backend-oriented so that reward-model implementations are decoupled from the trainer. This is an interface boundary inside `verl-omni`, not a proposal for a third-party library or plugin system.

## Motivation

The current `verl` reward path already provides a good design for one standard reward-model deployment:

```text
RewardLoopManager
  -> RewardModelManager
       -> HTTP router
            -> vLLM rollout replicas
  -> RewardLoopWorker
       -> send scoring requests to the router
```

Its lifecycle is also a useful baseline: `RewardModelManager` wakes its logical replicas before scoring, waits for accepted requests to finish, and sleeps them afterward. Model weights and TP/PP/DP execution remain inside the engine server ranks, while `RewardLoopWorker` acts only as a client.

`verl-omni` should preserve these properties while extending the design to multiple reward models and vLLM-Omni deployments.

## Proposed architecture

```text
TaskRunner / trainer
  -> Omni reward-model manager
       -> deployment A -> vLLM manager/router/replicas
       -> deployment B -> vLLM manager/router/replicas
       -> deployment C -> vLLM-Omni API/replicas/stages

RewardLoopWorker
  -> request selected deployments
  -> collect named model results
  -> aggregate the final reward
```

The control plane and data plane remain separate:

- `TaskRunner` or the trainer controls deployment creation, readiness, wake-up, and sleep.
- `RewardLoopWorker` sends scoring requests directly to deployment endpoints.
- Sample payloads do not pass through `TaskRunner`.
- Model weights live only in vLLM or vLLM-Omni server worker/rank processes.

## Backend interface

`verl-omni` should define a thin backend adapter around engine-managed deployments. At a high level, a backend needs to provide:

- startup and shutdown;
- a scoring endpoint or client descriptor;
- `wake_up()` and `sleep()`;
- readiness and health information;
- engine-native parallel configuration.

[...]

PR #524, opened 2026-09-03
[reward, trainer] feat: add managed multi-reward deployments

Sky-Trigger:
## What does this PR do?

Related to #432.

This PR adds opt-in named reward models under `reward.models`. A training job
can use multiple model-backed rewards, give each model an explicit lifecycle
and resource allocation, and keep model inference separate from reward
calculation.

Two backends are supported:

- `engine` wraps the pinned upstream `verl.RewardModelManager`. The currently
  supported and validated named-engine path is vLLM.
- `native` loads a user-supplied model class inside accelerator-bound Ray
  reward workers for checkpoints that are not supported by the serving engine.

Each configured `reward.reward_functions.<term>.path/name` function owns input
interpretation and final score semantics. Engine models expose router/model
arguments; native models expose a `reward_model` inference handle.

## Native backend contract

The native backend is deliberately small:

- it is replica-only in this PR: each `placement.devices` entry creates one
  complete model replica;
- the user-supplied `executor.model` class owns model construction and
  inference;
- the framework owns placement, lifecycle, and request dispatch;
- `infer(*args, **kwargs)` is an intentionally untyped internal contract
  between the native model adapter and its configured reward function;
- synchronous and asynchronous `infer()` and optional `close()` methods are
  supported.

Native TP/PP/FSDP is not implemented. A future sharded-native design needs an
RFC-level replica-group contract; a flat device list cannot distinguish
complete replicas from ranks within one sharded replica.

## Resource ownership and placement

The trainer still selects one parent resource pool:

- `reward.reward_model.enable_resource_pool=false` selects `global_pool`;
- `reward.reward_model.enable_resource_pool=true` selects the dedicated
  `reward_pool`.

Engine allocations consume a contiguous prefix of that parent pool in model
configuration order, with one upstream `SubRayResourcePool` per engine model.

Each native model receives an independent indexed allocation. Its
`placement.devices` values are global logical bundle indices within the
selected parent pool. They are not physical CUDA/NPU IDs and are not
tensor-parallel ranks. Indices may be non-contiguous, but must be in range and
cannot overlap another native model or the engine prefix.

For example:

```text
selected parent pool
├── bundles 0-3  -> engine model A
├── bundles 4,6  -> native model B, two complete replicas
├── bundles 5,7  -> native model C, two complete replicas
└── other bundles -> unused by named rewards
```

When the global pool is selected, bundles unused by named rewards may still be
used by actor/rollout workers.

## Reward manager behavior

Named-model aggregation currently supports the visual sample contract only.
Jobs must explicitly select:

```yaml
reward:
  reward_manager:
    name: MultiVisualRewardManager
```

The framework validates this requirement and does not silently rewrite a
user-provided reward manager. A modality-neutral manager and Audio/Visual
manager consolidation are follow-up work.

## Lifecycle and async API

Engine and native models share one user-facing `offload` option:

- `offload: true` wakes before scoring and sleeps afterward;
- `offload: false` keeps the model resident.

Scoring is bracketed by wake/score/sleep with cleanup in `finally`. Independent
model groups execute concurrently inside one scoring call. A manager-level
single-flight lock serializes overlapping calls to
`async_compute_rm_score()`, preventing one request from sleeping models while
another request is entering Ray dispatch.

`compute_rm_score()` remains the synchronous compatibility wrapper used by
current trainers.

## Compatibility

- Existing jobs without `reward.models` keep the legacy single-model path.
- `reward.reward_model.enable=true` cannot be combined with named models.
- Existing custom reward functions and legacy accelerator workers remain
available when named models are absent.
- The PR does not automatically migrate or rewrite legacy configurations.

The extension and migration guide is
`docs/algo/named_reward_models.md`.

## Validation

Validated in the `verl-omni:gpu` container:

```bash
python -m pytest -q \
  tests/reward_loop/test_audio_reward_manager_on_cpu.py \
  tests/reward_loop/test_multi_reward_manager_on_cpu.py \
  tests/reward_loop/test_named_reward_models_on_cpu.py \
  tests/reward_loop/test_reward_loop_manager_on_cpu.py \
  tests/trainer/diffusion/test_v1_colocate_reward_on_cpu.py
# 116 passed
```

Also passed:

- Ruff and format checks;
- dynamic import of the relocated pooling worker extension;
- `py_compile` for changed Python modules;
- reward configuration documentation test;
- docs timestamp and imported-doc validation;
- exported API docstring check for `verl_omni`;
- shell syntax checks for the native and engine/native NPU launchers.

The NPU engine/native E2E recipe is
`tests/special_e2e/run_qwen_image_edit_lora_v1_npu_engine_native.sh`.
It was manually rerun successfully on a 16-NPU setup after fixing duplicate
rollout sleep in the synchronous compatibility path. This manual result is
reported separately from CI.

## Current limitations and follow-up work

- Named-model aggregation currently uses `MultiVisualRewardManager`.
- Native execution is accelerator-only and replica-only.
- Native FSDP requires a separate RFC decision and replica-group design.
- Dynamic load balancing and work stealing are not implemented.
- Named models do not participate in streaming reward computation.
- vLLM-Omni reward serving is not implemented.


Sky-Trigger:
@chenyingshu PTAL. I've been working on this recently, but it still needs some time to be further refined.

Sky-Trigger:
@SamitHuang @zhtmike @chenyingshu PTAL.
feef8bf85f63afafa0c38107c9542a16
The existing reward implementations cannot yet migrate seamlessly to named deployments. Each one requires a small adapter change to consume the deployment-provided inference interface while keeping its scoring logic unchanged. Migrating the remaining reward functions can be handled separately in follow-up contributions.

verl_omni/utils/reward_score/pickscore_reward.py
zhtmike, line 1: does it mean we support pickscore only?

Sky-Trigger: No, this is a general framework. PickScore is currently used as an example. Other reward models not supported by vLLM/vLLM-Omni can also use it with minor adaptations.

Sky-Trigger: The main limitation is that existing reward models supported by vLLM/vLLM-Omni also require minor adaptations to migrate to this framework. To avoid broad, error-prone changes, I left them unchanged for now.

zhtmike:
I see, I will do a human review later, thanks

docs/algo/managed_reward_deployments.md
```diff
@@ -0,0 +1,549 @@
...
+Last updated: 09/09/2026
+
+`reward.deployments` manages multiple model-backed rewards in one training
+job. It keeps model lifecycle and resource ownership separate from reward
+semantics:
+
+- a deployment owns model resources, loading, router access, and lifecycle;
+- a reward function turns a rollout output and a model response into a scalar
+  score;
+- `MultiVisualRewardManager` combines all term scores with a weighted sum.
```
zhtmike: how about AudioRewardManger, etc, introduced in the repo.

I think we can make a unified MultiRewardManger, and depreciated the old AudioRewardManager and VisualRewardManager, the only difference is the `run_single` for audio and visual, we can implement add hook for them

docs/algo/managed_reward_deployments.md
```diff
@@ -0,0 +1,549 @@
...
+on the backend.
+
+Keeping a deployment resident requires enough accelerator memory for it to
+coexist with the actor rollout. When deployments share `global_pool`, prefer
+the default `offload: true` unless the combined resident footprint is known to
+fit; otherwise use a separate reward pool.
+
+Choose one backend for a normal single-model reward job:
+
+- use `engine` when vLLM or vLLM-Omni can serve the model and the configured
```
zhtmike: I am not sure vLLM-Omni has any reward model supported now; might just drop it

docs/algo/managed_reward_deployments.md
```diff
@@ -0,0 +1,549 @@
...
+
+```yaml
+reward:
+  reward_model:
+    enable: false
+    enable_resource_pool: false
+  deployments:
+    pickscore:
+      backend: native
+      adapter: pickscore
```
zhtmike: `adapter` -> `reward_name`

docs/algo/managed_reward_deployments.md
```diff
@@ -0,0 +1,549 @@
...
+
+```yaml
+deployments:
+  pickscore_native:
+    backend: native
+    offload: true
+    adapter: pickscore
+    model_path: /models/PickScore_v1
+    placement:
+      devices: [0, 1, 2, 3]
```
zhtmike: how is native model rounting handling, if there is any load balancing problem exists.

docs/algo/managed_reward_deployments.md
```diff
@@ -0,0 +1,549 @@
...
+        engine_kwargs:
+          vllm:
+            runner: pooling
+  reward_functions:
+    pickscore:
+      deployment: pickscore_engine
+      path: pkg://verl_omni.utils.reward_score.pickscore_reward
+      name: compute_score_pickscore_engine
+      logit_scale: 98.86447
+      weight: 1.0
```
zhtmike: looks good

docs/algo/managed_reward_deployments.md
```diff
@@ -0,0 +1,549 @@
...
+2. Copy the old `model_path` and engine rollout fields into one
+   `reward.deployments.<name>` entry.
+3. Move the old custom reward function `path/name` into one
+   `reward.reward_functions.<term>` entry and bind it with
+   `deployment: <name>`.
+4. Set the reward-manager configuration to `MultiVisualRewardManager` only
+   when the recipe does not already select it; named deployments do this during
+   reward-loop initialization.
+
+For example, migrate this legacy engine configuration:
```
zhtmike: we need to add depreciated message for old engine, if you want to mark it as legacy

docs/algo/managed_reward_deployments.md
```diff
@@ -0,0 +1,549 @@
...
+class or supported adapter, assign non-overlapping native placement bundle
+indices, and configure the matching reward term's `path/name`. The model class
+returns inference outputs; the reward function computes the score.
+
+Do not put the legacy `reward_model` and its named replacement in the same
+configuration. The reward loop rejects that mixture before allocating models.
+
+## Current limits
+
+- Native models are replicated, not tensor parallel.
```
zhtmike: Can we support fsdp for native model in future PR, it should be simple, no need model code hacking.

docs/algo/managed_reward_deployments.md
```diff
@@ -0,0 +1,549 @@
...
+      nnodes: 1
+      rollout:
+        name: vllm
+        tensor_model_parallel_size: 1
+    pickscore:
+      backend: native
+      adapter: pickscore
+      model_path: /models/PickScore_v1
+      placement:
+        devices: [0, 1, 2, 3, 4, 5, 6, 7]
```
zhtmike: it is possible to add a new option, put the reward model on cpu only? if there reward model is not that huge, cpu is just fine.

docs/algo/managed_reward_deployments.md
```diff
@@ -0,0 +1,549 @@
...
+native deployments must be disjoint.
+
+For example, if the native subpool has eight bundles:
+
+```yaml
+placement:
+  devices: [0, 1, 2, 3]
+```
+
+creates four full replicas on the first four native bundles. It does not shard
```
zhtmike: I like this symmetric.

examples/flowgrpo_trainer/qwen_image_edit/run_qwen_image_edit_lora_v1_npu.sh
```diff
@@ -83,10 +85,15 @@ python3 -m verl_omni.trainer.main_diffusion_v1 \
     actor_rollout_ref.rollout.val_kwargs.pipeline.num_inference_steps=40 \
     actor_rollout_ref.rollout.val_kwargs.algo.noise_level=0.0 \
     actor_rollout_ref.ref.log_prob_micro_batch_size_per_gpu=4 \
-    reward.num_workers=$REWARD_WORKERS \
     reward.reward_model.enable=False \
-    reward.custom_reward_function.path=$reward_function_path \
-    reward.custom_reward_function.name=compute_score_pickscore \
+    +reward.deployments.pickscore.backend=native \
+    +reward.deployments.pickscore.offload=$REWARD_OFFLOAD \
+    +reward.deployments.pickscore.adapter=pickscore \
```
zhtmike: `+reward.deployments.pickscore.adapter=pickscore` looks bit strange and duplicated.

examples/flowgrpo_trainer/qwen_image_edit/run_qwen_image_edit_lora_v1_npu.sh
```diff
@@ -83,10 +85,15 @@ python3 -m verl_omni.trainer.main_diffusion_v1 \
...
+    +reward.deployments.pickscore.model_path=$pickscore_model_path \
+    +reward.deployments.pickscore.placement.devices="$NATIVE_REWARD_DEVICES" \
+    +reward.reward_functions.pickscore.deployment=pickscore \
```
zhtmike: `+reward.reward_functions.pickscore.deployment=pickscore`, looks duplicated.

Sky-Trigger: yes, i will change this part

examples/flowgrpo_trainer/qwen_image_edit/run_qwen_image_edit_lora_v1_npu_engine_native_test.sh
zhtmike, line 1: better not to add the script calling `_test.sh`, if it is test, put it under `tests/`

verl_omni/reward_loop/accelerator_reward_workers.py
```diff
@@ -0,0 +1,97 @@
...
+It just binds workers to bundles in a trainer-selected accelerator pool. Both
+native deployments and legacy custom reward functions use this same mechanism.
+"""
+
+from ray.util.scheduling_strategies import PlacementGroupSchedulingStrategy
+from verl.plugin.platform import get_platform
+from verl.utils.device import get_device_name
+
+
+def build_accelerator_reward_workers(
```
zhtmike, line 26: write a args description for this

verl_omni/reward_loop/deployment.py
```diff
@@ -0,0 +1,712 @@
...
+    """
+    return bool(config.reward.reward_model.get("enable", False) or has_reward_deployments(config))
+
+
+def reward_pool_is_separate(config) -> bool:
+    """Whether the legacy reward-model role uses a dedicated resource pool."""
+    return bool(config.reward.reward_model.get("enable_resource_pool", False))
+
+
+def streaming_reward_enabled(config) -> bool:
```
zhtmike: any reward model support streaming?

verl_omni/reward_loop/deployment.py
```diff
@@ -0,0 +1,712 @@
...
+from omegaconf import OmegaConf
+from verl.experimental.reward_loop.reward_model import RewardModelManager
+from verl.single_controller.ray.base import split_resource_pool
+from verl.utils.device import get_device_id, get_device_name
+
+logger = logging.getLogger(__name__)
+
+_ENGINE_BACKENDS = {"engine"}
+_NATIVE_BACKENDS = {"native"}
+
```
zhtmike: add `__all__` for this. Move the most important class like `MultiRewardModelManager` at the top of the file, move the helper utils at the bottom.

verl_omni/reward_loop/deployment.py
```diff
@@ -0,0 +1,712 @@
...
+    """Static metadata shared with every reward-loop worker."""
+
+    name: str
+    backend: str
+    model_path: str | None
+    router_address: str | None
+    executor_config: dict[str, Any]
+
+
+class RewardDeployment(ABC):
```
zhtmike: I think we need to separate schemas to different file.

verl_omni/reward_loop/deployment.py
```diff
@@ -0,0 +1,712 @@
...
+    def wake_up(self) -> None:
+        if self.offload:
+            self.reward_model_manager.wake_up()
+
+    def sleep(self) -> None:
+        if self.offload:
+            self.reward_model_manager.sleep()
+
+
+class NativeRewardDeployment(RewardDeployment):
```
zhtmike: I still think `Deployment` is not a good name.

verl_omni/reward_loop/deployment.py
```diff
@@ -0,0 +1,712 @@
...
+            raise ValueError("Engine reward deployment replicas must be positive")
+        return replicas * values[0] * values[1] * values[2]
+
+    def _split_engine_resource_pool(self, engine_entries, base_config) -> dict[str, Any]:
+        """Backward-compatible engine-only view of the parent-pool split."""
+        engine_pools, _ = self._split_deployment_resource_pools(engine_entries, [], base_config)
+        return engine_pools
+
+    @staticmethod
+    def _validate_native_device_assignments(native_entries) -> dict[str, tuple[int, ...]]:
```
zhtmike: why don't we move the placement setting into a verl.base_config.BaseConfig, add validated it in `__post_init__`, it will looks clear.

verl_omni/reward_loop/deployment.py
zhtmike, line 1: too big file, consider separate it.

verl_omni/reward_loop/deployment.py
```diff
@@ -0,0 +1,712 @@
...
+        if not self.offload and self._resident:
+            return
+        self._run_worker_lifecycle("wake_up_reward_model")
+        self._resident = True
+
+    def sleep(self) -> None:
+        if not self.offload:
+            return
+        self._run_worker_lifecycle("sleep_reward_model")
+        self._resident = False
```
zhtmike: should we make it async? it might block they separated async training.

verl_omni/utils/reward_score/pickscore_reward.py
```diff
@@ -84,17 +106,133 @@ def score(self, prompts: list[str], images: list[Image.Image]) -> torch.Tensor:
...
+            "logit_scale": self.model.logit_scale.exp(),
+        }
+
+    def score(self, prompts: list[str], images: list[Image.Image]) -> torch.Tensor:
+        """Legacy local PickScore path retained for non-deployment callers."""
+        output = self.infer(prompts, images)
+        return _pairwise_pickscore(**output)
+
+
+class PickScoreNativeModel:
```
zhtmike, line 124: we need a guide of how to add support other reward model for this

verl_omni/utils/reward_score/pickscore_reward.py
```diff
@@ -84,17 +106,133 @@ def score(self, prompts: list[str], images: list[Image.Image]) -> torch.Tensor:
...
+    """Lifecycle-friendly PickScore inference model for native deployments.
+
+    ``RewardLoopWorker.compute_score_batch`` fans a batch out into concurrent
+    single-item calls. This model preserves the old PickScore batching
+    behavior by collecting those calls locally and running one CLIP forward for
+    up to ``_MAX_BATCH_SIZE`` items. The queue belongs to this model instance
+    (rather than module globals), so ``NativeRewardExecutor.sleep`` can stop it
+    before actor update. It returns embeddings; the configured reward function
+    remains responsible for computing PickScore.
```
zhtmike: pls clean AI debugging comment.

docs/algo/managed_reward_deployments.md
Sky-Trigger: I plan to unify them later.

docs/algo/managed_reward_deployments.md
Sky-Trigger: The PickScore reward engine uses vLLM. vLLM-Omni is only used for the actor rollout in this example. I will correct the wording.

docs/algo/managed_reward_deployments.md
Sky-Trigger: I’ll update this part.

docs/algo/managed_reward_deployments.md
Sky-Trigger: Native batches are currently split evenly across workers. Dynamic load balancing is not supported yet, and I will clarify this in the documentation.

docs/algo/managed_reward_deployments.md
Sky-Trigger: Yes. I plan to open a follow-up PR to migrate all existing reward models to the new framework.

docs/algo/managed_reward_deployments.md
Sky-Trigger: Is FSDP necessary for reward models?

docs/algo/managed_reward_deployments.md
Sky-Trigger: I will support a CPU execution path either in this PR or in a follow-up PR.

examples/flowgrpo_trainer/qwen_image_edit/run_qwen_image_edit_lora_v1_npu_engine_native_test.sh
Sky-Trigger: Got it.

verl_omni/reward_loop/accelerator_reward_workers.py
Sky-Trigger: Got it.

zhtmike:
Consider make the api async. for future separate async training. So training / rollout / rewards are all fully async, no bubbles, fully device utilization.

Sky-Trigger:
> Consider make the api async. for future separate async training. So training / rollout / rewards are all fully async, no bubbles, fully device utilization.

Got it.

docs/algo/managed_reward_deployments.md
zhtmike: because TP support needs to hack into the native reward model code. So I think FSDP is better for this?

Sky-Trigger: I will add FSDP2 feature for native reward model

Sky-Trigger:
For this PR, I plan to:
- simplify the naming and remove duplicated configuration;
- split the large deployment module and move schemas/placement validation into BaseConfig;
- make the reward lifecycle and scoring APIs async-ready, while keeping a synchronous compatibility wrapper;
- clarify native routing, streaming limitations, and the currently supported vLLM engine path;
- clean up the examples, documentation, argument descriptions, and debugging comments.

For follow-up work:
- I will first open an RFC for migrating existing reward models to a unified MultiRewardManager, and then deprecate the old reward-model paths and AudioRewardManager/VisualRewardManager;
- CPU-native execution, FSDP support, vLLM-Omni reward serving, dynamic load balancing, and full async/streaming integration will be handled in separate PRs.

zhtmike:
> For this PR, I plan to:
> 
> - simplify the naming and remove duplicated configuration;
> 
> - split the large deployment module and move schemas/placement validation into BaseConfig;
> 
> - make the reward lifecycle and scoring APIs async-ready, while keeping a synchronous compatibility wrapper;
> 
> - clarify native routing, streaming limitations, and the currently supported vLLM engine path;
> 
> - clean up the examples, documentation, argument descriptions, and debugging comments.
> 
> 
> 
> For follow-up work:
> 
> - I will first open an RFC for migrating existing reward models to a unified MultiRewardManager, and then deprecate the old reward-model paths and AudioRewardManager/VisualRewardManager;
> 
> - CPU-native execution, FSDP support, vLLM-Omni reward serving, dynamic load balancing, and full async/streaming integration will be handled in separate PRs.

I think this will be a great feature for this repo: fully actor/rollout/multi reward async + native reward life cycle management + FSDP + dynamic routing. Cc @SamitHuang @knlnguyen1802

verl_omni/utils/reward_score/pickscore_reward.py
chenyingshu: Agree, but maybe add detailed and formal instructions in next/follow-up PRs after migrating existing reward models to a unified MultiRewardManager.

docs/algo/named_reward_models.md
chenyingshu, line 1: As mentioned before, PickScore, as a lightweight model, applies `engine` backend here might not be a smart solution. 
It would be helpful to report latency and memory usage between `engine` and `native` backends.
If possible, please provide guidelines for backend selection for different rewards.

Sky-Trigger: PickScore uses the engine backend only as an example for the future migration of other models; it does not mean that the engine backend is used by default.

Sky-Trigger:
PTAL @zhtmike @chenyingshu

verl_omni/workers/config/reward.py
zhtmike, line 1: let us move the config under `verl_omni/workers/config/`, e.g, `reward.py`

verl_omni/trainer/config/reward/reward.yaml
```diff
@@ -20,14 +20,67 @@ custom_reward_function:
...
+#   quality:
+#     path: pkg://my_package.reward_score
+#     name: compute_quality_score
+#   judge:
+#     path: pkg://my_package.reward_score
+#     name: compute_judge_score
+# Existing router-based rewards (for example GenRM OCR) may also bind their
+# existing path/name function to a model; it receives that model's
+# ``reward_router_address`` and ``model_name`` instead of the legacy global
+# reward router.
```
zhtmike: clean the comment here, we put the instructions in doc.

zhtmike:
**Verdict:** the engine path is upstream-aligned, the async API shape is right, and the CPU test suite is genuinely good. Holding merge on three design items (RFC sync for the native backend, manager coupling, native placement addressing) — all are cheap to fix now and expensive to fix after configs pile up on top. The async ask from the earlier thread landed; one concurrency gap below.

---

## Blocking — design

### 1. Native backend: right call, wrong paper trail — amend RFC #432 and pin the contract now
`reward_model_executor.py:65`, `reward_model.py:227`

Native is the right call for this repo: PickScore, HPSv3, CLIP scorers and most small non-LLM reward checkpoints will never be a vLLM/SGLang serving priority. But RFC #432 — the governing design doc this PR implements — says the opposite in three places:

- "The adapter does not implement model execution, batching, scheduling, parameter sharding, or collectives."
- Non-goals: "It does not add a new inference runtime."
- Non-goals: "It does not move model weights into TaskRunner or RewardLoopWorker."

`NativeRewardExecutor` does all three (weights in reward-loop workers, manager-side batching via even split + padding, framework-side placement). The PR description doesn't mention this divergence either.

The boundary question needs answering *deliberately*, because the follow-up plan (CPU-native, FSDP) walks further in. Upstream verl's history is the cautionary tale: it deprecated and deleted its framework-hosted FSDP/Megatron reward-model workers (`cb01f10ba`, #3352; removed in #5255) because sharded model execution inside the trainer was the maintenance cost center — that's the subsystem being incrementally rebuilt if FSDP lands. The property that keeps native healthy is what this PR already does: **replica-only, user owns the model class (`executor.model`), framework owns only placement/lifecycle/dispatch**.

**Ask:** update RFC #432 in this PR to make native a first-class backend, with the contract written down: (a) replica-only — no framework-side TP/PP/FSDP, models too big for one device belong in the engine backend; (b) the user-supplied executor class owns inference; (c) `infer(*args, **kwargs)` is the deliberately untyped adapter-internal contract (fine — but say so, it's the load-bearing interface every native adapter builds against). If FSDP-native is genuinely wanted later, that's an RFC decision, not a follow-up PR.

### 2. Named models are locked to the visual manager, via a silent config rewrite
`reward_loop.py:129`

```python
with open_dict(config.reward.reward_manager):
    config.reward.reward_manager.name = "MultiVisualRewardManager"
```

Two problems:

**The rewrite breaks non-default configs.** Upstream resolves managers via `resolve_reward_manager_cls` on `source` (`register` | `importlib`) + `module.path` + `name`. Mutating only `.name` works solely for verl-omni's exact default: with `source: register` it raises "Unknown reward manager" (nothing registers that name), and with a user's `module.path` it tries to load `MultiVisualRewardManager` from the *user's* module. The PR body promises "does not silently rewrite legacy configurations" — this is a silent rewrite of the new configuration path.

**The lock-in blocks half this repo's use cases.** `reward.models` is presented as general, but scoring flows through `MultiVisualRewardManager.run_single`, which consumes the visual sample contract (`solution_image` …). Audio/omni reward models can never use named models as-is. The docs admit it ("Named-model aggregation currently uses the visual reward manager contract") — that's a limitation only if it stays.

The fix is cheap because the abstraction already exists: `VisualRewardManager` already extends upstream's modality-neutral `RewardManagerBase`, and upstream's managers (`naive`, `dapo`, `gdpo`, `limited`, `remote`) all register under modality-neutral names via `@register`.

**Ask:** (a) don't rewrite — validate: error if the configured manager is user-set or not multi-capable; (b) introduce a modality-neutral `MultiRewardManager` on `RewardManagerBase` holding the executor dispatch (`set_reward_executors`) + weighted sum + `required` policy, with visual sample extraction in a `MultiVisualRewardManager(MultiRewardManager)` subclass. Then `reward_loop.py:129` disappears.

### 3. `placement.devices` addressing is order-dependent and unrooted
`reward_model.py:140-172`, `reward.yaml`

Native indices address an *internal* native subpool that is allocated after all engine allocations, in engine config order, and sized `max(index)+1`. Consequences:

- Adding or reordering one engine model silently remaps what every native index means.
- Sparse indices strand bundles (`[0,1]` + `[4,5]` → indices 2,3 allocated, unused; plus the leftover tail pool).
- The docs need repeated disclaimers ("not physical CUDA/NPU IDs, not TP ranks") — when an abstraction needs three paragraphs of disclaimers, that's the design asking to be changed.

**Ask:** one subpool per native model (exactly what engine models already get), so `devices` are per-model, stable, and dense; the shared-native-subpool becomes an implementation detail that can vanish. Direction check for the long term: upstream async trainers are moving toward *elastic* accelerator allocation (dynamic resource scheduling #6556, idle-GPU lending #7373) — keep the static split internal so a future dynamic allocator can replace it without breaking configs.

---

## Should fix

### 4. `async_compute_rm_score` is not single-flight
`reward_loop.py` (`async_compute_rm_score`), `reward_model_executor.py:114`

The worker-side drain (`lock` + `inflight` + `idle` event, with the re-check loop in `sleep()`) is correctly built. But idle-tracking can only see requests that have *landed* on the worker. With `offload: true`: batch A finishes and enters its `finally` → `sleep()` observes idle while batch B's `compute_score_batch` RPCs are still in Ray dispatch → A destroys the model → B's infers raise "not awake". Trainers call it serially today, so it's latent — but publishing an async API is an invitation to overlap.

**Ask:** a manager-level `asyncio.Lock` around the wake/score/sleep bracket (two lines), which also matches RFC #432's "sequential execution on shared resources" stance until phase-3 admission fencing.

### 5. Config schemas and trainer predicates live in the wrong layer
`reward_loop/reward_model_config.py` (392 lines)

Upstream keeps reward schemas in `verl/workers/config/reward.py` (`RewardModelConfig` at :82, post-#5255 refactor); verl-omni already mirrors that layout with `verl_omni/workers/config/{diffusion,omni}`. And the trainer predicates (`reward_is_enabled`, `reward_role_required`, `reward_pool_is_separate`, `streaming_reward_enabled`) are trainer concerns — upstream's equivalent `need_reward_model` lives in `verl/trainer/ppo/utils.py`, not in the reward loop. Right now five trainer modules import from `verl_omni.reward_loop.reward_model_config`, inverting the layering (trainers reaching into an experimental reward-loop package).

**Ask:** schemas → `verl_omni/workers/config/reward.py`; predicates → trainer utils (or the config module's public API), so trainer imports stop crossing into `reward_loop`.

### 6. Pooling worker extension is misplaced and untracked
`reward_loop/vllm_worker.py:20`

Every vLLM colocate extension in verl-omni lives in `verl_omni/workers/rollout/vllm_rollout/` — `vLLMOmniColocateWorkerExtension` (`utils.py:35`), NPU variant (`npu_utils.py:123`). A module named `vllm_worker.py` inside `reward_loop/` misleads on both counts (not a reward-loop worker; not where vLLM extensions live). The PR body admits it's temporary ("can be removed after upstream verl handles pooling models generically") but the code carries no `TODO(owner)` + upstream issue, per repo policy for temporary workarounds. Note upstream already supports pooling RMs elsewhere (`/classify` determinism, #7027) — the generative assumption in the colocated extension is arguably an upstream bug worth filing rather than shimming forever.

**Ask:** move next to the other extensions, add `TODO(owner)` + upstream issue link, and prefer fixing upstream.

### 7. Dead code in the manager
- `reward_loop.py:87` — `OmniRewardLoopWorker.close()` is never invoked: upstream `RewardLoopWorker` has no shutdown contract and no trainer path calls it. Resident native models (`offload: false`) therefore hold accelerator memory until process exit. Wire it into trainer shutdown or drop it.
- `reward_loop.py:122` — the second `reward.reward_model.enable` raise is unreachable (`reward_model.py:52` already raised for models+enable, and accelerator-workers+enable raises on the line above). Drop it.

---

## Nits

- `reward_loop.py` `_resolve_result`/`_gather_results` double-wrap Ray refs — `ObjectRef`s are directly awaitable inside `asyncio.gather`; one layer is enough.
- `MultiRewardModelManager` exposes internals (`native_device_assignments`, `native_resource_pool`, `bind_native_workers`) for `OmniRewardLoopManager` to consume — friend-class coupling across modules. A single `create_native_worker_group(name, factory)` hook would tighten the surface.
- `resolved_offload` accepting `rollout.free_cache_engine`/`enable_sleep_mode` as aliases gives two spellings for one semantic — your own cleanup plan covers this; make sure it survives the rebase.
- `NativeManagedRewardModel._resident` is controller-side bookkeeping of worker-side state — fine while these Ray actors don't restart, worth a comment saying so.

## PR hygiene

- PR body is stale vs the code: describes `reward.deployments` while the shipped config is `reward.models`; links the guide at a fork path (`docs/algo/managed_reward_deployments.md`) while the real file is `docs/algo/named_reward_models.md`.
- No test-run evidence in the body — per contribution policy (AI-assisted PRs) add the CPU test command + results and the NPU e2e status.

## What landed well (for calibration)

- `required`/`optional` failure policy with `required=false` default preserves legacy zero-contribution semantics — the test migration is honest about the behavior change.
- Engine deployments wrapping `verl.RewardModelManager` per model is RFC phase 1 exactly as written; `ManagedRewardModel`'s async `wake_up`/`sleep` ABC is the thin backend boundary the RFC asked for.
- Async discipline beats upstream: try/finally sleep (upstream skips sleep on failure), `sleep()` gathers with `return_exceptions=True` and raises first, reverse-order sleep, sync entry raises inside a running loop instead of blocking it.
- The CPU suite covers schema validation, pool splitting/overcommit, lifecycle delegation, executor contracts, and aggregation failure modes — this is the right test strategy for infra that can't run on CI GPUs.

---

AI assistance (ZCode) was used for this review.

Sky-Trigger:
Thanks for the detailed follow-up review. I addressed the items that belong in this PR and recorded the larger migrations separately in #432.

| Review item | Status | Resolution |
|---|---|---|
| Native execution contract | **Fixed** | Documented replica, ownership, inference, scoring, placement, and current parallelism boundaries in the PR and #432 |
| Silent reward-manager rewrite | **Fixed** | Removed it; named models now explicitly validate that the current manager is `MultiVisualRewardManager` |
| Generic `MultiRewardManager` | **Deferred** | Tracked in #432 as a separate migration PR, followed by Audio/Visual manager consolidation |
| Native placement / sparse indices | **Fixed** | Each native deployment now selects independent global logical bundle indices in the parent pool; indices may be non-contiguous, do not allocate `max(index)+1`, and overlaps/out-of-range values are rejected |
| Manager-level async single-flight | **Fixed** | Added a manager lock around the complete wake/score/sleep lifecycle |
| Config layering | **Fixed** | Moved reward-model configuration under `verl_omni/workers/config/reward.py` |
| Pooling worker extension | **Fixed** | Moved it beside rollout worker extensions at `verl_omni/workers/rollout/vllm_rollout/pooling_utils.py` and added an owner TODO. Upstream verl PR #7027 is the related pooling-RM context; I could not find a dedicated upstream tracking issue |
| Dead shutdown path | **Fixed** | Removed the uncalled native executor shutdown helpers |
| Duplicate validation | **Fixed** | Removed the unreachable second `reward.reward_model.enable` check |
| Ray result awaiting | **Fixed** | Removed redundant result helpers and await `asyncio.gather()` directly |
| Lifecycle aliases | **Fixed** | Named engine models now expose only `offload`; legacy rollout sleep flags are internal compatibility wiring and conflicting per-model aliases fail fast |
| Native `_resident` bookkeeping | **Clarified** | Documented that it is controller-side lifecycle state; actor-restart recovery is outside the current native contract |
| Manager resource helper coupling | **Deferred** | Kept the already NPU-validated upstream resource-manager surface stable; a public helper can be introduced with the broader manager/resource API work |

I also fixed a regression found by the manual NPU run: synchronous scoring could call rollout sleep twice. The compatibility wrapper now delegates lifecycle ownership only to the async path.

Validation for the current head:

- 116 focused CPU tests passed in the `verl-omni:gpu` container;
- Ruff lint and format checks passed;
- the relocated pooling extension was dynamically imported successfully;
- the engine/native PickScore recipe was manually rerun successfully on the 16-NPU setup.

Deferred follow-up work is tracked in the table on #432: unified `MultiRewardManager`, Audio/Visual manager consolidation, legacy path deprecation after model migrations, CPU-native execution, native FSDP, vLLM-Omni serving, dynamic load balancing, and full async/streaming integration.

verl_omni/workers/rollout/vllm_rollout/pooling_utils.py
```diff
@@ -0,0 +1,27 @@
...
+
+from typing import Optional
+
+from verl.workers.rollout.vllm_rollout.utils import vLLMColocateWorkerExtension
+
+
+class PoolingRewardModelWorkerExtension(vLLMColocateWorkerExtension):
+    """Keep verl's generation-only model patches out of pooling workers."""
+
+    # TODO(mike): Remove this workaround once upstream verl skips its
```
zhtmike: not me

Sky-Trigger: a mistake，ha，updated

zhtmike: still there..

Sky-Trigger: I know, i'm fixing the ci now.

zhtmike: ok

docs/algo/named_reward_models.md
```diff
@@ -0,0 +1,422 @@
+# Named Reward Models
+
+Last updated: 09/14/2026
+
+> **Scope:** This is the `verl-omni` extension guide for named model-backed
+> rewards configured under `reward.models`. The upstream `verl` Reward Loop
+> guide describes the legacy single-model path based on
```
zhtmike: well, I don't think it is appropriate

zhtmike (commented):
Looks good for me, wait CI pass

Sky-Trigger:
@zhtmike  CI passed

zhtmike (approved):

PR #527, opened 2026-09-04
[1/N][omni, rollout, fsdp] feat: support multi-stage AR model training

dongbo910220:
### What does this PR do?

This PR adds model-neutral extension points for staged autoregressive omni pipelines:

- adapter-selected Transformers auto-model registration and loading;
- stage-aware prompt construction, sampling parameters, output retention, and output assembly;
- stage-selective weight synchronization and async-server collective RPC;
- FSDP handling for partially frozen models and integer DTensor buffers.

### Checklist Before Starting

- [x] Search for similar PRs. Queries: [staged AR hooks](https://github.com/verl-project/verl-omni/pulls?q=is%3Apr+is%3Aopen+staged+AR+hooks), [adapter hooks](https://github.com/verl-project/verl-omni/pulls?q=is%3Apr+is%3Aopen+adapter+hooks), and [issue 428 references](https://github.com/verl-project/verl-omni/pulls?q=is%3Apr+is%3Aopen+428+in%3Abody).

### Test

```bash
python -m pytest --noconftest -q \
  tests/trainer/omni/test_main_omni_on_cpu.py \
  tests/workers/test_omni_fsdp_engine_on_cpu.py \
  tests/workers/rollout/rollout_vllm/test_vllm_omni_strategy_on_cpu.py \
  tests/pipelines/test_qwen3_omni_thinker_adapter_on_cpu.py
# all passed

```

### API and Usage Example

Adapters keep the existing defaults and override only the behavior required by their pipeline:

```python
class MyRolloutAdapter(OmniRolloutPipelineBase):
    @classmethod
    def weight_sync_stage_ids(cls, pipeline_mode):
        return [0]

    @classmethod
    def prepare_engine_prompt(cls, prompt, *, stage_id, pipeline_mode):
        return prompt

    @classmethod
    def combine_engine_outputs(cls, outputs, *, pipeline_mode):
        return outputs[-1]
```

### Design & Code Changes

- Keep default adapter behavior compatible with existing single-output AR pipelines.
- Let a model adapter select its Transformers auto-model class without moving `from_pretrained` ownership out of the FSDP engine.
- Preserve integer DTensor buffers during FSDP2 synchronization while casting floating DTensor weights to BF16 for rollout synchronization.
- Expose stage-local request and sampling hooks, retain declared final outputs, and fail closed when a multi-stage adapter does not define a valid assembly.
- Allow adapters to move capacity limits into per-stage engine extras; once any stage overrides a capacity field, materialize both capacity fields for every stage and suppress the conflicting top-level values.
- Route collective RPC to all stages by default, with an explicit adapter hook for trainable-stage-only weight synchronization.
- Document the public extension contract and cover default and custom behavior with CPU tests.

- [x] Add unit or end-to-end test(s) to [the CI workflow](https://github.com/verl-project/verl-omni/tree/main/.github/workflows) to cover all the code.

### DIFF
--- a/verl_omni/pipelines/model_base.py
+++ b/verl_omni/pipelines/model_base.py
@@ -513,6 +513,11 @@ class OmniModelBase(ABC):
         )
 
     @classmethod
+    def peek_class(cls, architecture: str, stage: str) -> Optional[type["OmniModelBase"]]:
+        """Return the registered adapter for ``(architecture, stage)`` or ``None``."""
+        return cls._registry.get((architecture, stage))
+
+    @classmethod
     def get_class_by_name(
         cls,
         architecture: str,
@@ -756,6 +761,11 @@ class OmniRolloutPipelineBase:
     def weight_sync_stage_ids(cls, pipeline_mode="thinker_only") -> list[int] | None:
         """Return stages that receive actor weights, or all stages by default."""
         return None
+
+    @classmethod
+    def policy_stage_id(cls, pipeline_mode="thinker_only") -> int:
+        """Return the stage whose sampling parameters and logprobs define the policy."""
+        return 0
 
     @classmethod
     def get_pipeline_id(cls, pipeline_mode: str = "thinker_only") -> str:
--- a/verl_omni/pipelines/model_base.py
+++ b/verl_omni/pipelines/model_base.py
@@ -479,6 +479,7 @@ class OmniModelBase(ABC):
     """
 
     _registry: dict[tuple[str, str], type["OmniModelBase"]] = {}
+    auto_model_class: Any = None
 
     @classmethod
     def register(cls, architecture: str, stage: str = "thinker"):
@@ -545,6 +546,11 @@ class OmniModelBase(ABC):
                 f"stage={stage!r}). Registered: {registered}. "
                 f"Set ``external_lib`` to load your training adapter."
             ) from None
+
+    @classmethod
+    def register_auto_classes(cls) -> None:
+        """Register optional model-package classes with Transformers auto APIs."""
+        return
 
     @classmethod
     @abstractmethod
@@ -615,7 +621,6 @@ class OmniModelBase(ABC):
         Default implementation strips the submodules returned by
         ``get_strip_modules``.  Override to also:
 
-        - Register the model class with ``AutoModelForCausalLM``.
         - Redirect ``forward()`` and embedding accessors to the
           trainable sub-component.
         - Force ``tie_word_embeddings=False`` for FSDP compatibility.
@@ -670,6 +675,7 @@ class OmniRolloutPipelineBase:
     """
 
     _registry: dict[str, type["OmniRolloutPipelineBase"]] = {}
+    supports_async_chunk = True
 
     @classmethod
     def register(cls, model_type: str):
@@ -747,6 +753,11 @@ class OmniRolloutPipelineBase:
         return {}
 
     @classmethod
+    def weight_sync_stage_ids(cls, pipeline_mode="thinker_only") -> list[int] | None:
+        """Return stages that receive actor weights, or all stages by default."""
+        return None
+
+    @classmethod
     def get_pipeline_id(cls, pipeline_mode: str = "thinker_only") -> str:
         """Return the vLLM-Omni pipeline model_type for *pipeline_mode*.
 
@@ -800,3 +811,25 @@ class OmniRolloutPipelineBase:
             dict: Extra key-value pairs merged into the stage's engine args.
         """
         return {}
+
+    @classmethod
+    def prepare_engine_prompt(
+        cls,
+        prompt_ids: list[int],
+        model_config,
+        multi_modal_data: dict,
+        mm_processor_kwargs: Optional[dict] = None,
+    ) -> dict | None:
+        """Build an architecture-specific rollout prompt when required."""
+        return None
+
+    @classmethod
+    def combine_engine_outputs(cls, outputs: list, prompt: dict) -> tuple[Any, dict[str, Any]]:
+        """Select the policy output and collect architecture-specific fields."""
+        if not outputs:
+            raise RuntimeError("The omni rollout engine returned no outputs.")
+        if len(outputs) != 1:
+            raise NotImplementedError(
+                "An omni rollout adapter with multiple final outputs must implement combine_engine_outputs()."
+            )
+        return outputs[0], {}
--- a/verl_omni/pipelines/model_base.py
+++ b/verl_omni/pipelines/model_base.py
@@ -835,7 +835,12 @@ class OmniRolloutPipelineBase:
 
     @classmethod
     def combine_engine_outputs(cls, outputs: list, prompt: dict) -> tuple[Any, dict[str, Any]]:
-        """Select the policy output and collect architecture-specific fields."""
+        """Select the policy output and collect architecture-specific fields.
+
+        Overriding this hook opts an adapter into retaining and assembling
+        multiple final stage outputs. The default preserves single-output AR
+        behavior.
+        """
         if not outputs:
             raise RuntimeError("The omni rollout engine returned no outputs.")
         if len(outputs) != 1:
--- a/verl_omni/workers/rollout/vllm_rollout/vllm_omni_ar_strategy.py
+++ b/verl_omni/workers/rollout/vllm_rollout/vllm_omni_ar_strategy.py
@@ -11,6 +11,7 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
+import copy
 import json
 import logging
 import os
@@ -56,6 +57,14 @@ class ARStrategy(OmniStrategyBase):
     rollout_config_cls = RolloutConfig
     model_config_cls = OmniModelConfig
 
+    def __init__(self, server: Any) -> None:
+        super().__init__(server)
+        self._rollout_adapter: type[OmniRolloutPipelineBase] | None = None
+        self._pipeline_mode = "thinker_only"
+        self._rollout_output_modalities: list[str] | None = None
+        self._weight_sync_stage_ids: list[int] | None = None
+        self._stage_sampling_constraints: dict[int, dict[str, Any]] = {}
+
     def validate_configs(self) -> None:
         if self.server.config.max_model_len is None:
             self.server.config.max_model_len = self.server.config.prompt_length + self.server.config.response_length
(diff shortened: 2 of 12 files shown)


dongbo910220:
@zhtmike As discussed in #428, I split the public utility changes into two focused PRs: this PR (#527) and  #528. I also opened RFC #538 to document the public interface design. PTAL, thanks!

zhtmike:
Can you add [1/N], [2/N], ... at title so I can track PR correctly? @dongbo910220

dongbo910220:
> Can you add [1/N], [2/N], ... at title so I can track PR correctly? @dongbo910220

Done. I added `[1/N]`, `[2/N]`, and `[3/N]` to the titles of #527, #528, and #428, respectively.

zhtmike:
hold until (1) is answered and CI runs on the final head; the rest is trim-level.

1. `vllm_omni_ar_strategy.py:177` — qwen3-omni `pipeline_mode=full` now hard-fails ("multiple final pipeline outputs"): the full topology has two `final_output=True` stages (thinker text + code2wav audio in vllm-omni's `QWEN3_OMNI_PIPELINE`), and the qwen3 adapter doesn't implement `combine_engine_outputs`. Previously it deployed and generated. If full mode was already unusable for training (last output is audio, no token IDs), say that in the PR body; otherwise this is a breaking config change → [BREAKING] or a qwen3 combiner.

2. PR body says "applying the configured floating-point synchronization dtype" — `omni_impl.py:47` hardcodes `torch.bfloat16`; no configured dtype is involved. Fix the wording or wire the real dtype.

3. `vllm_omni_ar_strategy.py:133-172` — ~40 lines validating adapter classmethod returns (bool-vs-int, list-only, uniqueness). These hooks are in-repo classmethods written by adapter authors; the unknown-stage membership check is the only real footgun, trim the rest. `list(weight_sync_stage_ids)` at :159 is dead — tuples are already rejected above. Same granularity on the `prepare_engine_prompt` return checks.

4. Only `drop_label` ran on head fed7d6ee — no cpu_unit_tests / pre-commit on the final push, so "72 passed" isn't checkable. Get CI to run on the final head.

5. `vllm_omni_ar_strategy.py:366` — stashing `_verl_omni_rollout_fields` on the engine's output object to reach `process_output` only works while those objects accept new attributes; key a dict on the strategy by request_id instead.

6. `vllm_omni_ar_strategy.py:319` — `copy.deepcopy(default_sampling_params_list)` per request in the hot path; snapshot it once instead of per request.

AI assistance (ZCode) was used for this review.

zhtmike:
need @knlnguyen1802 's comment first since it touch the vllm-omni server side a lot

dongbo910220:
@zhtmike  Thanks for the detailed review. I addressed the compatibility and trim items in the latest commit:

1. Multi-output retention is now opt-in through an adapter override of `combine_engine_outputs`. Qwen3-Omni `full` keeps the existing engine path instead of failing during configuration.
2. I corrected the PR wording to state that floating DTensor weights are cast to BF16 for rollout synchronization.
3. I removed the redundant return-type, empty-list, and uniqueness checks. The strategy now keeps only topology membership and the prompt structure needed downstream.
4. The focused CPU suite passes locally (`75 passed`) and pre-commit passes (`12/12`). Final-head GitHub CI is still required before merge.
5. Rollout fields are now stored by request ID and popped in `process_output`; engine output objects are no longer mutated.
6. Default stage sampling parameters are deep-copied once, and only the policy stage is copied per request.

verl_omni/workers/rollout/vllm_rollout/vllm_omni_ar_strategy.py
```diff
@@ -56,6 +57,18 @@ class ARStrategy(OmniStrategyBase):
     rollout_config_cls = RolloutConfig
     model_config_cls = OmniModelConfig
 
+    def __init__(self, server: Any) -> None:
+        super().__init__(server)
+        self._rollout_adapter: type[OmniRolloutPipelineBase] | None = None
+        self._pipeline_mode = "thinker_only"
```
knlnguyen1802: We might not need to store it as a class attribute

verl_omni/workers/rollout/vllm_rollout/vllm_omni_ar_strategy.py
```diff
@@ -56,6 +57,18 @@ class ARStrategy(OmniStrategyBase):
...
+        self._rollout_output_modalities: list[str] | None = None
+        self._weight_sync_stage_ids: list[int] | None = None
+        self._stage_sampling_constraints: dict[int, dict[str, Any]] = {}
+        self._default_stage_sampling_params: tuple[Any, ...] | None = None
```
knlnguyen1802: Why we putting the sampling params here ? It is better if can read from some config

verl_omni/workers/rollout/vllm_rollout/vllm_omni_ar_strategy.py
```diff
@@ -56,6 +57,18 @@ class ARStrategy(OmniStrategyBase):
...
+        self._rollout_fields_by_request_id: dict[str, dict[str, Any]] = {}
+        self._policy_stage_id = 0
+        self._policy_stage_index = 0
```
knlnguyen1802, line 66: For all field in here see what can be read from config directly, it might be more simpler and clean

verl_omni/pipelines/model_base.py
```diff
@@ -746,6 +757,16 @@ def rollout_flags(cls, pipeline_mode="thinker_only") -> dict[int, dict]:
         """
         return {}
 
+    @classmethod
+    def weight_sync_stage_ids(cls, pipeline_mode="thinker_only") -> list[int] | None:
+        """Return stages that receive actor weights, or all stages by default."""
+        return None
```
knlnguyen1802, line 763: We might need some test to cover newly function added to `OmniRolloutPipelineBase`  make sure it doesnt break the existing rollout that already adapt

verl_omni/workers/rollout/vllm_rollout/vllm_omni_ar_strategy.py
dongbo910220: Removed.

verl_omni/workers/rollout/vllm_rollout/vllm_omni_ar_strategy.py
dongbo910220: Changed. Per-stage defaults are now read directly from `engine.default_sampling_params_list`, and only the policy-stage object is copied before applying request overrides.

verl_omni/workers/rollout/vllm_rollout/vllm_omni_ar_strategy.py
dongbo910220: Simplified. I removed most of the fields. The remaining fields are resolved adapter state or per-request state that is not available directly from the trainer config.

verl_omni/pipelines/model_base.py
dongbo910220: Added coverage for the existing `Qwen3OmniRolloutAdapter` inheriting the new optional hook defaults, alongside the existing thinker-only and full-mode compatibility tests.

zhtmike:
hold until CI runs on the final head; items 3–6 are trim/doc-level.

1. still no CI on head `9724549b` — only `drop_label` ran (09-07). "75 passed" stays a local claim until cpu_unit_tests / pre-commit run there. same as last round, don't merge without it.

2. PR body still says "applying the configured floating-point synchronization dtype" — you said this was corrected, it wasn't (body edited 09:35, old sentence kept). `omni_impl.py:47` hardcodes bf16; fix the wording or wire a real dtype.

3. `vllm_omni_ar_strategy.py:169-175` — per-stage capacity handling (a stage extra with `max_model_len`/`max_num_batched_tokens` hoists both fields into every stage and nulls the top-level kwargs) is new behavior with no test and no mention in the PR body. cover it or name it.

4. `vllm_omni_async_server.py:194` — `collective_rpc` override is a new deviation in the #377-owned file: stage_ids passthrough (fine, intended adaptation) plus a return-contract change vs the verl parent (parent discards the result, ours returns it). add a D2 row to #377; today's callers are exactly `{set_pending_lora_peft_config, update_weights_from_ipc}`, `list_loras` stays engine-direct.

5. `combine_engine_outputs` will see abort-synthesized outputs (empty `token_ids`) and, under [vllm-omni#6973](https://github.com/vllm-project/vllm-omni/issues/6973), possibly no outputs at all — the RuntimeError is loud and fine for now, but #428's combiner must handle both. one line in the hook docstring or the integration guide.

6. `integrating_an_omni_model.md` — when `prepare_engine_prompt` returns a prompt, `mm_processor_kwargs` is NOT auto-merged (`vllm_omni_ar_strategy.py:318`, `not adapter_prepared_prompt`); the adapter must include it itself. undocumented; will bite the first talker/minicpm-o adapter author.

verified fixed from last round: full-mode keeps single-output path (modalities=None, pinned by tests), validation trimmed, rollout fields keyed by request_id (no engine-object mutation; store and pop are adjacent sync ops, no leak), per-request copy only on the policy stage. thinker regression surface checked clean: `stage_ids=None` is byte-identical at pin `ded8934` (engine default, all stages), FSDP sync cast unchanged for all-float params, FSDP1 mixed-freeze guard unreachable for thinker (strip-based adapter + fsdp2 recipes).

AI assistance (ZCode) was used for this review.

zhtmike:
@dongbo910220 pls address the ai review, I will then trigger the CI

dongbo910220:
@zhtmike Addressed the follow-up review:

1. I will wait for the final-head CPU and pre-commit CI before requesting merge.
2. Updated the PR wording to state that floating DTensor weights are cast to BF16 for rollout synchronization.
3. Added a focused test for stage-local capacity propagation and documented the behavior in the PR description.
4. Aligned `collective_rpc` with the verl parent by discarding the engine return value while retaining strategy-selected `stage_ids`.`collective_rpc` adds strategy-selected `stage_ids` for shared-server weight-update RPCs while preserving the verl parent's `None` return contract: **INTENDED-ADAPTATION**. Current callers are `set_pending_lora_peft_config` and `update_weights_from_ipc`; `list_loras` remains engine-direct.
5. Documented that `combine_engine_outputs` implementations must handle abort outputs with empty token IDs and, pending vllm-omni#6973, an empty output list. The Qwen3-TTS-specific handling remains in #428.
6. Documented that an adapter returning a custom prompt from `prepare_engine_prompt` is responsible for propagating `mm_processor_kwargs`; the shared strategy adds them only on the default prompt path.

The related local CPU tests and pre-commit checks all passed.

dongbo910220:
@zhtmike I noticed CI falied and update the existing deploy-config CPU test fixture to implement the staged-AR adapter contract. I also reran the complete CPU unit-test suite and all 1,225 tests passed locally. Could you please add the `ready-for-ci` label again?

zhtmike (approved):

knlnguyen1802 (approved):

zhtmike:
All CI green.

Issue #543, opened 2026-09-04
[RFC] Qwen-Image DMD2 adapter and MVP validation


NancyFyong:
## Feature request

Complete the **Qwen-Image DMD2 distribution-only MVP** in #519 using the loss/config work in #520 and the concrete engine/trainer integration in #535.

Following [the review](https://github.com/verl-project/verl-omni/pull/545#issuecomment-5609385564), replace the combined DMD/DMD2 scope with one real model/algorithm path. This can land with the runtime PR, not as a third framework layer. Reuse [the Qwen prototype](https://github.com/NancyFyong/verl-omni/pull/9), but do not treat its old results as refactor acceptance.

## Motivation

The MVP is selected by **`actor_rollout_ref.model.algorithm=dmd2`** and trains from prompts. It uses a distribution-matching student objective, a generated-sample fake-score denoising objective, separate optimizers and a two-timescale update schedule.

It is **not original DMD**: there is no paired `reference_noise`/teacher-target regression, LPIPS dependency or regression data loader. Reserve `dmd` for the future original-DMD implementation; do not stack `dmd` and `dmd2` registrations on one adapter and claim both algorithms are supported. DMD2 GAN, consistency-renoise ablations, causal Wan and ODE training stages are not required for this MVP.

Training is offline in the framework sense: prompts come from the dataset and differentiable samples are generated inside the training worker. There is no RL reward/PickScore training objective, detached vLLM training rollout, PPO log-probability batch or advantage estimator.

## Qwen implementation boundary

Register the concrete `(QwenImagePipeline, dmd2)` training adapter through the existing registry and pipeline imports. The target is base Qwen-Image T2I; edit checkpoints and other pipeline class names are not implicitly supported.

Keep conditioning, packing and scheduler/forward adaptation in `verl_omni/pipelines/qwen_image_distillation/diffusers_training_adapter.py`; the engine owns sampling/optimization. Reuse Qwen helpers, not a forked forward stack or trainer architecture branches. Add only methods the engine uses, not a mandatory computer/capability/recipe hierarchy or new abstract methods for unrelated pipelines.

### Qwen flow and latent geometry

Read VAE geometry/normalization from the checkpoint. Work with normalized image latents, pack/unpack using Qwen helpers, and validate the declared image resolution and packing alignment. Do not guess a layout from arbitrary tensor dimensions or hardcode model size.

```text
x_sigma = (1 - sigma) * x0 + sigma * epsilon
v_target = epsilon - x0
[...]

Issue #520, opened 2026-09-03
[RFC] DMD2 loss and standalone configuration


NancyFyong:
## Feature request

Implement the **DMD2 loss and standalone DMD configuration** for the Qwen-Image MVP in #519. This replaces this issue's former generic-controller/immutable-plan scope, following [the review of #545](https://github.com/verl-project/verl-omni/pull/545#issuecomment-5609385564).

This is a CPU-verifiable foundation for the concrete engine/trainer and Qwen integration in #535/#543. It does not introduce a production trainer shell, a four-recipe registry, or claim runnable training by itself. Existing #545 code still follows the previous design until it is updated; prior passing tests do not certify this revision.

## Motivation

Separate **algorithm identity**, **shared mathematics**, and **execution**:

- Original DMD (`dmd`, deferred) combines distribution matching with paired teacher-trajectory regression.
- DMD2 (`dmd2`, MVP) removes that paired regression and trains the student against frozen real-score and trainable fake-score predictions with two-timescale optimization.
- Fake-score denoising MSE is not original DMD's paired generator-regression term.
- A class named `DMDLoss` represents the shared distribution-matching term, not proof that original DMD is implemented.

Use the existing loss registry and configuration conventions rather than creating parallel objective, recipe, rollout and initialization registries. Prompt-only data must not be silently accepted as original-DMD regression data, and `dmd2` must not be aliased to `dmd`.

### 1. Registered loss and pure tensor helpers

Add `DMDLoss(DiffusionLossFn)` to `verl_omni/trainer/diffusion/diffusion_algos.py`, using the existing `register_diffusion_loss`, `DiffusionLossResult`, validation and worker loss-dispatch conventions. Register the DMD2 path explicitly and update `DiffusionLossConfig` validation consistently.

Retain the pure functions actually used by Qwen DMD2 from `verl_omni/trainer/diffusion/distillation/utils.py`. Do not rename sound numerical code merely to reduce file count. Do not retain a universal distillation package as a prerequisite for calling these functions.

`DMDLoss` computes the student distribution-matching term. Reuse the existing pure fake-score MSE helper through the DMD engine's loss/result integration; neither loss function owns a model, scheduler, optimizer, data iterator or Ray lifecycle. The engine in #535 owns the student/fake computation branch.

### 2. Exact MVP mathematics

All tensors below describe the **same normalized Qwen latent space**. `x_g` retains the student graph; score-model evaluation is detached. For Qwen rectified-flow prediction:

```text
[...]

Issue #519, opened 2026-09-03
[RFC] Qwen-Image DMD2 distribution-only MVP


NancyFyong:
## Feature request

Deliver a **Qwen-Image DMD2 distribution-only MVP** through verl-omni's existing diffusion loss, FSDP engine, worker and trainer infrastructure. Separate original DMD from DMD2 rather than implementing both behind interchangeable recipe/profile names.

Following [cr-gao's review](https://github.com/verl-project/verl-omni/pull/545#issuecomment-5609385564), this revision **supersedes the framework-first, four-recipe plan**. It is a proposed target, not already-refactored #545/#546 code or approval of every choice below.

### Algorithm identities and first-release scope

| Method | Student objective and data | Delivery |
| --- | --- | --- |
| **Original DMD (DMD1)** | Distribution matching **plus paired teacher-trajectory regression**, with matching noise/targets and provenance. | Separate follow-up; reserve algorithm name `dmd`. |
| **DMD2 distribution-only** | No paired regression; independently optimized fake score, two-timescale updates and backward-simulated multistep inputs. Prompts suffice. | **MVP**, selected by `actor_rollout_ref.model.algorithm=dmd2`. |
| **DMD2 adversarial extension** | Additional real-data classification objectives, not original DMD's paired regression. | After MVP acceptance. |
| **Self-Forcing + DMD** | Blockwise student rollout with self-generated context, temporal caches and gradient truncation. | Follow-up after causal architecture/rollout validation. |
| **CausVid** | ODE-initialized causal student and bidirectional score models, with method-specific sampling. | Separately scoped follow-up. |

`DMDLoss` names the shared distribution-matching loss term; it does **not** mean that a prompt-only DMD2 run implements original DMD. Do not alias `dmd` to `dmd2`, select original DMD via a DMD2 `paper` profile, or fabricate paired targets for prompt-only data. The MVP rejects unsupported algorithm/profile combinations rather than silently falling back.

## Motivation

The earlier proposal introduced plans, generic role graphs, recipe/rollout/initialization registries, an executor facade and a separate controller before the first model integration was required to work. The revised goal is a concrete, reviewable implementation with fewer new abstractions.

DPO/NFT provide an organizational precedent: a registered loss, an algorithm-specific FSDP engine, a trainer using the existing lifecycle, and a configuration group. DMD2 still needs its own differentiable sampling and dual-optimizer computation; it does not become a preference loss or PPO replay merely by reusing that infrastructure.

[...]

PR #545, opened 2026-09-05
[trainer, cfg, tests] feat: add DMD2 loss and standalone configuration

NancyFyong:
### What does this PR do?

Rescope this existing PR to **DMD2 loss + standalone configuration**, following [cr-gao's review](https://github.com/verl-project/verl-omni/pull/545#issuecomment-5609385564) and the revised RFCs #519 / #520.

This slice registers the distribution-only DMD2 objective in the existing diffusion loss infrastructure and adds a top-level `dmd` config. It removes the earlier generic controller/plan/recipe/trainer shell. **It does not provide runnable training by itself**; stacked #546 supplies the offline runtime and Qwen-Image closed loop, including #543's validation scope.

Original DMD with paired trajectory/perceptual regression, GAN, CausVid and Self-Forcing are deferred, not aliases of `dmd2`.

### Checklist Before Starting

- [x] Rechecked issues #520/#535/#543 and open PRs with `520 in:body`, `535 in:body`, `543 in:body`, and `DMD2`. Only the existing #545/#546 implement these slices; this updates them rather than opening duplicates. [Search](https://github.com/verl-project/verl-omni/pulls?q=is%3Apr+is%3Aopen+DMD2).
- [x] Existing PR history is preserved: the narrowing is appended after the previous head, not force-pushed. Experimental later-algorithm branches and checkpoints are not migrated.

### Test

Validated at `fd76f7b649e527bf1b8aa455eb02d678b4805b3a`:

```bash
python -m pytest -o 'python_files=*_on_cpu.py' --asyncio-mode=auto tests/
# 1,338 passed

bash scripts/generate_trainer_config.sh
# Generated configuration is stable

pre-commit run --files $(git diff --name-only --diff-filter=ACMR origin/main...HEAD)
# All hooks passed, including mypy and configuration generation
```

Config-doc validation and whitespace checks also passed. The changed-file pre-commit invocation covers the complete PR diff; the local Git version cannot support pre-commit's `--all-files` discovery option.

Coverage includes loss registration and invalid inputs, stopped teacher/fake gradients, fixed-target numerical gradient checks, unequal-microbatch equivalence, exactly-once accumulation/SP scaling through the existing loss dispatcher, strict DMD config validation, and unchanged OPD inline / `one_step_off` composition. No GPU or runnable-trainer claim belongs to this slice.

### API and Usage Example

The existing `DiffusionLossFn` dispatcher selects `dmd2`; `dmd_stage` distinguishes student from fake-score denoising without PPO tensors:

```python
import torch
from verl_omni.trainer.diffusion.diffusion_algos import DMDLoss

student = torch.randn(2, 16, 64, requires_grad=True)
loss, metrics = DMDLoss.compute_loss(
    generated_x0=student,
    fake_x0=torch.zeros_like(student),
    teacher_x0=torch.ones_like(student),
    normalization_epsilon=1e-6,
)
loss.backward()
```

`DiffusionDMDConfig` is instantiated through the top-level `dmd` Hydra group. The `algorithm.trainer_type=distribution_matching` production route is added by #546, not by this loss/config-only slice.

### Design & Code Changes

- **`diffusion_algos.py`:** `DMDLoss` is registered only as `dmd2`. Student loss uses a detached, normalized fake-minus-real x0 gradient; fake-score loss is flow denoising against detached `noise - generated_x0`.
- **`distillation/utils.py`:** retain the needed pure fp32 flow/DMD2 utilities. Normalize over non-batch dimensions, count nonfinite values before sanitization, and preserve the stop-gradient surrogate boundary. Legacy method-specific equations are removed from this MVP slice.
- **`workers/config/diffusion/dmd.py`:** independent `DiffusionDMDConfig`, exported and wired through YAML/typed construction. Defaults cover ratio 2, separate role microbatches and fake optimizer, teacher CFG 4 / layer norm / space negative prompt, shifted four-step sampling, bounded score sampling, EMA and student export.
- **Existing integration:** actor loss validation recognizes `dmd2`; generated diffusion configs include the new group. Existing OPD fields, activation and scheduling remain unchanged. Deferred recipe/profile/adversarial settings are rejected rather than silently accepted.
- **Removed prototype surface:** no generic role graph, controller state machine, recipe registry, empty strategy classes or production trainer shell in PR 1.

Compatibility: the unmerged prototype's generic-distillation API is intentionally replaced. Existing mainline policy-gradient, direct-preference and OPD APIs are not repurposed. #546 documents the new runtime/checkpoint contract and explicit rejection of legacy prototype checkpoints.

- [x] Followed the contribution, configuration, pipeline and testing guides.
- [x] Complete changed-file pre-commit and relevant CPU/config tests passed.
- [x] Updated the RFC direction; runtime usage/documentation remains with #546.
- [x] Added CPU tests covering this slice, including OPD coexistence.
- [ ] **Renewed human line-by-line review of this rewrite is required before merge.** Previous human review of the generic prototype does not cover this new delta.

AI assistance: ChatGPT via Pi assisted with implementation, tests and this description. Commits retain attribution and sign-off trailers. This disclosure is not a claim that a human has reviewed the newly rewritten lines.


NancyFyong:
@zhtmike @wtomin @SamitHuang @knlnguyen1802 PTAL

zhtmike:
ping @cr-gao for review

NancyFyong:
### PR 1 design question: is a separate trainer lifecycle necessary?

I would like to discuss this boundary in **PR 1**, beyond the configuration-reuse proposal in #546. Could we reuse the existing diffusion trainer lifecycle and OPD configuration surface, while retaining only DMD-specific update-cycle logic here?

At `1ef230a`, `DistillationTrainerController` primarily implements a CPU-testable cycle state machine; it does **not** duplicate model allocation or checkpoint I/O. However, PR 1 also introduces a separate trainer shell, `run`/`fit` entrypoints, `BatchProvider`, and completed-step hooks. Are all of these new lifecycle abstractions needed, or could this be a smaller component hosted by the existing trainer infrastructure? PR #546 already makes `DistillationRayTrainer` a `BaseRayDiffusionTrainer` subclass.

**Suggested ownership:**

- **Existing trainer infrastructure:** configuration, initialization, dataloaders, the outer run/resume loop, tracking/profiling, and deciding when to checkpoint or validate. Reuse these where available and extract only concrete shared pieces where they are currently embedded in individual `fit()` methods.
- **DMD update-cycle component:** student/fake-score ordering, update ratios, warmup, per-role optimizer counters, and committing `global_step` only after the required cycle succeeds. A failed partial cycle must remain terminal until checkpoint restore; this does not imply rollback of already-updated parameters.
- **PR 2 runtime:** role-specific computation/execution and multi-role checkpoint contents. Reusing lifecycle orchestration must not reduce a student/fake-score/EMA checkpoint to the existing actor-only payload.

Likewise, use the existing `DiffusionDistillationConfig` / teacher configuration as the common input, with algorithm-specific options kept under `distribution_matching`, rather than maintaining a second teacher-configuration source.

There is a real integration task here: `BaseRayDiffusionTrainer.fit()` is abstract today, so there is no generic loop we can simply call with `super().fit()`. The proposal is to consolidate common orchestration, **not** run DMD through the PPO/OPD reward, log-probability, advantage, or teacher-transition-mean data flow. A shared outer loop also cannot assume one batch or one optimizer update per iteration, or advance the student counter during fake-only warmup.

** If a separate controller is preferable, which responsibilities should it own versus the existing base trainer? CPU-testability alone should not require a parallel production lifecycle.

This is a proposed scope/design adjustment to #520 before extending the stack; no implementation changes have been made for it yet.

cr-gao:
Nice work. Some suggestions:

Scope. Start with Qwen-Image DMD2 distribution-only, with Self-Forcing in a follow-up, rather than a runtime covering all four recipes upfront. DMD1's co-trained regression term and the standalone CausVid recipe can be deferred. ODE-regression initialization can be supplied as an input checkpoint rather than implemented as a trainer stage. The GAN head can be discussed once distribution-only is validated.

Structure. DMD can follow the same path DPO/NFT took: a loss, an engine, a trainer, a config group. Roughly:

- DMDLoss(DiffusionLossFn) in trainer/diffusion/diffusion_algos.py, keeping the DMD math from utils.py, which looks faithful.
- A DMD engine under workers/engine/fsdp/, built on DiffusersFSDPEngine. This is where the real new work is: the differentiable student rollout, the detached samples for the fake phase, and the phase switch. Whether student and fake_score share one module via named adapters or the fake score gets its own engine is your call based on what works under FSDP and activation checkpointing; either way, please build on the existing worker, optimizer/offload, and checkpoint infrastructure, using LoRAAdapterMixin where applicable and adding only the DMD-specific state and lifecycle handling needed.
- DistributionMatchingRayTrainer(BaseRayDiffusionTrainer) in ray_diffusion_trainer.py, reusing the offline init_workers/fit skeleton from DirectPreferenceRayTrainer. The 1:K cycle is a short loop there; note update_actor can run several optimizer steps per call, so the ratio needs a config constraint or a returned step count, and EMA should follow a successful student step.
- DiffusionDMDConfig in workers/config/diffusion/dmd.py as a top-level dmd: group next to distillation, and a trainer_type name that doesn't overlap with OPD, so users don't set DMD fields under an OPD group that must stay enabled=false.

With this narrower scope, could we reduce the distillation package to the pieces exercised by the Qwen-Image
implementation? The update schedule can be an explicit loop in fit, and role placement can stay local to the DMD
engine/worker integration. Reuse the existing model registry, resource-pool management, and weight-sync/checkpoint
infrastructure where applicable, keeping only the DMD-specific contracts they do not already cover. General-purpose 
role, rollout, transport, and export abstractions can wait until a second concrete integration demonstrates the need.
  
One concrete issue with the controller: a skipped student step due to non-finite gradients becomes a fatal error.
Could we handle numerical skips explicitly, keeping EMA and successful-update counters consistent, while retaining
fail-and-resume behavior for unexpected exceptions?

Reviewed with AI, happy to discuss further. @zhtmike @knlnguyen1802 @SamitHuang @wtomin PTAL.

NancyFyong:
> Nice work. Some suggestions:
> 
> Scope. Start with Qwen-Image DMD2 distribution-only, with Self-Forcing in a follow-up, rather than a runtime covering all four recipes upfront. DMD1's co-trained regression term and the standalone CausVid recipe can be deferred. ODE-regression initialization can be supplied as an input checkpoint rather than implemented as a trainer stage. The GAN head can be discussed once distribution-only is validated.
> 
> Structure. DMD can follow the same path DPO/NFT took: a loss, an engine, a trainer, a config group. Roughly:
> 
> * DMDLoss(DiffusionLossFn) in trainer/diffusion/diffusion_algos.py, keeping the DMD math from utils.py, which looks faithful.
> * A DMD engine under workers/engine/fsdp/, built on DiffusersFSDPEngine. This is where the real new work is: the differentiable student rollout, the detached samples for the fake phase, and the phase switch. Whether student and fake_score share one module via named adapters or the fake score gets its own engine is your call based on what works under FSDP and activation checkpointing; either way, please build on the existing worker, optimizer/offload, and checkpoint infrastructure, using LoRAAdapterMixin where applicable and adding only the DMD-specific state and lifecycle handling needed.
> * DistributionMatchingRayTrainer(BaseRayDiffusionTrainer) in ray_diffusion_trainer.py, reusing the offline init_workers/fit skeleton from DirectPreferenceRayTrainer. The 1:K cycle is a short loop there; note update_actor can run several optimizer steps per call, so the ratio needs a config constraint or a returned step count, and EMA should follow a successful student step.
> * DiffusionDMDConfig in workers/config/diffusion/dmd.py as a top-level dmd: group next to distillation, and a trainer_type name that doesn't overlap with OPD, so users don't set DMD fields under an OPD group that must stay enabled=false.
> 
> With this narrower scope, could we reduce the distillation package to the pieces exercised by the Qwen-Image implementation? The update schedule can be an explicit loop in fit, and role placement can stay local to the DMD engine/worker integration. Reuse the existing model registry, resource-pool management, and weight-sync/checkpoint infrastructure where applicable, keeping only the DMD-specific contracts they do not already cover. General-purpose role, rollout, transport, and export abstractions can wait until a second concrete integration demonstrates the need.
> 
> One concrete issue with the controller: a skipped student step due to non-finite gradients becomes a fatal error. Could we handle numerical skips explicitly, keeping EMA and successful-update counters consistent, while retaining fail-and-resume behavior for unexpected exceptions?
> 
> Reviewed with AI, happy to discuss further. @zhtmike @knlnguyen1802 @SamitHuang @wtomin PTAL.

good idea, We can start by supporting DMD and validating it on Qwen-Image and MiniMax.

verl_omni/trainer/diffusion/distillation/ray_trainer.py
```diff
@@ -0,0 +1,116 @@
...
+
+from verl_omni.trainer.diffusion.diffusion_trainer_utils import validate_distillation_config
+from verl_omni.trainer.diffusion.distillation.contracts import DistillationPlan
+from verl_omni.trainer.diffusion.distillation.controller import DistillationTrainerController
+from verl_omni.trainer.diffusion.distillation.recipes import build_plan_from_config
+
+__all__ = ["DistillationRayTrainer"]
+
+
+class DistillationRayTrainer:
```
wtomin: `DistillationRayTrainer` does not inferit from `BaseRayDiffusionTrainer`, so that worker initialization, checkpointing, logs, etc., cannot be reused.

Is it intentional?

verl_omni/trainer/main_diffusion.py
```diff
@@ -130,8 +130,13 @@ def _get_trainer_cls(config):
         return PolicyGradientRayTrainer
     if trainer_type == "direct_preference":
         return DirectPreferenceRayTrainer
+    if trainer_type == "distillation":
```
wtomin: How about using `main_diffusion_v1.py` because I think `main_diffusion.py` is about to be deprecated.

Also hear from @knlnguyen1802 .

verl_omni/trainer/diffusion/distillation/ray_trainer.py
```diff
@@ -0,0 +1,116 @@
...
+        self.batch_provider = batch_provider
+        self.hooks = hooks
+        self.controller_instance: Optional[DistillationTrainerController] = None
+
+    def init_workers(self) -> None:
+        """Validate the PR 1 boundary before PR 2 supplies role-group workers."""
+        if self.executor is None or self.batch_provider is None:
+            raise NotImplementedError(
+                "The multi-role distillation workers and architecture capability binding land in PR 2. "
+                "PR 1 accepts the production trainer interface but does not allocate model workers."
```
wtomin: Do not mention about PR1 and PR2, because they are vague.

verl_omni/trainer/diffusion/diffusion_algos.py
```diff
@@ -1051,6 +1051,72 @@ def __call__(
...
+        missing = [key for key in required if key not in model_output]
+        if missing:
+            raise KeyError(f"Diffusion loss `{loss_name}` is missing model_output keys: {missing}")
+
+    def __call__(self, *, config: DiffusionActorConfig, model_output: dict[str, Any], data: TensorDict):
+        """Adapt the selected engine computation to the existing loss dispatcher."""
+        self.validate_inputs(loss_name="dmd2", model_output=model_output, data=data)
+        stage = tu.get_non_tensor_data(data, "dmd_stage", default="student")
+        if stage == "student":
+            loss, metrics = self.compute_loss(
```
SamitHuang, line 1105: `compute_loss` and `dmd_surrogate_loss` already accept `gradient_mask`, and `test_distillation_utils_on_cpu.py` treats mask-only reduction as part of the math contract, but this `__call__` never forwards a mask from `model_output` or `data`. The `fake_score` branch has the same gap. Once #546 feeds padded or sequence-parallel tensors, dummy tokens will contribute to the mean unless this is wired here.

verl_omni/trainer/diffusion/distillation/utils.py
```diff
@@ -0,0 +1,158 @@
...
+    """Convert flow velocity ``epsilon - x0`` to ``x0 = x_sigma - sigma * v``."""
+    return noisy.float() - sigma.float() * velocity.float()
+
+
+@torch.no_grad()
+def dmd_gradient(
+    x0_fake: Tensor,
+    x0_real: Tensor,
+    x_g: Tensor,
+    normalization_epsilon: float = 1e-5,
```
SamitHuang: `dmd_gradient` defaults `normalization_epsilon` to `1e-5`, but `DMDLoss.compute_loss`, `DiffusionDMDConfig`, and the Hydra group all use `1e-6`. Any direct call that omits the kwarg (and any engine that forgets to pass the batch key) silently uses a 10× larger floor than the config claims.

verl_omni/workers/config/diffusion/dmd.py
```diff
@@ -0,0 +1,102 @@
...
+    # Rational shift for discrete score sampling only.
+    score_timestep_shift: float = 3.0
+    # EMA decay, applied only after successful student updates.
+    ema_decay: float = 0.999
+    # Successful student update count at which EMA starts.
+    ema_start_step: int = 0
+    # Default inference artifact; EMA is an explicit alternative.
+    export_role: str = "student"
+
+    def __post_init__(self):
```
SamitHuang, line 64: `__post_init__` does not call `super().__post_init__()`. Config subclasses here are required to chain that so parent `BaseConfig` validation is not skipped. Add the super call at the top of this method.

SamitHuang (changes requested):
Loss + standalone `dmd` group is the right slice for this PR (runtime stays in #546). Three issues before this lands: `DMDLoss.__call__` drops `gradient_mask` even though the math helpers and CPU tests treat masking as part of the contract; `dmd_gradient`’s default epsilon (`1e-5`) disagrees with config/`compute_loss` (`1e-6`); `DiffusionDMDConfig.__post_init__` must call `super().__post_init__()`.

verl_omni/workers/config/diffusion/dmd.py
NancyFyong: Thanks. Our installed `base_config.py` is identical to the pinned [`BaseConfig` implementation](https://github.com/verl-project/verl/blob/fefb080262e1c015a0ea05f958822a6a512dc795/verl/base_config.py#L21-L86). It has no `__post_init__`, so adding the super call raises `AttributeError`. Could we retain subclass validation without that call for the current pin?

SamitHuang: Yes — skip the `super()` call on this pin. The installed `verl.base_config.BaseConfig` has no `__post_init__`, so `super().__post_init__()` would raise `AttributeError`. Same pattern as `DiffusionDistillationConfig` / `DiffusionLossConfig`. Keep subclass validation only.

SamitHuang:
Re-reviewed `c7ed39cd`.

The previous request-changes items are addressed:
- `DMDLoss.__call__` now forwards `gradient_mask` from `model_output` then the batch, for both student and fake-score stages, with fail-closed all-masked tests.
- `dmd_gradient` default `normalization_epsilon` is `1e-6`, matching config / `compute_loss` / YAML.
- Skip `super().__post_init__()` on this pin: `verl.base_config.BaseConfig` has no `__post_init__`; subclass-only validation matches other diffusion configs.

SamitHuang:
Re-reviewed `c7ed39cd`.

The previous request-changes items are addressed:
- `DMDLoss.__call__` now forwards `gradient_mask` from `model_output` then the batch, for both student and fake-score stages, with fail-closed all-masked tests.
- `dmd_gradient` default `normalization_epsilon` is `1e-6`, matching config / `compute_loss` / YAML.
- Skip `super().__post_init__()` on this pin: `verl.base_config.BaseConfig` has no `__post_init__`; subclass-only validation matches other diffusion configs.

No remaining blocking issues on this slice. Runtime stays in #546.

SamitHuang (commented):
LGTM, the changes should be further verfied with #546

SamitHuang (approved):

Issue #535, opened 2026-09-04
[RFC] DMD2 FSDP engine and offline trainer integration


NancyFyong:
## Feature request

Build the **DMD2 FSDP engine and offline trainer integration** for the Qwen-Image MVP in #519, using the loss/config work in #520 and the concrete adapter/acceptance work in #543.

Following [cr-gao's review](https://github.com/verl-project/verl-omni/pull/545#issuecomment-5609385564), replace the general role-runtime/controller/executor design with a concrete DMD2 integration. Existing #546/prototypes are migration inputs, not validation of this revision.

## Motivation

DMD2 needs differentiable sampling, detached fake-stage samples and independent student/fake optimization, not another Ray lifecycle or arbitrary role/transport framework. Adapt the existing worker/engine's single-active-optimizer assumptions rather than treating DMD2 as PPO replay or an OPD request.

Only **DMD2 distribution-only** is in scope. Original DMD, GAN, causal Wan and Self-Forcing are not prerequisites.

## Proposed structure

```text
main_diffusion / existing task and resource-pool setup
  -> DistributionMatchingRayTrainer(BaseRayDiffusionTrainer)
       -> existing worker API with minimal DMD-specific integration
            -> DMDDiffusersFSDPEngine(DiffusersFSDPEngine)
                 -> Qwen architecture adapter / frozen conditioning
                 -> registered DMDLoss and fake-score denoising math
                 -> existing optimizer, offload, LoRA and checkpoint helpers
```

| Component | Responsibility |
| --- | --- |
| Trainer in `trainer/diffusion/ray_diffusion_trainer.py` | Offline lifecycle, data iteration, short `1:K` loop, counters, logging/profiling and save/resume scheduling |
| Existing `TrainingWorker`, with a narrow construction seam or thin specialization if necessary | Distributed setup, dispatch, typed config, batching/result transport and engine delegation |
| Proposed `workers/engine/fsdp/dmd_impl.py` | DMD forwards/backward, active-model/optimizer ownership, numerical update outcome, EMA and DMD-specific checkpoint state |
| Qwen pipeline adapter | Prompt encoding, latent geometry/packing, scheduler/model-input conversion and raw prediction adaptation |
| `DMDLoss` / pure math helpers | Student surrogate and detached fake-score target/loss, not execution or resource management |

Reuse `EngineRegistry` and the model registry, without requiring `DistillationPlan`, a generic controller/executor, placement graph or architecture-computer protocol.

[...]

PR #546, opened 2026-09-05
[diffusion, trainer, worker, cfg, tests] feat: add reusable DMD2 training runtime

NancyFyong:
### What does this PR do?

Add the reusable **distribution-only DMD2 training runtime** on top of #545 and RFC #535.

#545 provides the registered DMD2 loss, pure tensor utilities and standalone `dmd` configuration. This PR provides the architecture-independent runtime layer:

- `DistributionMatchingRayTrainer(BaseRayDiffusionTrainer)` with an explicit one-student-to-K-fake-score cycle;
- `DMDTrainingWorker(TrainingWorker)`, co-located with the other workers in `engine_workers.py`, using the existing distributed and mini/microbatch lifecycle;
- `DMDDiffusersFSDPEngine(DiffusersFSDPEngine)`, co-located with the other diffusion FSDP engines in `diffusers_impl.py`, with student/fake-score optimizers, frozen adapter-disabled teacher scoring, student EMA and rank-agreed numerical skips;
- DMD-owned role-specific export, complete checkpoint/resume state and DMD-local base-model provenance;
- generic CPU coverage for routing, accounting, adapter ownership, accumulation, skip and checkpoint boundaries.

The concrete Qwen-Image adapter, conditioning, recipe, inference tool and architecture tests were removed from this PR into a dependent branch. This keeps the runtime review separate from model-specific behavior.

The runtime is deliberately bounded to shape-preserving, single-latent flow-matching DMD2 with deterministic Euler student sampling. It does not claim original DMD, GAN, CausVid, Self-Forcing, multimodal schedules, causal caches, vLLM serving or arbitrary architectures through registration alone.

`algorithm.sample_source=offline` means **engine-local differentiable sampling**, not fixed pregenerated samples: fresh student samples are generated inside FSDP so the selected student forward retains autograd. No rollout/reward service is started.

### Checklist Before Starting

- [x] Rechecked RFC #535 and open PR searches. This is the existing authorized runtime PR rescope, not a duplicate contribution. [Search](https://github.com/verl-project/verl-omni/pulls?q=is%3Apr+is%3Aopen+DMD2).
- [x] The split preserves history and does not rewrite upstream `main` or later experimental branches.

### Test

Validated at `3251ab6818d9cd2888edd591424040fed8e6dfbc` in the repository-required environment with GPUs hidden:

| Check | Result |
|---|---|
| Focused DMD2 worker/engine/trainer/config tests | **106 passed** |
| Repository L1 CPU suite | **1,644 passed, 1 skipped** |
| Sphinx HTML build | Passed with the same 90 existing warnings; no warning from the DMD2 page |
| Generated trainer configs | Stable |
| Config-doc validation | Passed |
| Complete incremental changed-file pre-commit | Ruff, format, mypy, generated config, docstrings, license, device/DataProto usage, structure, naming and compilation passed |
| Whitespace / publication privacy | Passed |

Representative commands:

```bash
  tests/trainer/diffusion/test_dmd_trainer_on_cpu.py \
  tests/workers/test_dmd_engine_on_cpu.py \
  tests/workers/test_diffusers_fsdp_merged_lora_on_cpu.py \
  tests/trainer/diffusion/test_dmd_loss_on_cpu.py \
  tests/trainer/diffusion/test_distillation_utils_on_cpu.py \
  tests/trainer/diffusion/test_dmd_config_on_cpu.py

  -o 'python_files=*_on_cpu.py' tests/

python -m sphinx -b html docs /tmp/dmd2-docs
bash scripts/generate_trainer_config.sh
pre-commit run --files $(git diff --name-only --diff-filter=ACMR distillation-acceleration...HEAD)
```

No GPU test is claimed for the split runtime-only head. Real FSDP and decoded-generation acceptance belongs to the dependent architecture integration, where a concrete adapter exists.

### API and Usage Example

Runtime routing is unchanged:

```yaml
algorithm:
  trainer_type: distribution_matching
  sample_source: offline
actor_rollout_ref:
  model:
    algorithm: dmd2
    model_type: diffusion_dmd_model
```

A dependent `DiffusionModelBase` adapter registered for `(architecture, dmd2)` must supply conditioning, latent geometry, latent packing, transformer inputs, prediction-to-x0 conversion and sampling sigmas. This PR intentionally provides no runnable architecture or recipe by itself.

### Design & Code Changes

**Why an algorithm-specific trainer remains**

Existing policy-gradient and direct-preference `fit()` methods require rollout/reward/log-probability or preference-score semantics. DMD2 instead requires one student attempt followed by K fake-score attempts, with independent optimizer clocks and rank-agreed recoverable numerical skips. `DistributionMatchingRayTrainer` owns only that update order plus complete-cycle checkpoint/export orchestration; it inherits the shared dataloader, resource-pool and offline worker setup.

**Reuse rather than a parallel framework**

- `DMDTrainingWorker` remains a `TrainingWorker` subclass in `engine_workers.py` and now delegates process-group, NUMA, profiler, engine and dispatch initialization to `TrainingWorker.__init__`. The base constructor accepts an optional typed `dmd_config` and forwards it only when present, so existing worker callers retain their original engine arguments.
- `DMDDiffusersFSDPEngine` remains a `DiffusersFSDPEngine` subclass in `fsdp/diffusers_impl.py`; DMD-specific adapter restoration, FSDP1/FSDP2 clipping, rank-agreed skips and role export metadata live only in this subclass. The generic optimizer/export methods and LoRA mixin stay unchanged. The former one-purpose `dmd_worker.py` and `dmd_impl.py` modules are removed.
- Base revision/config fingerprinting is computed only by the DMD worker export RPC; `verl_omni.utils.fs` remains unchanged and exposes no DMD-specific provenance API.
- Student, fake score and EMA are named LoRA adapters on one frozen physical base; teacher scoring disables adapters. Student/fake optimizer parameter sets must be nonempty and disjoint.
- Dense nondivisible microbatch tails use TensorDict splitting and sample-weighted means. Dynamic batching remains unsupported.

**Cycles and failures**

`global_step` counts completed finite cycle budgets; student/fake counters count successful optimizer steps. A nonfinite update clears gradients and advances neither scheduler nor EMA. An unexpected exception is terminal; counters do not pretend to roll back already-applied model updates, and recovery starts from the last complete checkpoint.

**Checkpoint and export**

The standard actor FSDP shards store the physical model and student optimizer once. Per-rank DMD state adds the fake optimizer/scheduler, named RNG streams and role counters; trainer/data state stores cycle, dataloader and driver RNG state. Completeness and compatibility checks precede atomic publication/load. Semantic export writes one selected student or student-EMA PEFT artifact; teacher and fake-score weights are never inference artifacts.

- [x] Followed the contribution, config and testing guides.
- [x] Added algorithm/runtime documentation and CPU regressions.
- [x] Complete incremental pre-commit and L1 CPU suites passed.
- [x] Scope and unverified GPU/model boundaries are stated explicitly.
- [ ] Renewed human line-by-line review of the split runtime diff is required before merge.

AI assistance: ChatGPT via Pi assisted with the split, testing and description. Attribution/sign-off trailers are retained. This does not state that the newly split lines have already received renewed human review.


NancyFyong:
### Proposal: share the existing OPD configuration, keep the DMD execution flow

For #545 and this PR (RFC #519 / #535), would it make sense to consolidate further around the existing `DiffusionDistillationConfig` and `DiffusionDistillationTeacherModelConfig`, rather than maintain separate sources of teacher configuration?

At `28e2943`, the configs already share the same top-level container, but DMD still derives teacher role paths from the student model and treats `distillation.enabled` as OPD-only. The proposed change is **actual configuration reuse**, not just placing the fields in the same file.

**Suggested boundary:**

- Use `distillation.teacher_models` as the shared source of teacher identities/checkpoints, reusing teacher parsing/validation and compatible model-loading/resource-pool helpers. Resolve these settings into DMD's internal role/group plan rather than asking users to configure the same teacher twice.
- Select execution through `algorithm.trainer_type`: preserve the existing policy-gradient + OPD path, while `distillation` uses the DMD controller and multi-role runtime. Accepting `enabled=true` for DMD requires separating OPD-specific activation guards in the entrypoint/base trainer, not merely removing a validation error.
- Keep using existing actor optimizer/FSDP, trainer and data settings. Retain only DMD-specific options—fake-score optimization, update ratios, role storage and EMA—under `distribution_matching`.

Illustrative **target configuration, not supported by the current head**:

```yaml
algorithm:
  trainer_type: distillation
  sample_source: offline

distillation:
  enabled: true
  nnodes: 0
  teacher_models:
    teacher_model:
      model_path: ${actor_rollout_ref.model.path}
  distribution_matching:
    recipe: dmd2
    fake_update_ratio: 2
    role_storage: shared_base_adapters
```

Sharing configuration would **not** mean sending DMD through the PPO update loop or treating OPD's `teacher_prev_sample_mean` as DMD's arbitrary-sigma score/x0. Those computation contracts remain distinct.

For a bounded first change, I suggest a single colocated teacher. Roles may share a physical base only when their checkpoint/configuration identities are compatible; a different teacher must occupy an explicitly supported separate group, not silently use the student's base. Unsupported DMD multi-teacher routing, standalone teacher pools and OPD `one_step_off` scheduling should fail before allocation, not be silently ignored.

Compatibility tests should preserve existing OPD behavior, explicitly handle legacy DMD configs with `enabled=false`, and prove that changing the shared teacher checkpoint changes the teacher DMD actually loads and records in its plan/checkpoint identity.

**Would you support making this consolidation in #545/#546 before merge?** In particular, does sharing the teacher/configuration infrastructure while retaining the DMD-specific execution flow and subconfiguration look like the right boundary? No implementation changes have been made for this proposal yet.

wtomin:
A new general-purpose runtime is introduced by this PR, rather than a routine internal refactor. At a minimum, it should include an architecture/usage contract document.

Suggested addition:

- `docs/algo/diffusion_distillation.md`

The document should focus on explaining:

`algorithm.trainer_type=distillation`
`algorithm.sample_source=offline`
The responsibilities of `DistributionMatchingModelAdapter`
The roles of `student`, `teacher_score`, `fake_score`, and `student_ema`
`shared_base_adapters` and `colocated_independent`
The two optimizers, EMA, micro-batches, and checkpoint/resume
Current limitations: concrete model computation must be provided by an architecture adapter; the PR itself is not a directly runnable recipe

References:

- `docs/algo/diffusion_opd.md`

Refer to its algorithm documentation structure and background on diffusion distillation, but do not directly reuse its OPD semantics.

- `docs/contributing/integrating_a_new_direct_preference_algorithm_for_diffusion_model.md`

Refer to its approach to trainer routing, offline data sources, and configuration documentation.

wtomin:
The existing gpu smoke tests are insufficient to cover the failure paths described above.

The PR description explicitly states that the current head has not rerun the historical FSDP1/FSDP2 validation. In addition, the newly added tests for real multi-role scenarios depend on CUDA and model weights, while the CPU tests primarily use a fake engine. Therefore, the claimed 1,441 CPU tests do not demonstrate that the default shared-LoRA production path is functional.

verl_omni/workers/engine/fsdp/distillation_impl.py
```diff
@@ -0,0 +1,439 @@
...
+
+        if not any(binding.trainable for binding in self.role_bindings.values()):
+            self.module.requires_grad_(False)
+
+        owned_parameters: dict[int, str] = {}
+        for role, binding in self.role_bindings.items():
+            if not binding.trainable:
+                continue
+            with self.use_role(role):
+                parameters = tuple(parameter for parameter in self.module.parameters() if parameter.requires_grad)
```
wtomin: `_build_model_optimizer()` selects an adapter via with `self.use_role(role)`, but then collects parameters using `self.module.parameters()`. `set_adapter()` only switches the active adapter; it does not set `requires_grad=False` for the other adapters. As a result, the optimizers for `student`, `fake_score`, and other roles may collect the same LoRA parameters, which leads to an error in optimizer initialization..

Parameters should be filtered precisely by adapter, or all adapters other than the current role should be explicitly frozen before constructing each optimizer. In addition, a real initialization test for the default `shared_base_adapters` configuration should be added.

NancyFyong:
@zhtmike @wtomin @SamitHuang @knlnguyen1802, this is the final PR for the DMD integrated trainer; we should expedite it to facilitate the subsequent integration of Minimax. Please take a look when you have a moment. Thank you very much.

zhtmike:
I think we need ping @cr-gao to take a look?

NancyFyong:
> I think we need ping @cr-gao to take a look?

this 

> I think we need ping @cr-gao to take a look?

Sure, but my approach doesn't involve OPD. I previously tried merging it with OPD, but that required too many changes; simply adding a new trainer is relatively cleaner.

verl_omni/workers/engine_workers.py
cr-gao, line 1: For process group/NUMA/config/profiler/engine/dispatch stuff I think we can just reuse TrainingWorker.__init__ to avoid duplication.

docs/algo/diffusion_distillation.md
```diff
@@ -0,0 +1,292 @@
...
+adapter reloads into its architecture.
+
+## Validation boundary
+
+Runtime CPU tests cover routing, 1:K accounting, numerical skips, unequal
+microbatch tails, checkpoint publication/rejection, named-adapter ownership,
+EMA and semantic export. Tests use fake modules or test-local adapters so this
+slice does not gain an implicit production architecture.
+
+A dependent architecture PR must additionally provide:
```
cr-gao: I think we can remove these things in the doc.

docs/algo/diffusion_distillation.md
```diff
@@ -0,0 +1,292 @@
...
+
+The DMD2 trainer remains a thin algorithm-specific subclass because one cycle
+contains one student attempt followed by `K` fake-score attempts with separate
+optimizer clocks. Policy-gradient and direct-preference `fit()` implementations
+require rollout/reward or preference-score tensors and cannot express this update
+order without importing unrelated semantics. Dataloader, resources, worker
+initialization, FSDP and mini/microbatch machinery are inherited rather than
+reimplemented.
+
+### Required model-adapter contract
```
cr-gao: Maybe we can refer to the style of diffusion_dpo.md on the following parts.

tests/workers/test_diffusers_fsdp_merged_lora_on_cpu.py
```diff
@@ -32,6 +32,10 @@ def __init__(self):
         # Carrying a ``peft_config`` is all ``get_per_tensor_param`` needs to
         # take the LoRA branch.
         self.peft_config = {"default": SimpleNamespace(to_dict=lambda: {"r": 8})}
+        self.active_adapter = "default"
+
+    def set_adapter(self, name):
```
cr-gao: Not sure, is the change here necessary?

cr-gao:
> > I think we need ping @cr-gao to take a look?
> 
> this
> 
> > I think we need ping @cr-gao to take a look?
> 
> Sure, but my approach doesn't involve OPD. I previously tried merging it with OPD, but that required too many changes; simply adding a new trainer is relatively cleaner.

Agree, OPD can reuse the PG trainer while DMD need a separate trainer.

verl_omni/trainer/config/algorithm.py
cr-gao, line 1: Should we also update docs/examples/config.md and verl_omni/trainer/config/diffusion_trainer.yaml with links to DMD doc?

cr-gao:
The previous concerns are addressed. One remaining issue is in DMDDiffusersFSDPEngine.export_student(): with FSDP1
hybrid sharding (e.g. world_size=4, fsdp_size=2), the world-wide all_reduce(expected) counts replicated shards twice,
causing a complete adapter export to fail the size check.

Please reduce over the FSDP sharding group within the existing FSDP1 branch and add regression coverage. A single
machine with 4 GPUs is sufficient; this change can stay local to DMD export without changing the FSDP2 path.

Except this problem the PR looks good to me. @zhtmike @SamitHuang @knlnguyen1802 PTAL.

verl_omni/trainer/config/diffusion_trainer.yaml
```diff
@@ -63,10 +64,12 @@ algorithm:
   # Required when using verl.utils.omega_conf_to_dataclass to instantiate dataclass configs
   _target_: verl_omni.trainer.config.DiffusionAlgoConfig
 
-  # Trainer loop: policy_gradient (FlowGRPO, MixGRPO, …) or direct_preference (DPO, DiffusionNFT, AWM)
+  # Trainer loop: policy_gradient, direct_preference, or distribution_matching (DMD2)
+  # See docs/algo/diffusion_distillation.md for distribution-matching configuration.
```
zhtmike: do not emphasize on distlaltion

verl_omni/trainer/config/diffusion_trainer.yaml
```diff
@@ -63,10 +64,12 @@ algorithm:
...
   trainer_type: policy_gradient
 
-  # Sample provenance: online (rollout+reward engines) or offline (actor-only, precomputed batches)
+  # Sample source: online uses rollout/reward engines; offline is actor-only execution
+  # DMD2 uses offline for fresh engine-local differentiable sampling, not pregenerated images.
```
zhtmike: same

docs/examples/config.md
```diff
@@ -60,8 +60,8 @@ algorithm:
   rollout_correction: { ... }   # mirrors upstream RolloutCorrectionConfig
 ```
 
-- `algorithm.trainer_type`: Trainer loop. `policy_gradient` (FlowGRPO, MixGRPO, Flow-DPPO, …) or `direct_preference` (DPO, DiffusionNFT, AWM).
-- `algorithm.sample_source`: `online` (rollout + reward engines) or `offline` (actor-only, precomputed batches).
+- `algorithm.trainer_type`: Trainer loop. `policy_gradient` (FlowGRPO, MixGRPO, Flow-DPPO, …), `direct_preference` (DPO, DiffusionNFT, AWM), or `distribution_matching` (DMD2).
+- `algorithm.sample_source`: `online` uses rollout and reward engines. `offline` selects actor-only execution; DMD2 uses it for fresh differentiable samples inside the training engine rather than pregenerated images.
```
zhtmike: do not over-emphasize on DMD2

verl_omni/trainer/config/diffusion_trainer.yaml
```diff
@@ -33,6 +33,7 @@ defaults:
   - diffusion/distillation@distillation: diffusion_distillation
 
   # DMD2 distribution matching, separate from OPD.
+  # See docs/algo/diffusion_distillation.md for the runtime contract.
```
zhtmike: no need to mention md here.

verl_omni/trainer/diffusion/ray_diffusion_trainer.py
```diff
@@ -1105,7 +1116,9 @@ def _start_profiling(self, do_profile: bool) -> None:
                 self._controller_nsys_profile_active = True
                 controller_profile_started = True
 
-            self.actor_rollout_wg.start_profile(role="e2e", profile_step=self.global_steps)
+            self.actor_rollout_wg.start_profile(
+                role="e2e", profile_step=self.global_steps if profile_step is None else profile_step
+            )
```
zhtmike: revert the profile_step change to the seprated PR; or you may validate this change to all trainers.

verl_omni/trainer/diffusion/ray_diffusion_trainer.py
```diff
@@ -1489,6 +1502,335 @@ def fit(self):
                     self.train_dataset.on_batch_end(batch=batch)
 
 
+class DistributionMatchingRayTrainer(BaseRayDiffusionTrainer):
+    """Offline DMD2: one student attempt followed by K fake-score attempts."""
```
zhtmike: I think  DMD2 is one version of  DistributionMatchingRayTrainer, right?

may the trainer general enough in docstring and implementation

verl_omni/trainer/diffusion/ray_diffusion_trainer.py
```diff
@@ -1489,6 +1502,335 @@ def fit(self):
...
+                        "python": random.getstate(),
+                    },
+                },
+                temporary / "trainer.pt",
+            )
+            os.replace(temporary, target)
+            tracker = root / f".latest_{uuid.uuid4().hex}"
+            tracker.write_text(str(self.global_steps))
+            os.replace(tracker, root / "latest_checkpointed_iteration.txt")
+        except Exception:
```
zhtmike: move try block with well defined checkpoint utils.

Check `verl` of how is the good coding practise.

verl_omni/workers/engine/fsdp/diffusers_impl.py
```diff
@@ -1464,6 +1471,464 @@ def forward_step(self, micro_batch: TensorDict, loss_function, forward_only, ste
...
+        active = getattr(module, "active_adapters", None)
+        if callable(active):
+            active = active()
+        if active is None:
+            active = getattr(module, "active_adapter", None)
+        if isinstance(active, list | tuple):
+            if len(active) != 1:
+                raise ValueError(f"DMD2 requires one active adapter, got {active!r}.")
+            active = active[0]
+        return active or "default"
```
zhtmike: it is over-defensive, samplfy it.

verl_omni/workers/engine/fsdp/diffusers_impl.py
```diff
@@ -1464,6 +1471,464 @@ def forward_step(self, micro_batch: TensorDict, loss_function, forward_only, ste
...
+        previous = self.selected_adapter()
+        if name == "reference":
+            with self.disable_adapter():
+                yield
+            return
+        self._set_adapter(name)
+        try:
+            yield
+        finally:
+            self._set_adapter(previous)
```
zhtmike: can't we directly use the LoRAAdapterMixin?

verl_omni/workers/engine/fsdp/diffusers_impl.py
```diff
@@ -1464,6 +1471,464 @@ def forward_step(self, micro_batch: TensorDict, loss_function, forward_only, ste
...
+        self.optimizers = {}
+        self.schedulers = {}
+        self.role_parameters = {}
+        self.optimizer_steps = {"student": 0, "fake_score": 0}
+        self.skipped_steps = {"student": 0, "fake_score": 0}
+        self.generators = {}
+        self.pending_generator_states = {}
+        self.last_step_succeeded = False
+        self.forward_finite = True
+        model_config = deepcopy(model_config)
```
zhtmike, line 1522: why deepcopy?

verl_omni/workers/engine/fsdp/diffusers_impl.py
```diff
@@ -1464,6 +1471,464 @@ def forward_step(self, micro_batch: TensorDict, loss_function, forward_only, ste
...
+        allowed = {"default", "fake_score", "student_ema", "reference"}
+        if not set(model_config.policy_state_adapters).issubset(allowed):
+            raise ValueError(
+                "DMD2 manages default, fake_score and student_ema adapters; other policy states are unsupported."
+            )
+        object.__setattr__(model_config, "policy_state_adapters", ("default", "fake_score", "student_ema"))
```
zhtmike, line 1530: unreadable, make it clear.

verl_omni/workers/engine/fsdp/diffusers_impl.py
```diff
@@ -1464,6 +1471,464 @@ def forward_step(self, micro_batch: TensorDict, loss_function, forward_only, ste
...
+        sample = noise
+        for index in range(exit_index + 1):
+            prediction = self.predict(
+                "student", sample, sigmas[index], condition, geometry, grad_enabled=grad_enabled and index == exit_index
+            )
+            if index == exit_index:
+                return self.model_adapter.prediction_to_x0(
+                    sample, prediction, self.expand_sigma(sigmas[index], sample)
+                ), exit_index
+            sample = ode_euler_step(sample, prediction, sigmas[index], sigmas[index + 1])
```
zhtmike: move the detail implementaiotn into `pipelines`

verl_omni/workers/engine/fsdp/diffusers_impl.py
```diff
@@ -1464,6 +1471,464 @@ def forward_step(self, micro_batch: TensorDict, loss_function, forward_only, ste
...
+            sigma = timestep_shift(timestep, total, cfg.score_timestep_shift) / total
+            sigma = sigma.clamp(cfg.score_sigma_min, cfg.score_sigma_max)
+        else:
+            sigma = torch.rand(
+                generated.shape[0], device=generated.device, generator=self.generator("score_sigma", generated.device)
+            )
+            sigma = cfg.score_sigma_min + (cfg.score_sigma_max - cfg.score_sigma_min) * sigma
+        noise = self.noise(generated.shape, generated.device, "score_noise")
+        expanded = self.expand_sigma(sigma, generated)
+        return (1 - expanded) * generated.detach().float() + expanded * noise, noise, sigma
```
zhtmike: same.

zhtmike (commented):
1. you should doing a human review first for each PR; understand what it is doing and fix some trivial problems
2. should consider abstraction; decouple algorithm specific things; move long implementaions into easy to read sub-utilies for a good coding practise. Don't rely on AI too much.

PR #547, opened 2026-09-05
[omni, rollout, tests] fix: normalize interleaved audio-video prompts before rollout

harris26-G:
## What does this PR do?

**This PR fixes a rollout prompt-consistency bug in Qwen3-Omni: incomplete audio-video normalization leaves extra tokens after vLLM-Omni re-expansion, causing rollout to use different prompt IDs from those retained by the agent loop. Fix incomplete normalization of Qwen3-Omni interleaved audio-video prompts before vLLM-Omni rollout.**

The existing `dedup_pad_tokens` collapses consecutive runs of each modality's pad token. For an HF-expanded audio-in-video span, that leaves multiple interleaved video/audio pads and the original audio boundary tokens. When vLLM-Omni expands the video placeholder into the full audio-video sequence, the remaining tokens cause a different prompt from the one retained by the agent loop. **This prompt mismatch may lead to training–inference inconsistency between vLLM-Omni rollout and the RL actor, as they may compute token probabilities conditioned on different prompt sequences.**

This change recognizes a complete span of the form:

```text
<|vision_start|><|audio_start|>[video/audio pads]<|audio_end|><|vision_end|>
```

and restores it to:

```text
<|vision_start|><|video_pad|><|vision_end|>
```

Both audio and video pads must be present, and no other tokens may occur inside the span. Existing pad-run deduplication then handles the other modality cases as before.

**Scope: this PR fixes only the V1 `Qwen3OmniThinkerAdapter.configure_processor` path. The deprecated `verl_omni/models/transformers/qwen3_omni_thinker.py` `hf_processor` fallback is unchanged and can still retain the original interleaved-span mismatch.** This scope is also documented in the adapter docstring.

### Additional experimental CPU entry validation

A separate local harness exercised the real `SingleTurnAgentLoop`, Continuous Token builder, HF processor, `OmniStrategyBase.generate`, `ARStrategy`, and pinned vLLM-Omni multimodal processor. It used deterministic synthetic decoded media, with a processor-only engine fixture returning a fixed answer instead of running a model forward.

Public processor assets: `Qwen/Qwen3-Omni-30B-A3B-Instruct`, revision `26291f793822fb6be9555850f06dfe95f2d7e695`. Input: four 64×64 RGB frames at 1 FPS and four seconds of 16 kHz synthetic audio.

| Case | HF reference audio/video pads | Before | After |
|---|---:|---:|---:|
| Audio only | 52 / 0 | 52 / 0 | 52 / 0 |
| Video only | 0 / 8 | 0 / 8 | 0 / 8 |
| Interleaved audio-video | 52 / 8 | 54 / 9 | 52 / 8 |
| Two interleaved items | 104 / 16 | 108 / 18 | 104 / 16 |
| Interleaved media with conversation history | 52 / 8 | 54 / 9 | 52 / 8 |

After the fix, the full prompt IDs—not only pad counts—matched the HF reference in all tested cases. An additional separate-audio-plus-video case and processor-only cache hits also preserved the full IDs. These are separate harness measurements, not claims that the new helper unit tests run the whole stack.

## Relationship to existing work

- Related to #445, specifically P15: the broader token-ID double-expansion issue and the existing dedup workaround are already known. This PR does not claim to discover that general issue.
- Related to #416: the NextQA recipe enables audio-in-video. The diff reviewed for that PR does not include this interleaved-span normalization; this PR does not duplicate its dataset extraction or recipe changes.
- Searches checked: [interleaved](https://github.com/verl-project/verl-omni/issues?q=interleaved), [dedup](https://github.com/verl-project/verl-omni/issues?q=dedup), and [use_audio_in_video](https://github.com/verl-project/verl-omni/issues?q=use_audio_in_video). No identical interleaved-span fix was found in the reviewed results as of 2026-09-05.

## Design and code changes

- Add a private normalization helper in the Qwen3-Omni Thinker training adapter and call it before the existing run deduplication.
- Preserve surrounding text and media order; do not mutate the caller's list.
- Leave incomplete spans, spans containing text, unknown special-token IDs, and aliased special-token IDs unchanged by this helper.
- Resolve all six span tokens from the processor bos/eos/pad attributes instead of a hardcoded token-string table.
- Add 28 helper regression cases covering valid/multiple spans, history, idempotence, non-mutation, malformed inputs, missing attributes, conversion errors, and unknown/aliased IDs.
- Extend `test_configure_processor_binds_multimodal_pad_dedup` through the bound `processor.dedup_pad_tokens` entry with both default and deliberately renamed token strings.

## Tests run

```bash
python -m pytest -q \
  tests/pipelines/test_qwen3_omni_interleaved_dedup_on_cpu.py \
  tests/pipelines/test_qwen3_omni_thinker_adapter_on_cpu.py
```

Result after the review revision: **35 passed**, no cases deselected (28 helper cases plus 7 adapter cases).

The configured-processor test verifies interleaved collapse and ordinary pad-run deduplication in the same prompt, multiple media items, idempotence, and preservation of the input list. It runs with both default and renamed token strings.

As a separate negative control, replacing the collapse helper with an in-memory no-op made both configured-processor cases fail. This confirms the tests reject a disconnected collapse call; no source files were changed by the negative control.

```bash
python -m ruff check \
  verl_omni/pipelines/qwen3_omni/thinker_training_adapter.py \
  tests/pipelines/test_qwen3_omni_interleaved_dedup_on_cpu.py \
  tests/pipelines/test_qwen3_omni_thinker_adapter_on_cpu.py
python -m ruff format --check \
  verl_omni/pipelines/qwen3_omni/thinker_training_adapter.py \
  tests/pipelines/test_qwen3_omni_interleaved_dedup_on_cpu.py \
  tests/pipelines/test_qwen3_omni_thinker_adapter_on_cpu.py
git diff --check
```

All three checks passed. The CPU test files match the existing `tests/**/*_on_cpu.py` collection rule. Full-repository pre-commit/CI has not been run locally.

### DIFF
--- a/tests/pipelines/test_qwen3_omni_interleaved_dedup_on_cpu.py
+++ b/tests/pipelines/test_qwen3_omni_interleaved_dedup_on_cpu.py
@@ -1,0 +1,68 @@
+# Copyright 2026 verl-omni contributors
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#     http://www.apache.org/licenses/LICENSE-2.0
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+from types import SimpleNamespace
+
+import pytest
+
+from verl_omni.pipelines.qwen3_omni.thinker_training_adapter import _collapse_interleaved_audio_video_tokens
+
+
+def tokenizer(**overrides):
+    names = ("vision_start", "audio_start", "video_pad", "audio_pad", "audio_end", "vision_end")
+    mapping = {f"<|{name}|>": i for i, name in enumerate(names, 101)}
+    mapping.update(overrides)
+    return SimpleNamespace(unk_token_id=0, convert_tokens_to_ids=lambda tok: mapping.get(tok, 0))
+
+
+SPAN = [101, 102, 103, 103, 104, 104, 103, 104, 105, 106]
+RAW = [101, 103, 106]
+
+
+@pytest.mark.parametrize("prefix,suffix", [([], []), ([7, 7], [8, 8]), ([1, 2, 3], [4, 5, 6])])
+def test_complete_interleaved_span_preserves_surrounding_tokens(prefix, suffix):
+    original = prefix + SPAN + suffix
+    saved = original.copy()
+    actual = _collapse_interleaved_audio_video_tokens(original, tokenizer())
+    assert actual == prefix + RAW + suffix
+    assert original == saved
+    assert _collapse_interleaved_audio_video_tokens(actual, tokenizer()) == actual
+
+
+def test_multiple_media_and_history_keep_order():
+    original = [7] + SPAN + [8, 9, 10] + SPAN + [11]
+    assert _collapse_interleaved_audio_video_tokens(original, tokenizer()) == [7] + RAW + [8, 9, 10] + RAW + [11]
+
+
+@pytest.mark.parametrize(
+    "tokens",
+    [
+        [],
+        [7, 7, 8],
+        RAW,
+        [101, 103, 103, 106],
+        [102, 104, 104, 105],
+        SPAN[:-1],
+        SPAN[:-2],
+        [101, 102, 103, 7, 104, 105, 106],
+        [101, 102, 103, 105, 106],
+        [101, 102, 104, 105, 106],
+        [101, 102, 105, 106],
+        [101, 102, 103, 104, 106, 105],
+    ],
+)
+def test_non_interleaved_or_malformed_span_is_unchanged(tokens):
+    assert _collapse_interleaved_audio_video_tokens(tokens, tokenizer()) == tokens
+
+
+@pytest.mark.parametrize("value", [None, 0, 101])
+def test_unknown_or_aliased_special_token_is_unchanged(value):
+    assert _collapse_interleaved_audio_video_tokens(SPAN, tokenizer(**{"<|audio_end|>": value})) == SPAN
--- a/tests/pipelines/test_qwen3_omni_interleaved_dedup_on_cpu.py
+++ b/tests/pipelines/test_qwen3_omni_interleaved_dedup_on_cpu.py
@@ -15,12 +15,23 @@ import pytest
 
 from verl_omni.pipelines.qwen3_omni.thinker_training_adapter import _collapse_interleaved_audio_video_tokens
 
+TOKEN_ATTRS = (
+    "vision_bos_token",
+    "audio_bos_token",
+    "video_token",
+    "audio_token",
+    "audio_eos_token",
+    "vision_eos_token",
+)
 
-def tokenizer(**overrides):
-    names = ("vision_start", "audio_start", "video_pad", "audio_pad", "audio_end", "vision_end")
-    mapping = {f"<|{name}|>": i for i, name in enumerate(names, 101)}
+
+def processor(**overrides):
+    # Deliberately not the default HF strings: the processor owns these names.
+    tokens = {attr: f"custom_{attr}" for attr in TOKEN_ATTRS}
+    mapping = {token: i for i, token in enumerate(tokens.values(), 101)}
     mapping.update(overrides)
-    return SimpleNamespace(unk_token_id=0, convert_tokens_to_ids=lambda tok: mapping.get(tok, 0))
+    tokenizer = SimpleNamespace(unk_token_id=0, convert_tokens_to_ids=lambda tok: mapping.get(tok, 0))
+    return SimpleNamespace(tokenizer=tokenizer, **tokens)
 
 
 SPAN = [101, 102, 103, 103, 104, 104, 103, 104, 105, 106]
@@ -31,15 +42,15 @@ RAW = [101, 103, 106]
 def test_complete_interleaved_span_preserves_surrounding_tokens(prefix, suffix):
     original = prefix + SPAN + suffix
     saved = original.copy()
-    actual = _collapse_interleaved_audio_video_tokens(original, tokenizer())
+    actual = _collapse_interleaved_audio_video_tokens(original, processor())
     assert actual == prefix + RAW + suffix
     assert original == saved
-    assert _collapse_interleaved_audio_video_tokens(actual, tokenizer()) == actual
+    assert _collapse_interleaved_audio_video_tokens(actual, processor()) == actual
 
 
 def test_multiple_media_and_history_keep_order():
     original = [7] + SPAN + [8, 9, 10] + SPAN + [11]
-    assert _collapse_interleaved_audio_video_tokens(original, tokenizer()) == [7] + RAW + [8, 9, 10] + RAW + [11]
+    assert _collapse_interleaved_audio_video_tokens(original, processor()) == [7] + RAW + [8, 9, 10] + RAW + [11]
 
 
 @pytest.mark.parametrize(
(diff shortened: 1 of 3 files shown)


verl_omni/pipelines/qwen3_omni/thinker_training_adapter.py
SamitHuang, line 1: HF `Qwen3OmniMoeProcessor.replace_multimodal_special_tokens` builds this span from `vision_bos_token` / `audio_bos_token` / `video_token` / `audio_token` / `audio_eos_token` / `vision_eos_token`, while `_dedup_pad_tokens` already resolves pads via those processor attributes. A second hardcoded `<|vision_start|>` table can silently no-op if the strings diverge.

Also, the new helper is only tested in isolation. The production path is `processor.dedup_pad_tokens` (called from `_preprocess_input`). Extend `test_configure_processor_binds_multimodal_pad_dedup` with an interleaved span so dropping the call on line 154 would fail CI.

The twin `_dedup_pad_tokens` in `verl_omni/models/transformers/qwen3_omni_thinker.py` is unchanged, so any Qwen3-Omni path still going through `hf_processor` keeps the old mismatch.

tests/pipelines/test_qwen3_omni_interleaved_dedup_on_cpu.py
SamitHuang, line 1: These cases only call `_collapse_interleaved_audio_video_tokens` directly, so they still pass if `_dedup_pad_tokens` never invokes it. Add one case through `configure_processor` / `processor.dedup_pad_tokens` with `SPAN` that asserts `RAW` after both collapse and run-dedup. That is the path `vllm_omni_async_server._preprocess_input` actually runs.

SamitHuang (changes requested):
The HF audio-in-video span shape matches `Qwen3OmniMoeProcessor.replace_multimodal_special_tokens`, so collapsing a complete `vision_start+audio_start+[pads]+audio_end+vision_end` run back to a single video placeholder is the right workaround for vLLM-Omni re-expansion.

Before this is mergeable:
1. Resolve bos/eos/pad IDs from the processor attributes the helper already has on `self`, not a second hardcoded `<|name|>` table.
2. Cover `processor.dedup_pad_tokens` (the `_preprocess_input` path), not only the private collapse helper.
3. Either apply the same collapse on `verl_omni/models/transformers/qwen3_omni_thinker.py` or document that only the V1 adapter path is fixed.

verl_omni/pipelines/qwen3_omni/thinker_training_adapter.py
harris26-G: Thanks for the concrete review. Addressed in dc0667c:

1. The helper now receives the processor and resolves the six tokens through `vision_bos_token`, `audio_bos_token`, `video_token`, `audio_token`, `audio_eos_token`, and `vision_eos_token`, using its tokenizer. There is no second hardcoded token-string table. Missing attributes and unresolvable/unknown/aliased IDs leave the span unchanged.
2. Extended `test_configure_processor_binds_multimodal_pad_dedup` through `configure_processor` and the bound `processor.dedup_pad_tokens`, covering interleaved collapse plus ordinary run-dedup, multiple media items, idempotence, and non-mutation. Both default and deliberately renamed token strings are tested.
3. I chose the explicit V1-only scope option. The adapter docstring states that the deprecated `hf_processor` fallback remains unchanged; the PR description is being updated to make this limitation explicit as well.

Both  test files pass in full: **35 passed**, with no cases deselected. Ruff lint/format and diff checks passed. The separate  processor-entry harness also passed full-token-ID and processor-cache assertions.

tests/pipelines/test_qwen3_omni_interleaved_dedup_on_cpu.py
harris26-G: Added the requested regression to `test_configure_processor_binds_multimodal_pad_dedup` in dc0667c. It calls `configure_processor` and then the bound `processor.dedup_pad_tokens`, asserting the raw video placeholder after interleaved collapse and ordinary run-dedup in the same prompt.

The test is parameterized over default and renamed processor token strings. As a separate negative control, replacing the collapse helper with a no-op makes both entry cases fail, confirming that these tests detect a missing collapse invocation. The two test files pass in full: **35 passed**.

zhtmike:
pls 1. run a e2e with reward_curve.png, thanks. 2. check the Pearson correlation before and after this fix.

harris26-G:
Thanks for the feedback. I completed the end-to-end before/after runs and a separate fixed-continuation comparison.

### 1. End-to-end training: 50 steps per arm

Both runs used the same initial Qwen3-Omni checkpoint, dataset, seed, and training configuration:

- 32 H20 GPUs; rollout TP=4.
- Seed=42, learning rate=5e-7.
- 32 prompts per step, 4 responses per prompt.
- 256 audio-video QA training examples, with rule-based answer scoring and no reward-model server.
- Both arms completed all 50 steps.

| Metric | Before | After |
|---|---:|---:|
| Mean training reward | 0.818750 | 0.821563 |
| **Probability Pearson** | **0.990668** | **0.990770** |
| Log-probability MAE ↓ | 0.094820 | 0.093691 |

The observed mean training reward increased by **0.0028125**, and log-probability MAE decreased by approximately **1.19%**.

Pearson values above are computed over all valid sampled response tokens pooled across the 50 steps, rather than averaging per-step correlations. Probability Pearson compares `exp(actor_logprob)` with `exp(rollout_logprob)`; log-probability Pearson compares the log probabilities directly.

The figure below shows the per-step reward and probability Pearson curves.

50-step training reward and probability Pearson

### 2. Controlled comparison: frozen weights and identical continuations

To isolate the prompt-normalization change from differences in generated answers, I also ran a separate comparison with:

- Learning rate set to zero.
- **128 identical response continuations**, covering **7,984 scored tokens**.
- The same resident vLLM instance for each before/after pair and the same Actor reference.
- Both arms scored through prefill `prompt_logprobs` / `raw_logprobs`.
- No additional M-RoPE or kernel fixes.

| Metric | Before | After |
|---|---:|---:|
| Engine input IDs exactly match Actor input IDs | 0/128 | **128/128** |
| **Probability Pearson** | **0.992460** | **0.992551** |
| Log-probability Pearson | 0.990351 | 0.990945 |
| Probability MAE ↓ | 0.024495 | 0.023943 |
| Log-probability MAE ↓ | 0.092626 | 0.090682 |

In this paired comparison, log-probability MAE decreased by **2.10%**, and probability MAE decreased by **2.25%**.

**The fix restored exact engine/Actor input-ID agreement for all 128 tested continuations.** The probability-alignment improvement is modest; the primary correctness result is removal of the prompt mismatch.

The training runs generate their own responses independently, while the frozen-weight experiment compares exactly the same continuations.

Frozen-weight same-continuation comparison

harris26-G:
@zhtmike @SamitHuang The requested 50-step end-to-end results, reward/Pearson curves, and fixed-continuation comparison are now available above. Could you please take another look? Thanks!

zhtmike:
what script are you using, avqa should have Pearson coefficient like 0.998 as baseline
see https://wandb.ai/verl-omni/gspo_demo/runs/kzzrq9pr

harris26-G:
@zhtmike Thanks for the reference. One important clarification: **our runs used an internally SFT-fine-tuned Qwen3-Omni checkpoint for our application, rather than the official `Qwen3-Omni-30B-A3B-Instruct` checkpoint in your reference run.** Both before/after arms started from exactly the same SFT checkpoint.

We used a custom A/B launcher through `verl_omni.trainer.main_omni`. After checking your W&B configuration, I found several differences:

| Setting | Your reference run | Our A/B runs |
|---|---|---|
| Initial checkpoint | Qwen3-Omni-30B-A3B-Instruct | Application-specific SFT checkpoint |
| Hardware | 4 × H800 | 32 × H20 |
| Rollout TP | 2 | 4 |
| Dataset | `avqa_r1_6k` | 256 AVQA examples selected from the “Both” modality category, with audio and video present |
| Prompts × responses per step | 128 × 16 | 32 × 4 |
| Learning rate | `3e-6` | `5e-7` |
| Maximum response length | 12,288 | 128 |
| `enforce_eager` | `false` | `true` |

Our dataset wrapper explicitly enables `use_audio_in_video=True` to exercise the interleaved audio-video normalization path in both arms. Both runs use FSDP2, BF16, LoRA rank 32 / alpha 64, and GSPO.

The reported probability Pearson is calculated as `corr(exp(actor_logprob), exp(rollout_logprob))` over valid sampled response tokens. The pooled 50-step baseline is **0.990668**; averaging the per-step correlations gives **0.990643**, so the aggregation difference alone does not explain the gap to approximately 0.998.

**These configuration differences mean our baseline is not a direct reproduction of your reference run; we have not isolated which differences account for the lower Pearson.**

Separately, the frozen-weight, identical-continuation comparison verifies the specific correctness issue: engine/Actor input-ID agreement improves from **0/128 to 128/128**, with log-probability MAE decreasing from **0.092626 to 0.090682**.

zhtmike:
I will take a look, thanks, majorly to check if it has any regression to current v1 qwen3-omni avqa script. Appreciated if you can help me to check that, too.

zhtmike:
@harris26-G I have checked it, it is correct but it will not affect our current training scripts. I will keep it open for now.

But isn't better to submitted a PR to vllm-omni instead?

Issue #406, opened 2026-08-18
[RFC]: MiniCPM-o 4.5 thinker-only text GSPO (collocated)


zhtmike:
Status: **ready to implement** — the [#363](https://github.com/verl-project/verl-omni/pull/363) prerequisite has merged; work against current `main`.
Revised **2026-09-09** against `main` @ `ef583165`: algorithm switched **GRPO → GSPO** (deliberate, see §3); pins and file references refreshed for the vLLM-Omni **0.28** pin bump ([#497](https://github.com/verl-project/verl-omni/pull/497)) and the raised `transformers` floor.
Audience: the implementation PR. This is the design, not a roadmap.
Related but **not** this work: [verl-omni#345](https://github.com/verl-project/verl-omni/issues/345) (full-duplex post-training). #345 assumed a working half-duplex / text RL baseline and never specified one. This RFC is that baseline. Do not expand this PR into duplex, speech, or vision RL.
Adjacent in-flight work, also not this PR: [#550](https://github.com/verl-project/verl-omni/pull/550) drafts MiniCPM-o **offline DPO** on the same V1 engine — it contributes the MiniCPM training-adapter mechanisms this RFC builds on (§5.1) and must share **one** adapter package with us; [#565](https://github.com/verl-project/verl-omni/issues/565) / [#566](https://github.com/verl-project/verl-omni/issues/566) track MiniCPM-o 4.5 on-policy distillation (simplex / streaming full-duplex; their tables still cite this RFC as the "GRPO baseline" — stale after this retitle).
Papers: [MiniCPM-o 4.5](https://arxiv.org/abs/2604.27393) (§5.4 RL, §6.4 text eval); [GSPO](https://arxiv.org/abs/2507.18071) (Group Sequence Policy Optimization).

---

## 1. Goal

Ship a **thinker-only, text-in / text-out, collocated GSPO** path for `openbmb/MiniCPM-o-4_5` on the existing V1 omni trainer.
Exit criterion (all of these, on 2×80GB):

1. Training + rollout adapters register and load the 4.5 checkpoint with `trust_remote_code=True`.
2. A short GSM8K GSPO run completes (`main_omni`, `omni_sync`, LoRA).
3. Actor and rollout return the same action token IDs / positions for a fixed prompt.
4. Weight sync after the update; a second rollout is not bit-identical to the first (LoRA actually landed in vLLM-Omni).

[...]

PR #550, opened 2026-09-07
[omni, recipe, tests] feat: add MiniCPM-o offline DPO training

wtomin:
## Summary

Add MiniCPM-o (thinker / multimodal understanding) **offline DPO** on the existing `main_omni` + `OfflineMLLMDPODataset` path. Training is LoRA on the AR LLM; vision/audio encoders (`vpm`, `apm`) are frozen; generation-only audio/TTS modules (talker, codec, TTS, …) are stripped. Batches are **image-only** or **audio-only** (no video parquet).

This is not a duplicate of an existing open PR: `gh pr list --repo verl-project/verl-omni --state open --search MiniCPM` / `minicpm DPO` returned no matches. Closest related work is Qwen3-Omni offline DPO; this PR reuses that dataset/trainer and adds a MiniCPM transform + adapter.

AI assistance (Cursor) was used to draft this description. A human submitter must review every changed line before merge.

## What changed

MiniCPM-o cannot use `AutoModelForMultimodalLM` or the Qwen3-Omni transform. This PR:

1. Registers `MiniCPMThinkerAdapter` (`architecture=MiniCPMO`, stage `thinker`).
2. Loads via remote-code `AutoModel.from_pretrained` (`build_module`).
3. Routes parquet through `data.base_transform=minicpm`.
4. Adds optional FSDP2 “ignore by module name” so frozen `apm` is not `fully_shard`’d (leftover params).
5. Forwards `override_config` into `AutoConfig.from_pretrained` so `init_tts=false` / `use_cache=false` / `stream_input=false` actually land on HF config.

### New

| File | Role |
| --- | --- |
| `verl_omni/pipelines/minicpm/thinker_training_adapter.py` | Core adapter: `AutoModel` load, strip TTS/talker/codec, freeze `vpm`/`apm`, pack HF kwargs into MiniCPMO `forward(data, **llm_kwargs)`, WhisperAttention 3-tuple patch, skip dummy vision forward, clone embeddings before vision scatter (LoRA-safe), FSDP ignore `apm`. |
| `verl_omni/pipelines/minicpm/__init__.py` | Exports `MiniCPMThinkerAdapter`. |
| `verl_omni/utils/dataset/minicpm_transform.py` | Compact `<image>`/`<audio>` → MiniCPM slots (`<image>./</image>`, `<audio>./</audio>`), processor call, labels, image/audio tensors. **No video path.** |
| `verl_omni/models/transformers/remote_code_compat.py` | transformers ≥5: remote MiniCPM-o omits `post_init()`, so `from_pretrained` fails; wrap `__init__` to call `post_init()` when `all_tied_weights_keys` is missing. |
| `examples/dpo_trainer/minicpm/README.md` | How to convert Omni-Preference parquet and launch MiniCPM DPO. |
| `examples/dpo_trainer/minicpm/run_minicpm_omni_preference_lora.sh` | 4-GPU LoRA DPO recipe: image+audio parquet, `base_transform=minicpm`, `init_tts=false`, exclude frozen/gen modules, FSDP2. |
| `tests/pipelines/test_minicpm_adapter_on_cpu.py` | Adapter CPU tests (kwarg packing, strip/freeze, Whisper patch, `build_module` hook). |
| `tests/utils/test_minicpm_transform_on_cpu.py` | Slot rewrite + `process_minicpm_sample` CPU tests. |
| `tests/models/test_remote_code_compat_on_cpu.py` | `post_init` shim CPU tests. |

### Modified

| File | Role |
| --- | --- |
| `verl_omni/pipelines/__init__.py` | Register the `minicpm` pipeline package. |
| `verl_omni/pipelines/model_base.py` | Optional hooks: `build_module` (default `None` → existing AutoModelForMultimodalLM path) and `get_fsdp_ignored_module_names` (default `[]`). |
| `verl_omni/utils/dataset/offline_mllm_dpo_dataset.py` | `base_transform` in `{minicpm, minicpm_o, minicpm_v}` skips Qwen RoPE/`get_rope_index`; numbered placeholders `<image_00>`; `extra_info` always has `index` + `dataset_index`. |
| `verl_omni/utils/fsdp_utils.py` | `fsdp_name_is_ignored` + `apply_fsdp2_excluding_module_names` for FSDP2 leftover params. |
| `verl_omni/workers/engine/fsdp/omni_impl.py` | Prefer `adapter.build_module`; if adapter reports ignored names, wrap FSDP2 apply. |
| `verl_omni/workers/config/omni/model.py` | Pass **all** `override_config` keys into `AutoConfig.from_pretrained` (not only `attn_implementation`). |
| `tests/utils/test_offline_mllm_dpo_dataset_on_cpu.py` | Numbered MiniCPM placeholders + minicpm transform does not require Qwen RoPE. |
| `tests/workers/test_omni_fsdp_engine_on_cpu.py` | `build_module` override skips `AutoModelForMultimodalLM`; existing tests stub `build_module → None`. |

## New docs (README)

`examples/dpo_trainer/minicpm/README.md` covers:

- Understanding-only DPO: train vision/audio **understanding** + AR LLM; exclude talker/codec/TTS/decoder/code2wav.
- **Two batch kinds only:** `image/*.parquet` and `audio/*.parquet` with standalone `audios` + `<audio>` placeholders.
- Data: `omni_preference_dpo_multisource.py --modalities image audio`; schema is model-agnostic; MiniCPM behavior is `data.base_transform=minicpm`.
- Keep compact markers (`<image>`, `<audio>`); transform rewrites to processor slots.
- Launch: `MODEL_PATH=openbmb/MiniCPM-o-4_5` + the LoRA script; `init_tts=false` via HF config override.

Launch-script defaults (also in README): train files = image+audio only; `ModalityGroupedBatchSampler` weights `{image, audio}` only.

## Possible breaking changes

**No `[BREAKING]` prefix** if existing Qwen3-Omni DPO recipes only set `override_config.attn_implementation`. Still review:

1. **`OmniModelConfig.override_config` semantics (highest risk).** Previously only `attn_implementation` was passed to `AutoConfig.from_pretrained`. Now the **entire** dict is splatted. Keys that are not valid HF config / `from_pretrained` kwargs will raise. MiniCPM needs this for `init_tts` / `use_cache` / `stream_input`. Qwen recipes that stuffed non-HF keys into `override_config` will break.

2. **`OfflineMLLMDPODataset.extra_info`.** Always injects `dataset_index` and fills `index` if missing. Extra key only; dump/log code that assumes an exact key set may need a glance.

3. **Media placeholder regex.** `<(image\|video\|audio)>` → `<(image\|video\|audio)(?:_\d+)?>`. Numbered tokens like `<image_00>` now parse as media, not text. Intended for MiniCPM; Qwen rows using those strings as literal text would change.

4. **Default `base_transform` is still `qwen3_omni_moe`.** New values: `minicpm` / `minicpm_o` / `minicpm_v`. Unsupported names still error.

5. **New `OmniModelBase` classmethods** have safe defaults (`build_module` → `None`, ignored names → `[]`). Other omni adapters unchanged unless they override.

6. **FSDP2 ignore path** only runs when the adapter returns ignored names (MiniCPM: `apm`). Other models keep `apply_fsdp2`.

Not API-breaking but operational: MiniCPM **does not train video parquet**; do not point `TRAIN_FILES` at `video/*.parquet`.

### CPU (no weights)

```bash
pytest -q \
  tests/pipelines/test_minicpm_adapter_on_cpu.py \
  tests/utils/test_minicpm_transform_on_cpu.py \
  tests/models/test_remote_code_compat_on_cpu.py \
  tests/utils/test_offline_mllm_dpo_dataset_on_cpu.py \
  tests/workers/test_omni_fsdp_engine_on_cpu.py

```

### Performance

image
image

### Convergence

**Training**
image

image

image

**Validation**

image

image

image


examples/dpo_trainer/minicpm/README.md
```diff
@@ -0,0 +1,54 @@
...
+MiniCPM's remote-code processor supports two training batch types only:
+
+1. **Image-only** batches from `image/*.parquet` rows.
+2. **Audio-only** batches from `audio/*.parquet` rows with standalone `audios`
+   paths and `<audio>` prompt placeholders.
+
+
+## Data
+
+Convert Omni-Preference into the offline MLLM DPO parquet schema:
```
SamitHuang, line 23: Can provide the link for Omni-Preference dataset?

SamitHuang (changes requested):
## Review (REQUEST CHANGES)

I agree with Copilot on `image_bound` tensor truthiness and the FSDP2 `TypeError` fail-open. Additional issues below. I could not attach pending line comments from this session (`verl-project` blocks classic PATs; GitHub’s `/changes` UI did not expose a line-comment control), so these are cited by file:line on `wtomin:minicpm-dpo`.

### Blocking

**1. Frozen `vpm` is unused on audio-only batches, but FSDP ignore only lists `apm`**
[`thinker_training_adapter.py` L132–153](https://github.com/wtomin/verl-omni/blob/minicpm-dpo/verl_omni/pipelines/minicpm/thinker_training_adapter.py#L132-L153) skips dummy `vpm` when `pixel_values` is empty. Audio-only DPO therefore never touches `vpm`. [`get_fsdp_ignored_module_names` L381–383](https://github.com/wtomin/verl-omni/blob/minicpm-dpo/verl_omni/pipelines/minicpm/thinker_training_adapter.py#L381-L383) still returns only `["apm"]`. That is the leftover-param footgun for the modality this recipe actually trains. Ignore both encoders (`["apm", "vpm"]`) or keep a dummy `vpm` path consistent with dummy `apm`.

**2. `apply_fsdp2_excluding_module_names` swallows `TypeError` and shards the ignored params anyway**
[`fsdp_utils.py` L103–110](https://github.com/wtomin/verl-omni/blob/minicpm-dpo/verl_omni/utils/fsdp_utils.py#L103-L110): if this PyTorch build rejects `ignored_params`, the `except TypeError` retries `fully_shard(model, **fsdp_kwargs)` with no ignore list. The adapter then believes `apm` is replicated. Fail closed (re-raise with a version hint) instead of silently dropping the exclusion. Add a CPU test that a `TypeError` does not call `fully_shard` without `ignored_params`.

**3. `image_bound = data.get("image_bound") or [[]]` is illegal for a 2-D tensor**
[`thinker_training_adapter.py` L194](https://github.com/wtomin/verl-omni/blob/minicpm-dpo/verl_omni/pipelines/minicpm/thinker_training_adapter.py#L194). `_sample_bound` already materializes `(N, 2)` tensors. `bool(tensor)` raises `RuntimeError: Boolean value of Tensor with more than one value is ambiguous` as soon as collation stops using a Python list. Use an explicit empty check (`None` / `[]`), not `or`.

**4. PR title fails `check_pr_title.py`**
Current: `[omni, recipe, tests feat: ...` — missing `]` before the type. Required: `[omni, recipe, tests] feat: add MiniCPM-o offline DPO training`.

### Non-blocking

**5. Global monkeypatch of `transformer_impl.apply_fsdp2`**
[`omni_impl.py` L238–259](https://github.com/wtomin/verl-omni/blob/minicpm-dpo/verl_omni/workers/engine/fsdp/omni_impl.py#L238-L259) mutates a module global around `super()._build_fsdp_module`. If wrap raises, `finally` restores it, but concurrent engine init or another import of `apply_fsdp2` still sees the swap. Prefer a first-class hook (engine-config ignore list, or call `apply_fsdp2_excluding_module_names` without patching verl).

**6. `OmniModelConfig.override_config` now splats every key into `AutoConfig.from_pretrained`**
[`workers/config/omni/model.py`](https://github.com/wtomin/verl-omni/blob/minicpm-dpo/verl_omni/workers/config/omni/model.py). In-tree Qwen recipes only set `attn_implementation`, so they should still load. Any extra key that is not a valid HF config/`from_pretrained` kwarg will now raise. Filter known MiniCPM fields onto the config object after load, or document this as `[BREAKING]` if you intend the new semantics.

**7. GPU evidence**
Commit `rm special e2e` dropped the smoke path. CPU tests cover packing/strip/freeze/transform. They do not cover FSDP leftover params, audio `get_omni_embedding` in-place fill under LoRA, or a training step. The PR body charts are not a substitute for a tiny `tests/special_e2e` script if MiniCPM is going to stay in-tree.

I did not re-run the listed pytest commands in this review.

verl_omni/utils/fsdp_utils.py
```diff
@@ -43,6 +45,78 @@
...
+        except TypeError:
+            fully_shard(model, **fsdp_kwargs)
+
+    if config.get("forward_prefetch", False):
+        fsdp_modules = [module for module in modules if isinstance(module, FSDPModule)]
+        for index, module in enumerate(fsdp_modules):
+            next_targets = fsdp_modules[index + 1 : index + 2]
+            if next_targets and hasattr(module, "set_modules_to_forward_prefetch"):
+                module.set_modules_to_forward_prefetch(next_targets)
+
```
zhtmike, line 81: I guess the PR is not READY, right?

wtomin: Well, I had to admit this PR looks ugly. There are quite a few patches because minicpm-o requires **remote code** in huggingface and the **transformers** version is `4.51.0`. I will be happy to hear some suggestions on how to remove these patches.

Because the old transformers version is not compatible with `transformers>=5.13.0,<=5.14.1`:

1. Transformers ≥ 5: call `post_init()` (`remote_code_compat.py`)
What: Before `from_pretrained`, wrap remote `MiniCPMO.__init__` so it ends with `post_init()`.

   Why: Transformers 5 expects every `PreTrainedModel` to call `post_init()` so `all_tied_weights_keys` is set. MiniCPM-o remote code only declares `_tied_weights_keys` and skips `post_init()`, so weight loading fails.

2. Whisper attention return value (`wrap_whisper_self_attn_forward`)
What: Pad apm Whisper `self_attn` output to (`hidden_states`, `attn_weights`, `past_key_values`), and rename `past_key_value` → `past_key_values`.

   Why: Remote `MiniCPMWhisperEncoderLayer` still unpacks three values. Newer version of `WhisperAttention` returns two. Training still runs the audio encoder so the unpack crashes.


The `fsdp_utils.py` modification here is because:

`FSDP2`: ignore apm by name (`apply_fsdp2_excluding_module_names`)
What: Do not `fully_shard` the `apm` subtree (including PEFT prefixes). Pass those params as `ignored_params` so they stay replicated.

Why: Whisper / audio-encoder shapes and leftover params do not shard cleanly under FSDP2 (`DTensor` / leftover errors). `apm` is frozen, so sharding it is unnecessary.

In the remote code of `MiniCPMWhisperEncoder.forward`([code link](https://huggingface.co/openbmb/MiniCPM-o-4_5/blob/main/modeling_minicpmo.py#L3989))
```python
def forward():
        ...
        embed_pos = self.embed_positions.weight
        ...
        hidden_states = inputs_embeds + embed_pos # adding the Tensor with DTensor causes error
```

Another patch is in `patch_minicpm_get_vllm_embedding`:

What: Clone each token-embedding row, scatter vision tokens out of place, then stack.

Why: Remote `get_vllm_embedding` does in-place `scatter_` on `vllm_embedding[i]`. PEFT `enable_input_require_grads` makes `embed_tokens` output a leaf; in-place writes on that view are illegal in backward.

In the remote code [code link](https://huggingface.co/openbmb/MiniCPM-o-4_5/blob/main/modeling_minicpmo.py#L518):
```python
    def get_vllm_embedding(self, data):
   ...

        cur_vllm_emb.scatter_(   # the monkey patch replaces scatter_ by scatter
                               0,
                               image_indices.view(-1, 1).repeat(1, cur_vllm_emb.shape[-1]),
                               cur_vs_hs.view(-1, cur_vs_hs.shape[-1]),
                     )
```

zhtmike: may move the patch to model specific, and loaded it on-fly.

examples/dpo_trainer/minicpm/README.md
wtomin: Sure. It's updated.

zhtmike:
I'm building MiniCPM-o 4.5 thinker GSPO (issue #406, AVQA variant) on
top of this PR's shared `minicpm` package. The AVQA path depends on the same
`apm` FSDP2 ignore, so I looked for a formal API instead of the patch —
proposal below before I file anything upstream.

**Where we are now:** I replaced the monkey-patch on my branch with an
explicit `OmniFSDPEngine._build_fsdp_module` override (the file's existing
idiom for verl engine methods) that copies verl@`fefb0802`'s fsdp2 path and
passes `ignored_params` on the root `fully_shard` call only, with a
maintenance note to re-diff at pin bumps and delete once verl ships a config
key. That also let me revert `model_base.py` / `fsdp_utils.py` to zero diff.
Happy to send the same override here if you prefer it over the swap — but the
better end state for both of us is the upstream fix below.

**The formal API exists in PyTorch:** `fully_shard(..., ignored_params=...)`.
The gap is verl-side:

- `verl/utils/fsdp_utils.py::apply_fsdp2` (pin `fefb0802`) calls
  `fully_shard(module, **fsdp_kwargs)` with the **same kwargs for every wrap
  target and the root** — no root-only channel;
- upstream verl has zero `ignored_params` support (code search), and it's a
  standing ask: verl#1346 (freeze vision tower), verl#596 (Qwen2.5-VL).

**Proposed fix (small upstream verl PR):** add a config key, resolved to a
param set, passed only on the root call:

```python
ignored_names = config.get("fsdp2_ignored_module_names", [])
ignored_params = (
    {p for n, p in model.named_parameters()
     if any(part in ignored_names for part in n.split("."))}
    or None
)
fully_shard(model, ignored_params=ignored_params, **fsdp_kwargs)  # root only
```

Then verl-omni recipes just set
`actor.fsdp_config.fsdp2_ignored_module_names=['apm']` and we both drop our
local copies (your patch, my override). Cost: one upstream PR + a verl pin
bump; it would also close verl#1346 / verl#596.

**Two bugs I hit while vendoring (caught by your own tests — patches in my
branch, take them regardless):**

1. `apply_fsdp2_excluding_module_names` injected `ignored_params` into *every*
   `fully_shard` call; `test_..._root_only` asserts root-only — fixed by
   gating on `module is model`.
2. The adapter test stub strips 2.6-era `audio_decoder`/`code2wav`, but
   `get_strip_modules()` returns `["tts"]` for 4.5 — aligned the stub.

Also: `OmniModelBase.build_module` is dead in the omni engine (only the
diffusion engine calls `build_module`; omni loads via the main-native
`auto_model_class`), so I dropped it when vendoring.

Sequencing: happy to file the upstream verl PR, or co-sponsor if you'd rather
drive it — whichever keeps one shared `minicpm` package and one upstream ask.

wtomin:
> **Where we are now:** I replaced the monkey-patch on my branch with an explicit `OmniFSDPEngine._build_fsdp_module` override (the file's existing idiom for verl engine methods) that copies verl@`fefb0802`'s fsdp2 path and passes `ignored_params` on the root `fully_shard` call only, with a maintenance note to re-diff at pin bumps and delete once verl ships a config key. That also let me revert `model_base.py` / `fsdp_utils.py` to zero diff. Happy to send the same override here if you prefer it over the swap — but the better end state for both of us is the upstream fix below.

Yes. I saw [Commit 9c30003](https://github.com/zhtmike/verl-omni/commit/9c300031d2c4a4117625d87b70e8c560780563bd#diff-d6e6b8b3d6234ec3ef317cf280d2831ab8f09af14dea28cf662acddf6c14d6c2) and I glad that it reverts `model_base.py` / `fsdp_utils.py`. I left a comment there. Overall, I agree to the edit in [Commit 9c30003](https://github.com/zhtmike/verl-omni/commit/9c300031d2c4a4117625d87b70e8c560780563bd#diff-d6e6b8b3d6234ec3ef317cf280d2831ab8f09af14dea28cf662acddf6c14d6c2), and I can sync this PR to it.


> **Proposed fix (small upstream verl PR):** add a config key, resolved to a param set, passed only on the root call:
> 
> ```python
> ignored_names = config.get("fsdp2_ignored_module_names", [])
> ignored_params = (
>     {p for n, p in model.named_parameters()
>      if any(part in ignored_names for part in n.split("."))}
>     or None
> )
> fully_shard(model, ignored_params=ignored_params, **fsdp_kwargs)  # root only
> ```

Nice idea! I suggest to also cover fsdp1 ignored paramsters, which is defined as the `ignored_modules` key name.

> **Two bugs I hit while vendoring (caught by your own tests — patches in my branch, take them regardless):**
> 
> 1. `apply_fsdp2_excluding_module_names` injected `ignored_params` into _every_
>    `fully_shard` call; `test_..._root_only` asserts root-only — fixed by
>    gating on `module is model`.
> 2. The adapter test stub strips 2.6-era `audio_decoder`/`code2wav`, but
>    `get_strip_modules()` returns `["tts"]` for 4.5 — aligned the stub.

I think these two are minor mistakes. Feel free to correct them. Thanks.

> Also: `OmniModelBase.build_module` is dead in the omni engine (only the diffusion engine calls `build_module`; omni loads via the main-native `auto_model_class`), so I dropped it when vendoring.

In the current PR, I actually define `auto_model_class` as `MiniCPMO` (`AutoModel` is called inside), and call `MiniCPMThinkerAdapter.build_module`. How is this aligned with your design?  @zhtmike

zhtmike:
> In the current PR, I actually define `auto_model_class` as `MiniCPMO` (`AutoModel` is called inside), and call `MiniCPMThinkerAdapter.build_module`. How is this aligned with your design?

Thanks for checking — we're already aligned. Your `4a72bfd` `_build_module` resolves `auto_model_class` and calls `MiniCPMO.from_pretrained`, which is exactly our mechanism (the omni engine never calls `build_module`; only the diffusion engine does). Dropping the unused omni hook keeps `model_base.py` at zero diff — happy either way, as long as we route one load path.

> I suggest to also cover fsdp1 ignored parameters, which is defined as the `ignored_modules` key name.

Good suggestion, agreed — one key (`fsdp_ignored_module_names`) resolving to modules for fsdp1 and to root-only `ignored_params` for fsdp2. 

> I think these two are minor mistakes. Feel free to correct them.

Both were already fixed in our vendor commit — nothing needed, thanks.

FYI in case they help your offline path: our branch also carries `convert_to_tensors` fixes (nulled tensor leaves; ragged `pixel_values`) and `<answer>`-tag demotion for `skip_special_tokens=True` decoders — feel free to cherry-pick.

verl_omni/models/transformers/minicpm_o.py
```diff
@@ -0,0 +1,282 @@
...
+        if llm_config is not None and hasattr(llm_config, "scale_emb"):
+            vllm_embedding = vllm_embedding * llm_config.scale_emb
+
+        vision_hidden_states = [
+            i.type(vllm_embedding.dtype) if isinstance(i, torch.Tensor) else i for i in vision_hidden_states
+        ]
+
+        rows = []
+        batch_size = len(data["input_ids"])
+        image_bound = data.get("image_bound") or [[] for _ in range(batch_size)]
```
SamitHuang, line 263: `bool(tensor)` is illegal once `image_bound` is a 2-D tensor. `_sample_bound` already materializes `(N, 2)` tensors, so this `data.get("image_bound") or [[]]` fallback raises `RuntimeError: Boolean value of Tensor with more than one value is ambiguous` as soon as collation stops using a Python list. Use an explicit `None` / empty-list check, not `or`, and add a CPU test that passes a 2-D tensor bound.

verl_omni/utils/fsdp_utils.py
```diff
@@ -43,6 +45,41 @@
...
+        return
+
+    ignored_params = {
+        param for name, param in model.named_parameters() if fsdp_name_is_ignored(name, ignored_module_names)
+    }
+    original_fully_shard = verl_fsdp_utils.fully_shard
+
+    def _fully_shard(module, *args, **kwargs):
+        if ignored_params:
+            kwargs = {**kwargs, "ignored_params": ignored_params}
```
SamitHuang, line 73: This injects the *root* `ignored_params` set into every `fully_shard` call. Nested wrap targets do not own those parameters, and torch rejects that. `test_apply_fsdp2_excluding_injects_ignored_params_on_root_only` asserts the child call gets `ignored_params=None`, so this helper and its CPU test currently contradict each other. Gate on `module is model` (root only).

verl_omni/pipelines/minicpm/thinker_training_adapter.py
```diff
@@ -0,0 +1,246 @@
...
+                module.set_input_embeddings = trainable_component.set_input_embeddings
+            if hasattr(trainable_component, "prepare_inputs_for_generation"):
+                module.prepare_inputs_for_generation = trainable_component.prepare_inputs_for_generation
+
+        module._no_split_modules = _MINICPM_NO_SPLIT_MODULES
+        return module
+
+    @classmethod
+    def get_fsdp_ignored_module_names(cls, model_config) -> list[str]:
+        return ["apm"]
```
SamitHuang, line 219: `patch_minicpm_get_vision_embedding` skips dummy `vpm` when `pixel_values` is empty, so audio-only DPO never touches `vpm`. This ignore list is still only `["apm"]`. Unused encoder params with `requires_grad=True` are the FSDP2 leftover-param footgun for a modality this recipe actually trains. Ignore both encoders (`["apm", "vpm"]`) or keep a dummy `vpm` path consistent with dummy `apm`.

verl_omni/workers/engine/fsdp/omni_impl.py
```diff
@@ -245,3 +245,26 @@ def _build_lora_module(self, module):
...
+            ignored = list(adapter_cls.get_fsdp_ignored_module_names(self.model_config))
+        if not ignored:
+            return super()._build_fsdp_module(module)
+
+        original_apply = transformer_impl.apply_fsdp2
+
+        def _apply(model, fsdp_kwargs, config):
+            apply_fsdp2_excluding_module_names(model, fsdp_kwargs, config, ignored)
+
+        transformer_impl.apply_fsdp2 = _apply
```
SamitHuang, line 266: Mutating `transformer_impl.apply_fsdp2` as a module global is still a race-prone hook even with `finally`. Concurrent engine init, or any other import of `apply_fsdp2`, sees the swap. Prefer an explicit `OmniFSDPEngine._build_fsdp_module` override that passes root-only `ignored_params` (as discussed with @zhtmike) instead of patching verl.

SamitHuang (changes requested):
## Review (REQUEST CHANGES)

Re-reviewed the current `wtomin:minicpm-dpo` tree after the force-push. The `image_bound` `or` crash moved from the adapter into `minicpm_o.py` and is still live. The FSDP2 helper still injects `ignored_params` on nested `fully_shard` calls, which contradicts `test_apply_fsdp2_excluding_injects_ignored_params_on_root_only`.

Inline comments cover:
1. `minicpm_o.py` tensor truthiness on `image_bound`
2. `fsdp_utils.py` nested `ignored_params` injection
3. FSDP ignore list is `apm` only while dummy `vpm` is skipped on audio-only batches
4. Global `transformer_impl.apply_fsdp2` monkeypatch

Also still blocking (not re-commented as a fifth line):
- PR title fails `check_pr_title.py`: missing `]` before `feat`. Required: `[omni, recipe, tests] feat: add MiniCPM-o offline DPO training`.
- `get_strip_modules()` returns `["tts"]` (4.5) but `test_configure_model_strips_generation_modules_and_keeps_outer_forward` still asserts 2.6-era `audio_decoder`/`code2wav` are gone.
- `OmniModelConfig.override_config` still splats every key into `AutoConfig.from_pretrained`; in-tree Qwen recipes only set `attn_implementation`, but extra non-HF keys will now raise. Filter MiniCPM fields after load, or mark `[BREAKING]`.

`OmniModelBase.build_module` is still unused on the omni engine path (load is `auto_model_class.from_pretrained`). CPU tests were not re-run in this review.

verl_omni/models/transformers/minicpm_o.py
wtomin: Fixed this issue in [81b78fe](https://github.com/verl-project/verl-omni/pull/550/commits/81b78fe7d90018e096922dece81c128fe7ae918f).

verl_omni/utils/fsdp_utils.py
wtomin: In  7e3012f, I added the guard `module is model` (root only).

SamitHuang:
Pls resolve the conflicts for further review, thx

Issue #247, opened 2026-07-09
[RFC] Layered L1-L4 CI/CD Plan for VeRL-Omni


wtomin:
### Feature request

[RFC] Layered L1-L4 CI/CD Plan for VeRL-Omni

### Motivation

VeRL-Omni needs a layered CI/CD system that catches API regressions before merge, validates GPU end-to-end training paths before PRs land, and continuously tracks numerical correctness, performance, and convergence quality for users and releases.

## Proposed CI/CD Layers

| Layer | Purpose | Trigger | Hardware | Blocking Scope | Output |
|---|---|---|---|---|---|
| L1 CPU API Tests | Validate CPU-only APIs, configs, data utilities, adapters, rewards, and unit behavior | Every PR, push to `main` / release branches | CPU only | Required before merge | Pass/fail, coverage report |
| L2 GPU Smoke Tests | Validate tiny-random GPU E2E training paths | Every PR before merge, usually via `ready-for-ci` | GPU | Required before merge | Smoke summary, logs, result JSON |
| L3 Nightly Regression | Detect numerical drift and performance regressions | Nightly, manual dispatch | Fixed GPU runners | Nightly regression gate | Metrics, baseline comparison, dashboard |
| L4 Convergence Tests | Validate real recipe convergence | Weekly, release candidate, manual dispatch | Production GPU cluster | Release readiness gate | Reward/loss curves, convergence report |

## Layer Details

| Layer | Scope | Pass Criteria | Non-Goals |
|---|---|---|---|
| L1 | CPU-only tests for APIs, configs, data pipeline, model adapters, reward managers, utility functions | Tests pass, coverage meets threshold, no import/config breakage | GPU runtime, model quality |
| L2 | Tiny-random models, dummy datasets, 1-2 training steps, GPU-only code paths | No exception, no OOM, no Ray failure, reaches configured steps | Accuracy, performance, convergence |
| L3 | 10-20 step fixed-seed runs across training/rollout backends | Loss/grad/log-prob drift within tolerance; performance within threshold | Long-term model quality |
| L4 | Real weights, real datasets, production-like recipes | Reward/loss curves match release baseline within tolerance | Fast PR feedback |

## Current State and Gaps

| Area | Current State | Gap |
|---|---|---|
| L1 CPU tests | `cpu_unit_tests.yml` runs `tests/**/*_on_cpu.py` | No true test coverage reporting or diff coverage |
[...]

PR #559, opened 2026-09-07
[diffusion] feat: add FSDP2 regional compilation

jiefangxuanyan:
### What does this PR do?

This PR adds optional regional `torch.compile` support for repeated Diffusers transformer blocks in the FSDP2 engine. The feature is disabled by default and forwards configurable options to Diffusers' `compile_repeated_blocks` API.

Qwen-Image training previously launched many small elementwise kernels inside its repeated transformer blocks. Regional compilation fuses many of these operations in both forward and backward execution, reducing kernel-launch and memory-traffic overhead without compiling the entire training engine.

The implementation compiles the repeated blocks after model structure and trainability changes, but before FSDP2 registers sharding hooks. Activation-checkpoint recomputation therefore uses the same compiled block path. FSDP1 and Ulysses sequence parallelism are rejected for now because these combinations have not received the required distributed validation.

FA3 variable-length metadata preparation remains an intentional eager boundary. This permits graph breaks around data-dependent third-party code while leaving the surrounding repeated block eligible for compilation. Dynamic shapes and a shared Dynamo cache avoid independently specializing every repeated block for changing prompt shapes.

Regional compilation directly affects actor evaluation in `old_log_prob` and actor forward/backward execution in `update_actor`. It also increases the long-lived Python object graph, which makes the existing full collections in `old_log_prob`, `update_actor`, and `update_weights` more expensive. The companion [diffusion GC PR](https://github.com/verl-project/verl-omni/pull/560) makes those existing collection points configurable; it is not required to use regional compilation, and this PR does not change their defaults.

AI assistance from OpenAI Codex was used to implement, test, profile, and draft this change. The human submitter reviewed the resulting changes and test results.

### Checklist Before Starting

- [x] Searched open PRs for [`compile`](https://github.com/verl-project/verl-omni/pulls?q=is%3Apr+is%3Aopen+compile). No open PR found by this search adds regional compilation for repeated Diffusers blocks.
- [x] Formatted the PR title according to the repository convention.

### Test

```bash
pytest \
  tests/models/test_flash_attention_3_graph_break_on_cpu.py \
  tests/workers/test_diffusion_regional_compile_on_cpu.py
```

- Regional compile and FA3 eager-boundary CPU tests on the standalone regional-compile branch with the repository-pinned `verl` revision: **5 passed**.
- Eight-GPU Qwen-Image FSDP2 training completed for two hours in both eager and compiled configurations without a training failure. These measurements were collected before the review branches were split, from an integration build containing the companion GC change with all GC settings left at their historical full-collection defaults. Extracting the standalone branch changed no functional compile code; the only compile-side difference is automatic formatting.
- Excluding the first warm-up/compilation step, each configuration contributed 57 steady-state steps:

| Metric | Eager | Regional compile | Change |
| --- | ---: | ---: | ---: |
| Old log probability | 9.74 s | 9.65 s | -0.09 s |
| Actor update | 29.64 s | 27.78 s | -1.86 s |
| Weight update | 7.02 s | 7.51 s | +0.49 s |
| End-to-end step | 110.34 s | 108.09 s | -2.25 s (-2.0%) |

The actor-update reduction is the main direct compute benefit. The small old-log-prob change is in the expected direction but is not used alone as evidence of a speedup. Weight update does not execute the compiled transformer; its regression comes from the higher cost of scanning the compiled Python object graph with the unchanged full-GC defaults. The [diffusion GC PR](https://github.com/verl-project/verl-omni/pull/560) reports the separately measured effect of tuning those collections.

Each configuration currently has one run on a different node, so these results demonstrate the Qwen-Image workload rather than a universal speedup across models or hardware. The first step is excluded because it includes compilation and cache population. Compilation therefore also affects first-step latency, but this PR does not claim a steady-state result for that initialization cost.

Nsight Systems confirms the intended qualitative change: the eager forward and backward contain many elementwise kernels, while regional compilation produces fused kernels.

These profiles were captured from the same pre-split integration build. The companion GC configuration does not change generated forward or backward kernels, so the profiles remain representative of this compile change.

Forward, before regional compilation:

Nsight Systems timeline showing many elementwise kernels in eager forward execution

Forward, after regional compilation:

Nsight Systems timeline showing fused kernels in compiled forward execution

Backward, before regional compilation:

Nsight Systems timeline showing many elementwise kernels in eager backward execution

Backward, after regional compilation:

Nsight Systems timeline showing fused kernels in compiled backward execution

### API and Usage Example

The feature is disabled by default:

```yaml
actor_rollout_ref:
  model:
    use_torch_compile: false
```

Enable the validated FSDP2 configuration with:

```yaml
actor_rollout_ref:
  actor:
    strategy: fsdp2
    fsdp_config:
      ulysses_sequence_parallel_size: 1
  model:
    use_torch_compile: true
    torch_compile_options:
      backend: inductor
      mode: default
      fullgraph: false
      dynamic: true
```

The Qwen-Image FSDP2 benchmark script enables this configuration by default. Append `actor_rollout_ref.model.use_torch_compile=False` to obtain the eager baseline.

### Design & Code Changes

- Add model configuration for enabling regional compilation and forwarding compiler options.
- Compile Diffusers repeated blocks at the FSDP2-compatible point in model construction.
- Keep FA3 variable-length metadata preparation outside compiled regions while allowing Dynamo to resume afterward.
- Reject unvalidated FSDP1 and Ulysses combinations with actionable errors.
- Add benchmark and Nsight controls for reproducible eager/compiled comparisons.
- Add CPU tests for configuration forwarding, supported and rejected strategies, compile placement, and the FA3 eager boundary.

- [x] Ran the full pre-commit suite on the standalone branch with the repository-pinned `verl` revision.
- [x] Added configuration comments and a benchmark usage path.
- [x] Added CPU tests for the new integration.


jiefangxuanyan:
Post-merge validation passed.

This was not a mechanical upstream merge: the previously bundled FA3 patch disappeared, so the workaround had to be adapted to the upgraded Diffusers implementation and revalidated instead of relying on the earlier results.

- Current branch tests: **11 CPU tests passed**, **2 Hub FA3 GPU contract tests passed**, and targeted pre-commit checks passed.
- The tests confirm that the same eager boundary avoids the Inductor cumsum failure and prevents an unbacked `max_seqlen_k` from reaching FA3.
- A three-step Linux eight-GPU Qwen-Image smoke completed with FA3 and regional compilation enabled. No compiler error, eager fallback, CUDA error, or training failure occurred.
- The graph-break locations and warm-up recompilation pattern were consistent with the previous long-run tests. All graph breaks and recompilations occurred before Step 1 completed, with none added during Steps 2 or 3.

The distributed smoke used merge commit `79d7f53`; the current tip `7ff78cb` only adds tests, CI coverage, and comments, without changing executable runtime behavior.

.github/workflows/l3_nightly.yml
```diff
@@ -42,6 +42,7 @@ on:
     paths:
       - "verl_omni/**"
       - "tests/nightly/**"
+      - "tests/workers/test_flash_attention_3_fake_backward.py"
```
wtomin: I think this test is out of the scope of nightly CI . Please remove it.

verl_omni/trainer/config/_generated_diffusion_trainer.yaml
```diff
@@ -309,6 +309,12 @@ actor_rollout_ref:
     config_path: null
     transformer_subfolder: transformer
     attn_backend: _flash_3_varlen_hub
+    use_torch_compile: false
```
wtomin: `actor_rollout_ref.actor.fsdp_config.use_torch_compile=False`: `fsdp_config` (`verl.workers.config.FSDPEngineConfig`) also has `use_torch_compile` key argument. 

I am a little concerned about introducing a similar key name as `actor_rollout_ref.model.use_torch_compile`, which is confusing to the users.

verl_omni/workers/engine/fsdp/diffusers_impl.py
```diff
@@ -107,6 +108,98 @@ def _fsdp_param_dtype(
     return configured_dtype
 
 
+def _keep_varlen_attention_metadata_eager() -> None:
```
wtomin: How about putting these new functions in `utils`, preventing `diffusers_impl.py‎` from being extensively long?

verl_omni/trainer/config/_generated_diffusion_trainer.yaml
wtomin: Besides, `verl.workers.config.ActorConfig` also has `use_torch_compile` key argument, so does `verl.workers.config.VeOmniEngineConfig`. Can we reuse exsiting key name for this feature?

zhtmike: > Besides, `verl.workers.config.ActorConfig` also has `use_torch_compile` key argument, so does `verl.workers.config.VeOmniEngineConfig`. Can we reuse exsiting key name for this feature?

verl's use_torch_compile is compiling the header for fsdp rollout for faster prediction (I guess it is legacy)

.github/workflows/l3_nightly.yml
jiefangxuanyan: I forgot about the [layered CI RFC](https://github.com/verl-project/verl-omni/issues/247) while developing this change. I reread it and think L2 fits this test better. The test is small, requires a GPU, and checks the affected functionality before merge. L3 is for nightly numerical and performance regression. I moved the test from L3 nightly to the L2 GPU smoke suite in d082501.

Would L2 be the right place for this test?

verl_omni/trainer/config/_generated_diffusion_trainer.yaml
jiefangxuanyan: The existing `use_torch_compile` options under the actor/FSDP engine compile the entropy helper. The new model-level option regionally compiles repeated Diffusers transformer blocks and keeps FA3 metadata preparation eager. The two options therefore control different parts of the model.

I kept a separate model-level option, but renamed it to `use_regional_compile`, together with `regional_compile_options`, in cc6fcd8.

Does this separation and naming look right to you?

verl_omni/workers/engine/fsdp/diffusers_impl.py
jiefangxuanyan: I moved the regional compile and FA3 eager-boundary helpers from `diffusers_impl.py` to `verl_omni/utils/diffusion_compile.py` in 3ff1a8d. The model implementation now handles model construction, while the compile integration logic is in the existing `verl_omni.utils` package.

Is this the location and separation you expected?

zhtmike:
let use make #560 settled first.

examples/flowgrpo_trainer/qwen_image/run_qwen_image_ocr_fsdp2_benchmark.sh
```diff
@@ -3,23 +3,42 @@
...
 NUM_NODES=${NUM_NODES:-1}
 REWARD_TP=1
+export TORCH_LOGS="${TORCH_LOGS:-graph_breaks,recompiles}"
+echo "Using TORCH_LOGS=$TORCH_LOGS for torch.compile diagnostics."
 
 NUM_GPUS=$NUM_GPUS NUM_NODES=$NUM_NODES bash "$SCRIPT_DIR/run_qwen_image_ocr.sh" \
     actor_rollout_ref.actor.strategy=fsdp2 \
+    actor_rollout_ref.actor.fsdp_config.forward_prefetch=True \
+    actor_rollout_ref.model.use_regional_compile=True \
+    'actor_rollout_ref.model.regional_compile_options={backend:inductor,mode:default,fullgraph:false,dynamic:true}' \
```
zhtmike, line 33: any reason why can't we use full graph compiling?

examples/flowgrpo_trainer/qwen_image/run_qwen_image_ocr_fsdp2_benchmark.sh
```diff
@@ -3,23 +3,42 @@
...
     actor_rollout_ref.rollout.step_execution=True \
     reward.num_workers=$((NUM_GPUS / REWARD_TP)) \
     reward.reward_model.rollout.tensor_model_parallel_size=$REWARD_TP \
     reward.reward_model.rollout.enforce_eager=False \
     reward.reward_model.rollout.max_num_seqs=128 \
     reward.reward_model.rollout.max_model_len=8192 \
     trainer.logger='["console", "tensorboard"]' \
-    trainer.experiment_name=qwen_image_ocr_8x80g_fsdp2_benchmark \
+    trainer.experiment_name=qwen_image_ocr_8x80g_fsdp2_fa3_compile_graph_breaks_benchmark \
```
zhtmike: no need to change name.

tests/gpu_smoke/run_gpu_smoke_core.sh
```diff
@@ -43,4 +43,8 @@ run_test 7 "diffusers ulysses sp" \
     env CUDA_VISIBLE_DEVICES="${CUDA_DEVICE_LIST}" \
     torchrun --nproc_per_node="${NUM_GPUS}" --local-ranks-filter=0 tests/workers/test_diffusers_ulysses.py
 
+run_test 8 "FA3 fake backward contract" \
```
zhtmike: no need this, pls revert

tests/workers/test_flash_attention_3_fake_backward.py
zhtmike, line 1: no need this., pls remove

verl_omni/trainer/config/diffusion/model/diffusion_model.yaml
```diff
@@ -45,6 +45,27 @@ transformer_subfolder: transformer
...
+
+  # Compiler mode; default avoids CUDA graph capture during the first rollout.
+  mode: default
+
+  # Allow intentional eager boundaries around unsupported third-party code.
+  fullgraph: False
+
+  # Text masks, resolutions, and frame counts commonly produce input-dependent shapes.
+  # Workloads that guarantee fixed shapes can override this for stronger specialization.
+  dynamic: True
```
zhtmike, line 67: if the configs are not tunable or not validated, I think we can keep them hard coded in scripts.

verl_omni/utils/diffusion_compile.py
```diff
@@ -0,0 +1,116 @@
...
+    resumes tracing, so token packing, the FA3 custom op, and the rest of the
+    repeated transformer block remain eligible for regional compilation.
+    The same eager boundary protects against both failures and must remain
+    while either one is reproducible.
+
+    The patched names are private Diffusers APIs. Remove this workaround once
+    upstream fixes both symbolic cumsum lowering and FA3's handling or
+    contract for data-dependent maximum sequence lengths, then revalidate the
+    real masked and unmasked FA3 paths without this boundary, including with
+    ``fullgraph=True``. Otherwise revisit it whenever those helpers change.
```
zhtmike: make your docstring short and model-agnostic.

verl_omni/utils/diffusion_compile.py
```diff
@@ -0,0 +1,116 @@
...
+) -> None:
+    """Regionally compile repeated diffusion blocks before FSDP2 mutates them."""
+    if not model_config.use_regional_compile:
+        return
+    if engine_config.strategy == "fsdp":
+        # Regional compilation has not been validated with this engine's FSDP1
+        # path, so reject the combination as a whole. ``use_orig_params=True``
+        # is a necessary prerequisite for eventually enabling FSDP1 with
+        # torch.compile, but satisfying it alone would not make the currently
+        # untested integration supported.
```
zhtmike: trim comment, it is already shown in the below error.

verl_omni/utils/diffusion_compile.py
```diff
@@ -0,0 +1,116 @@
...
+
+    # Regional compilation has not been validated with Ulysses SP, so reject
+    # the combination as a whole. Ulysses places SP collectives inside each
+    # compiled block. Enabling it requires distributed validation that Dynamo,
+    # Inductor, and AOTAutograd preserve the SP all-to-all participation and
+    # ordering in the original forward, activation-checkpoint recomputation,
+    # and backward while FSDP2 independently schedules parameter all-gathers
+    # and gradient reduce-scatters. Rank-specific graph breaks, guard misses,
+    # or recompilation could otherwise make ranks enter SP and DP collectives
+    # in different orders and hang the job.
```
zhtmike: same, trim it.

tests/workers/test_flash_attention_3_fake_backward.py
```diff
@@ -0,0 +1,99 @@
...
+import pytest
+import torch
+from diffusers.models import attention_dispatch
+from torch._subclasses.fake_tensor import FakeTensorMode
+from torch.fx.experimental.symbolic_shapes import GuardOnDataDependentSymNode, ShapeEnv
+
+
+def _run_fa3_backward_fake(*, symbolic_max_seqlen_k: bool):
+    assert torch.cuda.is_available(), "The FA3 contract test requires a CUDA runner."
+    backend = attention_dispatch.AttentionBackendName._FLASH_3_VARLEN_HUB
```
wtomin: Unfortunately, current L2 smoke test ci env cannot support fa3 (kernel download may fail). I think `tests/workers/test_flash_attention_3_fake_backward.py` will leads to an error in L2 test.

I suggest to move the major guard to cpu, in `test_flash_attention_3_graph_break_on_cpu.py`

wtomin:
I suggest to update `docs/examples/config.md` for `use_regional_compile`.

I also suggest to update the recipe `examples/flowgrpo_trainer/qwen_image/README.md` to briefly introduce this feature.

tests/workers/test_flash_attention_3_fake_backward.py
jiefangxuanyan: Thanks. The test works in my CUDA-enabled WSL environment with network access, but a CPU-only PyTorch environment has an additional limitation. It can download the Hub metadata, but `kernels` rejects every available FA3 artifact because the repository has no CPU build variant. Forcing a CUDA variant is also rejected because that environment supports only the CPU backend. Therefore, the current CI infrastructure cannot run the real FA3 contract test.

I removed the real-FA3 test and its GPU-smoke entry in b431a6f.

This leaves one narrow maintenance gap, which is also worth flagging to @zhtmike: if the Inductor issue is fixed first and that rationale and canary are removed, then FA3 later starts accepting the unbacked value, the remaining automated tests will not report that the last reason for the eager boundary has disappeared. This will require manual revalidation of real FA3 when relevant dependencies change.

tests/gpu_smoke/run_gpu_smoke_core.sh
jiefangxuanyan: Removed in b431a6f. The CI limitation and remaining maintenance gap are discussed in https://github.com/verl-project/verl-omni/pull/559#discussion_r4035274572.

tests/workers/test_flash_attention_3_fake_backward.py
jiefangxuanyan: Removed in b431a6f. The remaining maintenance gap is discussed in https://github.com/verl-project/verl-omni/pull/559#discussion_r4035274572.

examples/flowgrpo_trainer/qwen_image/run_qwen_image_ocr_fsdp2_benchmark.sh
jiefangxuanyan: For the Qwen-Image path I validated, the known full-graph blockers are specific to FA3. The native/SDPA path can compile with `fullgraph=True`, but FA3 is the main high-performance path that this change needs to support. Its current Diffusers path has two independent blockers: Inductor cannot lower the symbolic cumulative-length calculation, and FA3 fake backward cannot consume the resulting unbacked maximum-sequence-length symbol. I therefore keep metadata preparation eager and use `fullgraph=False` so compilation can resume around that boundary. We can enable full-graph compilation for FA3 after both issues are fixed and the unpatched masked and unmasked paths are revalidated.

examples/flowgrpo_trainer/qwen_image/run_qwen_image_ocr_fsdp2_benchmark.sh
jiefangxuanyan: Reverted in 5f0b661.

verl_omni/trainer/config/diffusion/model/diffusion_model.yaml
jiefangxuanyan: The scope here is `regional_compile_options` and its four current entries. The dictionary is passed to `compile_repeated_blocks` and then to `torch.compile`, so it can currently contain arguments beyond these four entries. The four entries are not universally fixed across the repository:

- `fullgraph=False` is required by the current FA3 workaround, while native/SDPA can use full-graph compilation.
- `dynamic=True` is required by the validated Qwen-Image workload because prompt-dependent text shapes otherwise cause repeated specialization. The repository also has a small number of plausibly fixed-shape workloads, such as SD3.5 recipes with fixed image resolution and prompt embeddings padded to fixed CLIP/T5 lengths, although regional compilation has not been validated for them.
- `backend=inductor` and `mode=default` are the validated performance settings, but both are real `torch.compile` choices rather than structural invariants.

Would you prefer me to hard-code all four settings to expose only the validated Qwen-Image/FA3 behavior, or hard-code only the settings that are constraints of that path? If the dictionary remains, should users be allowed to pass other `torch.compile` keyword arguments through it?

verl_omni/utils/diffusion_compile.py
jiefangxuanyan: Addressed in 5f0b661.

verl_omni/utils/diffusion_compile.py
jiefangxuanyan: Addressed in 5f0b661.

verl_omni/utils/diffusion_compile.py
jiefangxuanyan: Addressed in 5f0b661.

jiefangxuanyan:
@wtomin, thanks. I will update `docs/examples/config.md` and `examples/flowgrpo_trainer/qwen_image/README.md` after the discussion with @zhtmike about whether `regional_compile_options` remains configurable or is hard-coded settles the interface.

examples/flowgrpo_trainer/qwen_image/run_qwen_image_ocr_fsdp2_benchmark.sh
zhtmike: interesting, it worth a PR in `kernels`.

verl_omni/trainer/config/diffusion/model/diffusion_model.yaml
zhtmike: i see, let us keep current design.

zhtmike (approved):

jiefangxuanyan:
@wtomin, thanks. After confirming the configuration design with @zhtmike, I updated both documents in af41321.

jiefangxuanyan:
@zhtmike, the GPU smoke failure was caused by a transient port collision (`EADDRINUSE` on port 18836) while vLLM-Omni was starting the DiffusionNFT rollout workers. That test ran with `use_regional_compile=False`, and the other nine smoke tests passed, so the failure is unrelated to this PR. Could you please trigger the CI again?

examples/flowgrpo_trainer/qwen_image/run_qwen_image_ocr_fsdp2_benchmark.sh
jiefangxuanyan: Thanks, this could be a useful direction to explore in the future. However, I think addressing these issues in this PR would expand its scope too much, as the fixes involve changes across upstream projects. I would keep the eager metadata boundary here and consider the following options separately:

1. **PyTorch:** The symbolic `full(...).cumsum()` failure is addressed by [pytorch/pytorch#193959](https://github.com/pytorch/pytorch/pull/193959). A minimal CPU reproducer fails on 2.13.0 and passes on `2.15.0.dev20260919+cpu`. The fix is not in the current 2.14.0 release, so we can pick it up once it reaches a compatible stable dependency stack.

2. **Diffusers:** Independently, simplify metadata preparation for uniform sequences: construct cumulative lengths with `arange * seq_len` and use the known length instead of `max().item()`. This applies to both Q and K/V without a mask, but only Q in the general masked path. Mask-derived K/V lengths still need their cumulative sum and exact maximum, so this optimization would not resolve the second blocker.

3. **FlashAttention, then the Hub kernel:** Investigate simplifying the internal backward custom-op interface so it does not expose the temporary `softmax_d` tensor. The FA3 autograd backward and Diffusers backward callers inspected so far ignore that return value and use the populated `dq/dk/dv` tensors instead. Keeping the temporary output inside the custom op should let the fake implementation avoid reproducing its length-dependent allocation logic, while preserving runtime kernel selection using exact lengths. This is a proposed fix, not yet validated; it needs caller-compatibility checks and compiled forward/backward tests, including finite windows. The natural upstream is `Dao-AILab/flash-attention`, followed by a sync and Hub release from `huggingface/kernels-community`, whose [FA3 sync instructions](https://github.com/huggingface/kernels-community/blob/main/AGENTS.md#flash-attn3) explicitly copy the upstream Python interface.

For this fix, I would prefer not to assume that padded shapes equal exact maximum valid lengths, or require callers to pass additional CPU length metadata through the model and attention APIs. The shape assumption does not hold with arbitrary padding, and the metadata approach introduces cross-layer API and consistency requirements. We can keep computing exact K/V lengths from the mask. Capturing that scalar extraction would still involve GPU-to-CPU synchronization at runtime; the FA3 change targets compilation compatibility, not removal of that synchronization.

Any future removal of the eager boundary would require validation of the real masked and unmasked paths, packing, forward/backward, and changing lengths with `fullgraph=True`. Full-graph FA3 support remains outside the scope of this PR.

zhtmike:
All CI green.

PR #560, opened 2026-09-07
[diffusion, perf] fix: remove unnecessary full GC

jiefangxuanyan:
### What does this PR do?

This PR removes the three explicit full Python GC points that remain in the verl-omni diffusion actor path:

- loading the actor onto the device for evaluation;
- loading the actor and optimizer for training;
- offloading the actor after training.

The evaluation and training cases share one physical `gc.collect()` call. For actor offload, this PR follows the upstream cleanup sequence: synchronize the device, clear the allocator cache, and log GPU memory before resuming the KV cache, without scanning the Python heap.

[verl #7873](https://github.com/verl-project/verl/pull/7873) independently removed the corresponding full-GC sites in verl. It also fixed the CUDA IPC release-ordering issue at weight-transfer cleanup before removing that collection. After discussing the follow-up privately with @zhtmike, we agreed that verl-omni should follow the same direct-removal approach for its remaining three sites instead of retaining public GC controls or a compile-dependent policy. The weight-transfer cleanup change itself will reach verl-omni through a future verl pin update; this PR does not duplicate its protocol fix.

This change has no configuration API and does not depend on regional compilation. Regional compilation made the cost easier to observe because its larger long-lived Python object graph made each full-heap scan more expensive, but the measured collections did not provide a repeatable memory benefit with or without compilation.

AI assistance from OpenAI Codex was used to implement, test, benchmark, and draft this change. The human submitter reviewed the resulting changes and test results.

### Checklist Before Starting

- [x] Searched open PRs separately for [`GC`](https://github.com/verl-project/verl-omni/pulls?q=is%3Apr+is%3Aopen+GC) and [`garbage collection`](https://github.com/verl-project/verl-omni/pulls?q=is%3Apr+is%3Aopen+%22garbage+collection%22).
- [x] Reviewed the related direct-removal work in [verl #7873](https://github.com/verl-project/verl/pull/7873).
- [x] Formatted the PR title according to the repository convention.

### Test

```bash
pytest \
  tests/workers/test_diffusers_timestep_staging_on_cpu.py \
  tests/workers/test_omni_lora_weight_sync_on_cpu.py
```

- Targeted CPU tests: **52 passed**.
- Ruff check and format passed for the changed production and test files.
- Eight-GPU Qwen-Image FSDP2 training completed for two hours with both historical full GC and an experimental policy that disabled these three collections. Excluding warm-up, the latter completed 58 steady-state steps without CUDA allocated-memory accumulation or greater per-rank RSS growth than the full-GC run.

The [records extracted from the GC diagnostic logs](https://github.com/user-attachments/files/32083887/gc_diagnostics_final_runs_public.csv) provide the direct evidence for the removal:

- Under regional compilation, each device-load point had 456 steady-state full-GC records and actor offload had 450 captured records. None reduced CUDA allocated or reserved memory, and none showed a repeatable RSS decrease.
- Mean full-GC time was approximately 0.915 seconds per call with regional compilation and 0.57 seconds without it. The time was almost entirely Python thread CPU time.
- With the three collections disabled, per-rank RSS growth over the steady-state run averaged 745 MiB, compared with 948 MiB under compiled full GC; CUDA allocated memory did not accumulate in either run. Each configuration has one run on a different node, so these values support the absence of a regression in this workload rather than a universal memory comparison.

The experimental policy also changed the separate weight-transfer cleanup from full GC to generation 1. The phase timings are retained here for context, but the weight-update result therefore must not be attributed only to the actor-offload change in this PR:

| Metric | Full GC | Experimental policy | Change |
| --- | ---: | ---: | ---: |
| Old log probability | 9.65 s | 8.95 s | -0.70 s |
| Actor update | 27.78 s | 27.11 s | -0.67 s |
| Weight update | 7.51 s | 6.13 s | -1.38 s |
| End-to-end step | 108.09 s | 106.06 s | -2.03 s (-1.9%) |

The first two phase changes correspond directly to the evaluation and training device-load collections removed here. Actor offload and weight-transfer cleanup partially overlap in the LoRA fast path, so their individual GC times cannot be added or separated from the weight-update result using this run alone. Generation, reward, and output-dump timings also varied across the independent runs and are not attributed to this PR.

Python sampling places full GC directly in an observed training stall. The profile was captured from the same integration build with regional compilation enabled:

Nsight Systems timeline with Python sampling identifying garbage collection during a training stall

### API and Usage Example

No API, CLI, or configuration change.

### Design & Code Changes

- Remove the shared full collection after loading the diffusion FSDP actor or optimizer onto the device.
- Replace actor-offload `aggressive_empty_cache(force_sync=True)` with device synchronization and allocator-cache cleanup, matching verl #7873 while preserving the post-offload GPU-memory log.
- Remove the superseded GC configuration, diagnostics forwarding, generated configuration fields, and their dedicated tests.
- Keep the LoRA weight-sync regression test focused on weight-version stamping by executing its mocked background work inline.

- [x] Ran the targeted CPU tests and lint checks listed above.
- [ ] Run the full pre-commit suite on the final branch.
- [x] No documentation or usage update is required because there is no public API.

<details>
<summary>Revision history: superseded configuration and diagnostics designs</summary>

The first version proposed four independently configurable GC settings plus one global diagnostics switch. A later revision removed the per-site public settings and derived the four GC decisions from regional compilation, while retaining diagnostics forwarding. Those designs preserved historical full GC outside the validated Qwen-Image compile workload and depended on verl #7780 and verl-omni #559.

The diagnostic implementation and its dedicated smoke tests were useful during investigation: they produced the public records and long-run evidence summarized above. They are no longer part of the production change because the resulting decision is to remove the three verl-omni collections directly. The configuration fields, forwarding paths, generated YAML changes, and policy-selection tests have therefore been removed rather than kept as unused framework surface.

</details>


verl_omni/workers/engine/fsdp/diffusers_impl.py
```diff
@@ -698,11 +700,24 @@ def to(self, device: str, model: bool = True, optimizer: bool = True, grad: bool
 
         assert device in (device_name, "cpu")
         if device == device_name:
+            if self.mode not in ("train", "eval"):
```
SamitHuang: **Blocking:** this rejects any accelerator load outside `train`/`eval` context, but `DiffusionDetachActorWorker._actor_model_for_snapshot()` still does `engine.to(get_device_name(), ...)` with `mode is None` (`verl_omni/workers/detach_actor_worker.py`). Param-offload snapshot/restore will now raise here.

`TrainingWorker.to("device")` has the same gap — it is a public manual load API and does not enter a train/eval context.

If the GC choice needs a mode, default it (e.g. use the train setting, or `getattr` a fallback) when `mode` is unset instead of failing closed on a previously valid path. The new CPU-only test does not cover this caller.

SamitHuang: **Blocking:** `extra_context={"gc_diagnostics": ...}` is passed for every diffusion actor, including `strategy=veomni`. `EngineRegistry.new(..., **extra_context)` forwards that kwarg into `VeOmniDiffusionEngine.__init__`, which only accepts the four existing constructor args — VeOmni init will TypeError.

Gate this on FSDP (`fsdp`/`fsdp2`), add `gc_diagnostics=False` to `VeOmniDiffusionEngine`, or accept `**kwargs`. Ref/teacher `TrainingWorkerConfig` construction still omits `extra_context`, so diagnostics never reach those engines even on FSDP.

SamitHuang: **Blocking:** `_offload_actor_and_empty_cache` is shared with omni, but `gc_on_actor_offload` lives only on `DiffusionRolloutConfig`. Omni generated YAML in this PR adds `gc_diagnostics` / `gc_on_weight_transfer_cleanup` and does **not** add `gc_on_actor_offload`, so omni `update_weights` will fail when resolving this attribute.

Use `getattr(self.config.rollout, "gc_on_actor_offload", True)` (historical default) or put the field on the shared rollout/checkpoint-engine config from the companion verl change, not only on the diffusion dataclass.

SamitHuang (changes requested):
**Request changes.** The per-point GC API, Hydra defaults, and CPU tests look directionally right, and keeping historical `true` defaults is the correct compatibility choice. A few call-site issues need to be fixed before this can merge.

### Blocking (inline)
1. FSDP `to(device)` now requires train/eval mode, which breaks `DiffusionDetachActorWorker` snapshot restore and `TrainingWorker.to("device")`.
2. `extra_context={"gc_diagnostics": ...}` is forwarded for all diffusion backends; `VeOmniDiffusionEngine` does not accept that kwarg.
3. Shared `ActorRolloutRefWorker._offload_actor_and_empty_cache` reads `rollout.gc_on_actor_offload`, which only exists on the diffusion dataclass — omni `update_weights` will miss the attribute.

### Also required
This PR imports `collect_garbage` / `GCSetting` / `validate_gc_setting` and new `aggressive_empty_cache` / `BucketedWeightSender` kwargs from the companion verl GC change ([verl#7780](https://github.com/verl-project/verl/pull/7780)), but `.github/verl_pin.txt` is unchanged. CI on current pin cannot import those symbols. Please bump the pin (or land a pin PR first) and do not merge this independently.

### Nits
- Per `.agents/rules/config.md`, add `#` field comments on `DiffusionFSDPEngineConfig` and `gc_on_actor_offload`.
- CPU tests for the four GC points look good; I did not rerun them. GPU evidence in the PR body is Qwen-Image-specific as noted.

AI assistance was used to inspect the diff; the findings above were verified against the PR branch and current `main` call sites.

verl_omni/workers/engine/fsdp/diffusers_impl.py
jiefangxuanyan: Fixed in 80c4772. `mode=None` remains valid for manual accelerator loads. These loads preserve the historical default full GC without being attributed to either the train or eval diagnostics point.

jiefangxuanyan: Fixed in f14a98e. I restored the original four-argument engine construction contract. The global switch is copied only into actor engine configs that declare GC diagnostics support, so VeOmni receives no new argument. Ref and teacher remain intentionally unwired because those paths have not been validated.

jiefangxuanyan: Fixed in 80c4772. The shared offload path now checks whether `gc_on_actor_offload` is present. Configs exposing the field pass it explicitly; other configs retain the historical default full GC. This remains independent of model type.

jiefangxuanyan:
@SamitHuang Added documentation for the supported `gc_on_actor_offload` values in a33a5f7: `True` runs a full collection, `False` disables collection, and an integer selects a GC generation.

verl_omni/workers/config/diffusion/actor.py
```diff
@@ -19,13 +19,15 @@
 from verl.base_config import BaseConfig
 from verl.trainer.config import CheckpointConfig
 from verl.trainer.config.algorithm import RolloutCorrectionConfig
+from verl.utils.memory_utils import GCSetting, validate_gc_setting
```
wtomin, line 22: I didn't find `GCSetting` or `validate_gc_setting` in today's main branch of https://github.com/verl-project/verl/blob/main/verl/utils/memory_utils.py ( `a9f2985` commit id)

Please update `.github/verl_pin.txt` to the correct commit id, otherwise the CI may fail.

zhtmike:
I think adding a config for gc is very strange ... how about say if we enable compile -> then disable the gc

zhtmike:
gc is not supposedly to be configurable.

jiefangxuanyan:
@zhtmike, following up on your question in the community meeting, I traced how these four points entered the codebase:

| Current diagnostic point | Historical origin | How it was introduced | Documented rationale |
| --- | --- | --- | --- |
| `eval_device_load` / `train_device_load` | VeRL [PR #1977](https://github.com/verl-project/verl/pull/1977), diffusion [PR #5802](https://github.com/verl-project/verl/pull/5802), and migration [PR #6200](https://github.com/verl-project/verl/pull/6200) | These two labels came from one historical `gc.collect()` in the FSDP engine's `to(device)` method. It was present when the generic engine was created, mirrored by the diffusion engine, and then migrated to VeRL-Omni. | No point-specific explanation or measurement found. |
| `actor_offload` | VeRL [PR #3285](https://github.com/verl-project/verl/pull/3285) and [PR #4211](https://github.com/verl-project/verl/pull/4211), then VeRL-Omni [PR #156](https://github.com/verl-project/verl-omni/pull/156) | Added during rollout/trainer-mode construction, carried through the model-engine refactor and diffusion migration, and later moved into an offload helper for the LoRA update path. | No evidence specific to this call site was found. The general `aggressive_empty_cache()` helper came from [PR #2678](https://github.com/verl-project/verl/pull/2678), which reported up to 30 GB of cached VRAM and hybrid-engine OOMs in a Megatron workload; later placement comments only note that KV-cache resume needs the actor memory freed. |
| `weight_transfer_cleanup` | VeRL [PR #4280](https://github.com/verl-project/verl/pull/4280), [PR #5029](https://github.com/verl-project/verl/pull/5029), and [PR #5309](https://github.com/verl-project/verl/pull/5309) | Added with the CUDA-IPC weight-transfer path. A later refactor temporarily added a second GC in the sender cleanup sequence; helper extraction ultimately retained one in `BucketedWeightSender._cleanup()`. | No point-specific explanation or measurement found. Its placement nevertheless has the clearest lifecycle semantics: transfer references are removed before GC, followed by CUDA IPC collection and cache emptying. |

All three historical call-site decisions, corresponding to the four current diagnostic points, were added incidentally as part of larger engine, mode-switching, or IPC changes. I found no point-specific experimental evidence that Python cyclic GC was necessary at any of these boundaries, although the history also does not prove that the calls are unnecessary.

---

This history also illustrates the difficulty with disabling the calls unconditionally when compile is enabled. In hindsight, without point-specific evidence, the better original choice may have been not to add them, or to add them only after an experiment demonstrated a need; a call could also be introduced later if a workload exposed a concrete problem. Once such calls become existing behavior, however, they become part of the baseline against which many pipelines have been validated. We cannot feasibly disprove their necessity across every RL workload, so how to govern the existing calls needs further discussion. [The companion comment in verl PR #7780](https://github.com/verl-project/verl/pull/7780#issuecomment-5630153644) provides the related governance and measurement context.

The current experiment establishes that disabling or reducing these calls is safe for this Qwen-Image workload, not for every workload using compile. Tying their removal to compile globally would turn that workload-specific result into a framework-wide policy and could regress an untested pipeline. The configuration therefore preserves the existing defaults while allowing workloads with their own measurements to tune the four points covered by this PR.

verl_omni/workers/config/diffusion/actor.py
jiefangxuanyan: You are right that the current pin does not contain these symbols. As stated in the PR description:

> This PR depends on the configurable collection utility and weight-transfer cleanup setting from the [verl GC PR](https://github.com/verl-project/verl/pull/7780).

That PR has not merged yet, so there is not yet a valid commit ID in `verl-project/verl` that I can reference from `.github/verl_pin.txt`. I will update the pin after it merges; this PR should not be merged independently before then.

zhtmike:
I still prefer not to add the option, the gc is hard to tune and too implicit to the users. I better to say just let the compile to determine whether the gc is enable or not. And keep the default compile to be off.

jiefangxuanyan:
@zhtmike, thanks for clarifying. Our hesitation about coupling GC behavior to regional compile is that a GC policy validated on one workload may not carry over safely to every training pipeline. That limitation may be acceptable here, since regional compile is opt-in and currently has Qwen-Image as its reference setup. We could treat compile and its GC behavior as a single validated configuration for Qwen-Image, while making it clear that other pipelines need to validate the combination before enabling compile. Does that match the scope you have in mind?

One detail remains unclear from “just let the compile to determine **whether the gc is enable or not**.” In the validated Qwen-Image run, the first three GC points were disabled, while weight-transfer cleanup used a generation-1 collection rather than disabling GC entirely. Should compile retain this generation-1 cleanup, or did you mean that all four explicit collections should be disabled whenever compile is enabled?

zhtmike:
> @zhtmike, thanks for clarifying. Our hesitation about coupling GC behavior to regional compile is that a GC policy validated on one workload may not carry over safely to every training pipeline. That limitation may be acceptable here, since regional compile is opt-in and currently has Qwen-Image as its reference setup. We could treat compile and its GC behavior as a single validated configuration for Qwen-Image, while making it clear that other pipelines need to validate the combination before enabling compile. Does that match the scope you have in mind?

Agree.
 
> One detail remains unclear from “just let the compile to determine **whether the gc is enable or not**.” In the validated Qwen-Image run, the first three GC points were disabled, while weight-transfer cleanup used a generation-1 collection rather than disabling GC entirely. Should compile retain this generation-1 cleanup, or did you mean that all four explicit collections should be disabled whenever compile is enabled?

generation-1 cleanup.

jiefangxuanyan:
@zhtmike We have updated the PR based on our discussion. The GC behavior is now derived from `use_regional_compile` instead of separate per-site configuration.

This PR copies the regional compile configuration fields from #559 so that the GC policy can be connected and tested here. The two PRs are expected to merge automatically. However, the fields copied into this PR do not implement regional compile by themselves: that functionality still depends on #559 being merged.

tests/gpu_smoke/run_gpu_smoke_diffusion_e2e.sh
```diff
@@ -71,4 +71,8 @@ run_test 9 "Diffusion OPD v1 separate_async standalone teachers e2e" \
     env CUDA_VISIBLE_DEVICES="${CUDA_DEVICE_LIST}" NUM_GPUS="${NUM_GPUS}" SMOKE=async \
     bash tests/special_e2e/run_diffusion_teacher_smoke.sh
 
+run_test 10 "GC diagnostics" \
+    env CUDA_VISIBLE_DEVICES="${CUDA_DEVICE_LIST}" \
+    bash tests/special_e2e/run_gc_diagnostics_qwen_image.sh "${diffusion_trainer_args[@]}"
+
```
zhtmike: no need this

tests/gpu_smoke/select_gpu_smoke_groups.py
```diff
@@ -91,6 +91,9 @@ class SmokeGroup:
         "verl_omni/pipelines/**",
         "verl_omni/trainer/config/diffusion/**",
         "verl_omni/trainer/diffusion/**",
+        "verl_omni/workers/config/diffusion/**",
+        "verl_omni/workers/engine/fsdp/diffusers_impl.py",
+        "verl_omni/workers/engine_workers.py",
```
zhtmike: and revert this

tests/special_e2e/run_gc_diagnostics_qwen_image.sh
zhtmike, line 1: no need this

verl_omni/workers/engine/fsdp/diffusers_impl.py
```diff
@@ -51,10 +51,10 @@
     offload_fsdp_model_to_cpu,
     offload_fsdp_optimizer,
 )
-from verl.utils.memory_utils import aggressive_empty_cache
+from verl.utils.memory_utils import aggressive_empty_cache, collect_garbage
```
zhtmike, line 54: there is no `collect_garbage` in verl.

jiefangxuanyan: @zhtmike Yes, `collect_garbage` is not present in the currently pinned VeRL revision because its prerequisite PR has not been merged yet. As stated in this PR description:

> This PR depends on the configurable collection utility and weight-transfer cleanup setting from the [verl GC PR](https://github.com/verl-project/verl/pull/7780).

This is the same dependency that @wtomin pointed out earlier. I can update the VeRL pin to a valid upstream commit containing `collect_garbage` after verl-project/verl#7780 is merged; until then, this PR needs to be reviewed and tested together with that prerequisite change.

tests/gpu_smoke/run_gpu_smoke_diffusion_e2e.sh
jiefangxuanyan: Removed in de772f4.

tests/special_e2e/run_gc_diagnostics_qwen_image.sh
jiefangxuanyan: Removed in de772f4.

tests/gpu_smoke/select_gpu_smoke_groups.py
jiefangxuanyan: Reverted in de772f4.

SamitHuang (approved):
LGTM

zhtmike:
This PR is not ready, check discussion.

zhtmike (changes requested):
as comment

verl_omni/workers/engine_workers.py
```diff
@@ -998,7 +1003,18 @@ def _offload_actor_and_empty_cache(self, timings: Optional[dict] = None):
         start = time.perf_counter()
         if self.actor.engine.is_param_offload_enabled:
             self.actor.engine.to("cpu", model=True, optimizer=False, grad=False)
-        aggressive_empty_cache(force_sync=True)
+        gc_diagnostics_point = "actor_offload" if self.gc_diagnostics else None
+        if self.config.model.get("use_regional_compile"):
+            aggressive_empty_cache(
+                force_sync=True,
+                gc_setting=False,
+                gc_diagnostics_point=gc_diagnostics_point,
```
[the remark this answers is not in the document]
jiefangxuanyan: @wtomin @zhtmike, `use_regional_compile` is part of the shared diffusion model config, while not every backend supports regional compilation. Which behavior should we use when it is enabled for an unsupported backend?

1. Reject the configuration with an error.
2. Let the compile and GC paths read the option independently. This is the current behavior, but it can change GC for a backend where regional compilation is not actually enabled, as reported above.
3. Keep the backend’s historical GC behavior when regional compilation is unsupported.

verl_omni/trainer/config/diffusion/engine/diffusion_fsdp.yaml
```diff
@@ -1,5 +1,5 @@
 # Target class for this configuration
-_target_: verl.workers.config.FSDPEngineConfig
+_target_: verl_omni.workers.config.diffusion.DiffusionFSDPEngineConfig
```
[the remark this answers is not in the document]
jiefangxuanyan: Fixed in a78fa13. The tests now expect `DiffusionFSDPEngineConfig` and use `isinstance` so that compatible subclasses do not cause unrelated failures.

verl_omni/workers/config/diffusion/actor.py
```diff
@@ -35,6 +36,14 @@
 ]
 
 
+@dataclass
+class DiffusionFSDPEngineConfig(FSDPEngineConfig):
+    _mutable_fields = FSDPEngineConfig._mutable_fields | {"gc_diagnostics"}
+
+    # Runtime copy of the global GC diagnostics switch; not a separate user setting.
+    gc_diagnostics: bool = False
```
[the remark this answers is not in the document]
jiefangxuanyan: The per-site GC settings were intentionally removed during review. The implementation now derives the GC policy from `use_regional_compile`; the only public diagnostics option is the global `gc_diagnostics` switch introduced by the prerequisite verl PR. The PR description has been updated to match the current API and implementation.

verl_omni/workers/engine/fsdp/diffusers_impl.py
```diff
@@ -730,11 +730,22 @@ def to(self, device: str, model: bool = True, optimizer: bool = True, grad: bool
...
                 load_fsdp_model_to_gpu(self.module)
             if optimizer and self.optimizer is not None:
                 load_fsdp_optimizer(self.optimizer, device)
-            gc.collect()
+            if self.mode in ("train", "eval"):
+                # Preserve the historical full collection unless regional compile is active.
+                # Compiled diffusion training skips these per-step device-transition collections.
+                if not self.model_config.use_regional_compile:
+                    collect_garbage(
+                        diagnostics_point=(f"{self.mode}_device_load" if self.engine_config.gc_diagnostics else None),
```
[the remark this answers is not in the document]
jiefangxuanyan: Updated the PR description to document the compile-derived GC policy, the ownership of `gc_diagnostics` by the prerequisite verl PR, and the current test coverage. The superseded per-site API and dedicated GPU-smoke plan are retained in a collapsed revision-history section.

jiefangxuanyan:
@zhtmike @wtomin, following the recent discussion, I have revised this PR to remove the three remaining full-GC sites in verl-omni directly, rather than expose GC configuration or select the behavior based on regional compilation.

This follows the direction of [verl #7873](https://github.com/verl-project/verl/pull/7873), which removes the corresponding upstream sites and handles weight-transfer cleanup after fixing its CUDA IPC release ordering. The verl-omni change removes the shared train/eval device-load collection and replaces actor-offload full GC with device synchronization plus allocator-cache cleanup.

The diagnostic records support this change: the three collections produced no CUDA allocated or reserved reduction and no repeatable RSS decrease in steady state. The two-hour Qwen-Image run with these collections disabled also showed no CUDA accumulation or greater per-rank RSS growth than the full-GC run. I kept the previous configuration and diagnostics designs only as collapsed revision history in the PR description.

Please take another look when convenient.

zhtmike (approved):

zhtmike:
Proved with evidence.

verl_omni/workers/config/diffusion/actor.py
jiefangxuanyan: The implementation no longer imports the configurable GC APIs or depends on verl #7780. It now removes the three verl-omni full-GC sites directly, so no verl pin update is required for this PR. Addressed in 50be823.

verl_omni/workers/engine/fsdp/diffusers_impl.py
jiefangxuanyan: The implementation no longer calls collect_garbage. The remaining verl-omni sites are now removed directly, following the direction of verl #7873. Addressed in 50be823.

verl_omni/workers/engine_workers.py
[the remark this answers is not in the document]
jiefangxuanyan: GC behavior is no longer derived from use_regional_compile; the GC configuration and diagnostics forwarding have been removed. Unsupported regional-compile backends therefore cannot have their GC behavior changed by this option. Addressed in 50be823.

zhtmike:
All CI green (except title)

PR #532, opened 2026-09-04
[diffusion, rollout, reward, data, recipe] feat: add FLUX.1-dev DanceGRPO training

RhoHwai:
### What does this PR do?

Add an end-to-end, full-parameter FLUX.1-dev DanceGRPO training path to verl-omni, using vLLM-Omni for rollout, the existing diffusion trainer for Actor updates, and HPSv2.1 as the default reward.

The implementation:

- adds a FLUX.1-dev rollout adapter that preserves separate CLIP and T5 prompt token IDs, performs FLUX-native prompt encoding, applies the shifted DanceSDE schedule, and returns explicit current/next latent pairs with rollout log-probabilities;
- adds a diffusers-based Actor adapter that reconstructs the same FLUX inputs and transition schedule for DanceGRPO policy updates;
- reuses the existing DanceGRPO trainer, diffusion configuration contracts, scheduler utilities, and rollout/Actor interfaces instead of adding FLUX-specific branches to `ray_diffusion_trainer` or introducing new trainer configuration fields;
- adds an HPSv2.1 custom reward scorer with lazy per-worker loading and CPU inference by default;
- adds explicit train/test prompt conversion to the standard chat-style parquet schema;
- adds an Ascend NPU full-parameter launcher and documentation for model, data, and HPSv2 checkpoint paths;
- adds CPU tests for transition selection, SDE windows, token handling, rollout/Actor schedule alignment, Actor guidance defaults, and HPSv2 image conversion.

The default recipe uses 720×720 images, 16 denoising steps, rollout guidance `3.5`, Actor guidance `1.0`, scheduler shift `3.0`, DanceSDE noise `0.3`, and a random 60% transition subset after dropping the final transition.

This PR is intentionally limited to FLUX.1-dev full-parameter DanceGRPO training with HPSv2.1. HPSv3 and a FLUX LoRA recipe are out of scope. The provided launcher is Ascend NPU-only.

### Relationship to existing work and non-duplication

Related issue/PR: None.

### Test

Pre-commit checks for the PR-owned files and generated trainer configuration:

```bash
PR_TITLE='[diffusion, rollout, reward, data, recipe] feat: add FLUX.1-dev DanceGRPO training' \
python tests/special_sanity/check_pr_title.py

pre-commit run ruff --files <changed-files>
pre-commit run ruff-format --files <changed-files>
pre-commit run mypy --files <changed-files>
pre-commit run autogen-trainer-cfg --all-files

[...]

Issue #275, opened 2026-07-16
RFC: L3 Nightly Regression for Qwen-Image + FlowGRPO


wtomin:
### Feature request

This RFC describes the architecture for **nightly regression**: run 20 steps of FlowGRPO training on local tiny random `Qwen-Image` weights with a single fixed sample, use a local tiny random Qwen-VL reward model, and detect **numerical drift** and **performance regressions** in one job.

The core idea is to complete both checks in a single run while avoiding cross-interference via **step partitioning**:

- **Steps 1-2**: dump forward intermediates and LoRA gradients; compare against baseline at tensor granularity.
- **Steps 3-20**: train normally and record timing / throughput / memory; compare performance against baseline (dump steps are excluded from performance stats due to I/O overhead).

All diagnostic logic lives **under `tests/nightly/`** and does not touch `verl_omni/` production code.

---

## Overall Architecture

```text
┌─────────────────────────────────────────────────────────────────┐
│  tests/nightly/qwen_image_flowgrpo_single_sample/               │
│  ┌──────────────┐  ┌──────────────────┐                         │
│  │ run.py       │→ │ install_debug_   │                         │
│  │ (entrypoint) │  │ hooks.py         │                         │
│  └──────┬───────┘  │ (hooks + I/O)    │                         │
│         │          └──────────────────┘                         │
│         │ env: DEBUG_DUMP_*                                     │
│         ▼                                                       │
│  ┌──────────────────────────────────────────────────────────┐   │
│  │ verl_omni.trainer.main_diffusion  (unchanged)            │   │
│  └──────────────────────────────────────────────────────────┘   │
│         │ post-run                                              │
│         ▼                                                       │
│  ┌─────────────────┐  ┌──────────────────┐                      │
│  │ compare_dumps.py│  │ collect_metrics.py│                     │
│  │ (dump steps)    │  │ (skip dump steps) │                     │
│  └─────────────────┘  └──────────────────┘                      │
└─────────────────────────────────────────────────────────────────┘
```

Lifecycle of one nightly run:

```mermaid
flowchart LR
    A[prepare single repeated data] --> B[run.py]
    B --> C[20-step training]
    C --> D{step}
    D -->|in DEBUG_DUMP_STEPS| E[dump tensors]
    D -->|all steps| F[log metrics]
    E --> G[compare_dumps]
    F --> H[collect_metrics]
    G --> I[pass/fail]
    H --> I
```

---

### Fixed Workload

[...]

PR #564, opened 2026-09-08
[rollout, vllm_omni, tests] fix: pad dual-GRPO AR responses and pick MASTER_PORT outside the ephemeral range

zhtmike:
### What does this PR do?

Fixes the two remaining recurring `gpu_smoke` failure classes on `main`, one
root-caused in verl-omni and one root-caused in vllm-omni (mitigated on our
side, upstream fix spelled out below):

1. **[verl-omni root cause]** `test_composite_agent_loop.py::test_single_turn`
   crashes in `DiffusionAgentLoopWorker._postprocess` with
   `RuntimeError: Sizes of tensors must match except in dimension 0. Expected
   size 20 but got size 18` (gpu_smoke run #732 on the merge of #532). The
   per-request agent loop runs HF `generate` with batch size 1, so each
   dual-GRPO AR response stops at its own EOS and samples arrive with
   *different* lengths; the batch `torch.cat` then fails. This PR right-pads
   `llm_response_ids` / `llm_all_log_probs` to `rollout.max_new_tokens` and
   emits `llm_response_attention_mask` (1 = real token, 0 = padding), mirroring
   verl's `attention_mask` convention. Right-padding is verified against the
   batched path: HF `generate` fills finished rows with `pad_token_id` after
   their EOS (`transformers/generation/utils.py`: `next_tokens *
   unfinished_sequences + pad_token_id * (1 - unfinished_sequences)`), so the
   padded agent-loop tensors are byte-identical to what the batched rollout
   would produce for the same samples. The batched (non-agent-loop) rollout
   path already gets uniform lengths from batched HF `generate`, which is why
   only the async/composite path crashed. With a real text encoder this is a
   near-certain production crash (EOS ends every caption), not just a CI flake
   — the tiny random CI model simply almost never samples EOS into top-5.

2. **[vllm-omni root cause, mitigated here]** 16 of the 17 failed `gpu_smoke`
   pushes on `main` since Aug 20 died with `torch.distributed.DistNetworkError
   ... EADDRINUSE` while a `DiffusionWorker` bound its rendezvous port, which
   surfaces as `EOFError` in `multiproc_executor._launch_workers` →
   "Orchestrator initialization failed" (runs #550, #560, #572, #591, #613,
   #617, #622, #624, #628, #631, #669, #699, #705, #706, #716, …). Root cause
   chain in vllm-omni: `OdConfig._resolve_master_port` probes a port
   (bind-then-close) at engine-config time, but the rank-0 `DiffusionWorker`
   only listens on it seconds later via torch's default `env://` init
   (`vllm_omni/diffusion/distributed/parallel_state.py:392`), so the number
   floats unprotected across worker spawn + model load. verl-omni contributed
   to the exposure by picking `MASTER_PORT` inside the kernel ephemeral range
   via `get_free_port(bind(0))` — and defeating the reservation on the next
   line (`with_alive_sock=True` immediately followed by `.close()`). Ports in
   `ip_local_port_range` can be claimed as *source ports* by unrelated outgoing
   connections (HF downloads, ray, health checks) in that window; the kernel
   never auto-assigns ports below the range.

   This PR makes verl-omni pick `MASTER_PORT` below the ephemeral range
   (marked with an in-code `TODO (mike)` to drop once vllm-omni ships the
   upstream fix)
   (`verl_omni/utils/net_utils.py::get_non_ephemeral_free_port`, reading
   `/proc/sys/net/ipv4/ip_local_port_range`), removing the dominant claimant
   class. The same principle is why vLLM's own HTTP server defaults to a low
   port (8000) rather than an ephemeral one. A reservation socket cannot be
   held instead: vllm-omni's `settle_port` treats a held port as busy and
   walks away from it.

   **Upstream recommendation for vllm-omni** (the actual root fix, matching
   vLLM's own v1 `multiproc_executor` standard): stop passing a floating TCP
   port to `env://` init — pass an explicit `distributed_init_method` backed
   by a FileStore (`vllm.distributed.utils.get_file_store_init_method`, as
   `vllm/v1/executor/multiproc_executor.py` does), or create the TCPStore in
   the executor parent and let workers connect as clients. Either removes the
   listen race entirely.

### Checklist Before Starting

- [x] Search for similar PRs:
  - `gh pr list --state open --search "llm_response_ids in:body"` → empty
  - `gh pr list --state open --search "EADDRINUSE in:body"` → empty
  - `gh pr list --state open --search "master port in:title"` → empty
  - Issue search for nightly/port/flaky signatures → only RFC #275 (nightly design), unrelated.
- [x] Title format `[rollout, vllm_omni, tests] fix: ...` (validated against `tests/special_sanity/check_pr_title.py` module list).

### Test

Local (macOS, miniforge `verl-omni-py312`):

```
$ python -m pytest tests/utils/test_net_utils_on_cpu.py tests/agent_loop/test_composite_agent_loop_padding_on_cpu.py -v
6 passed, 20 warnings in 11.13s

$ python -m pytest tests/pipelines/test_diffusion_io_spec_on_cpu.py tests/utils/test_net_utils_on_cpu.py \
    tests/agent_loop/test_composite_agent_loop_padding_on_cpu.py -q
22 passed in 9.08s

$ pre-commit run --files <the six changed files>
ruff / ruff format / mypy / generated-yaml / doc / license / device-api /
DataProto / validate-structure / naming / compileall — all Passed
```

Remote (GPU, to run with the `ci-core` label or on a dev box before merge):

```
env CUDA_VISIBLE_DEVICES=0,1 python3 -m pytest -s tests/agent_loop/test_composite_agent_loop.py
env CUDA_VISIBLE_DEVICES=0,1 python3 -m pytest -s tests/workers/rollout/rollout_vllm/test_vllm_omni_generate.py
```

Expected: both green; `llm_response_attention_mask` present with shape
`(bsz, max_new_tokens)`; server logs now show `MASTER_PORT` below 32768.
Not yet executed on GPU by the author — that is the one evidence gap of this PR.

### API and Usage Example

New batch key on the composite (dual-GRPO) agent-loop output, alongside
`llm_response_ids`:

```python
result.batch["llm_response_ids"]                  # [bsz, max_new_tokens], right-padded
result.batch["llm_response_attention_mask"]       # [bsz, max_new_tokens], long, 1=token 0=pad
result.batch["rollout_llm_log_probs"]             # [bsz, max_new_tokens, vocab], zero-padded
```

### Design & Code Changes

- `verl_omni/agent_loop/composite_agent_loop.py`: new
  `_pad_llm_generation_outputs()` helper (right-pad ids with
  `tokenizer.pad_token_id` — falling back to 0 like verl's `_pad_token_ids`;
  zero-pad log-probs; long attention mask; `ValueError` when a response exceeds
  `max_new_tokens`). Wired into `_agent_loop_postprocess` before the existing
  extra-field padding loop; `rollout_llm_log_probs` now derives from the padded
  tensor.
- `verl_omni/utils/net_utils.py`: `ephemeral_port_range()` + `get_non_ephemeral_free_port()`
  (random start, bind-probe, walks past busy ports, fails loudly when the
  candidate range is exhausted).
- `verl_omni/workers/rollout/vllm_rollout/vllm_omni_async_server.py`: use the
  new picker for `MASTER_PORT`; drop the defeated
  `get_free_port(..., with_alive_sock=True)` + immediate `close()` misuse and
  its import.
- Tests: `tests/agent_loop/test_composite_agent_loop_padding_on_cpu.py`
  (ragged padding, pass-through, over-length rejection, and the run-#732
  cat-after-padding regression), `tests/utils/test_net_utils_on_cpu.py` (range,
  busy-port skip, exhaustion), GPU contract extended in
  `tests/agent_loop/test_composite_agent_loop.py`.

- [x] Read the Contribute Guide.
- [x] `pre-commit run` on all changed files — passed (see Test).
- [ ] Docs: no schema/doc page lists the composite agent-loop batch keys; the
      new mask is documented in the helper docstring. Upstream vllm-omni issue
      to be filed for the FileStore fix (this PR body carries the write-up).
- [x] Unit tests added (CPU) + GPU contract updated; GPU execution pending
      (commands above).

---

**AI assistance (ZCode) was used for this change** — root-cause analysis, code,
tests, and this description. The human submitter has reviewed every changed
line and will run the GPU commands above before requesting review. No open PR
addresses either area (checks cited above).

### DIFF
--- a/tests/agent_loop/test_composite_agent_loop.py
+++ b/tests/agent_loop/test_composite_agent_loop.py
@@ -78,14 +78,19 @@ def _assert_text_encoder_outputs(result: DataProto, *, batch_size: int, max_toke
     """Validate Qwen-Image text-encoder returns by rollout."""
     llm_response_ids = result.batch["llm_response_ids"]
     llm_all_log_probs = result.batch.get("rollout_llm_log_probs")
+    llm_attention_mask = result.batch["llm_response_attention_mask"]
     text_encoder_responses = result.non_tensor_batch["text_encoder_responses"]  # list[str]
     _assert_non_empty_tensor(llm_response_ids, "llm_response_ids")
     _assert_non_empty_tensor(llm_all_log_probs, "llm_all_log_probs")
+    _assert_non_empty_tensor(llm_attention_mask, "llm_response_attention_mask")
 
     assert llm_response_ids.shape == (batch_size, max_token_len)
     if llm_all_log_probs is not None:
         assert llm_all_log_probs.shape[1] <= max_token_len
         assert llm_all_log_probs.shape == (batch_size, llm_all_log_probs.shape[1], llm_all_log_probs.shape[-1])
+    assert llm_attention_mask.shape == (batch_size, max_token_len)
+    assert llm_attention_mask.dtype == torch.long
+    assert llm_attention_mask.min() >= 0 and llm_attention_mask.max() <= 1
     assert len(text_encoder_responses) == batch_size
 
 
@@ -267,6 +272,7 @@ def test_single_turn(init_config, agent_reward_loop: bool):
             "negative_prompt_embeds_mask",
             "rollout_log_probs",
             "rollout_llm_log_probs",
+            "llm_response_attention_mask",
         ]
         expected_non_tensor_batch_keys = ["text_encoder_responses"]
         if agent_reward_loop:
--- a/tests/agent_loop/test_composite_agent_loop_padding_on_cpu.py
+++ b/tests/agent_loop/test_composite_agent_loop_padding_on_cpu.py
@@ -1,0 +1,63 @@
+# Copyright 2026 Bytedance Ltd. and/or its affiliates
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+"""CPU tests for the dual-GRPO AR generation padding."""
+
+import pytest
+import torch
+
+from verl_omni.agent_loop.composite_agent_loop import _pad_llm_generation_outputs
+
+
+def test_pads_ragged_llm_responses_to_max_new_tokens():
+    early_eos_ids = torch.tensor([[101, 2054, 8667, 102]])  # stopped at 4 tokens
+    full_ids = torch.tensor([[101, 2054, 8667, 2055, 6844, 2900]])  # hit max_new_tokens
+    early_eos_log_probs = torch.randn(1, 4, 11)
+
+    padded_early, mask_early, padded_early_lp = _pad_llm_generation_outputs(
+        early_eos_ids, early_eos_log_probs, max_new_tokens=6, pad_token_id=0
+    )
+    padded_full, mask_full, padded_full_lp = _pad_llm_generation_outputs(
+        full_ids, None, max_new_tokens=6, pad_token_id=0
+    )
+
+    assert padded_early.shape == (1, 6)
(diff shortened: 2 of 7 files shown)


zhtmike:
ping @chenyingshu for this, fixing flacky CI

chenyingshu:
> ping @chenyingshu for this, fixing flacky CI

LGTM for Dual-GRPO part.

knlnguyen1802 (approved):

PR #572, opened 2026-09-09
[omni, recipe, tests] feat: MiniCPM-o 4.5 thinker GSPO on AVQA

zhtmike:
Implements #406 — MiniCPM-o 4.5 thinker-only collocated GSPO on the V1
`omni_sync` trainer, LoRA r=32, 4×GPU, based on current `main`.

**What's added**

- **Rollout adapter** (`minicpmo_4_5`) — a runtime clone of vLLM-Omni's
  MiniCPM-o stage 0 with `engine_output_type` flipped from `latent` (the Talker
  bridge) to `text`, so a thinker-only text-out rollout works;
  `model_arch=MiniCPMO45OmniLLMForConditionalGeneration` gives logprobs, and
  weight sync is LLM-only.
- **Training adapter** — `vpm`/`apm` stay present-but-frozen through the
  multimodal `MiniCPMO.forward(data)` path; packed (rmpad) batches fold into the
  single-row scatter and run the inner Qwen3 on FA2 varlen from position ids.
- **Prompt parity** — keeps the three renderings in agreement: agent-loop ids →
  engine prompt (slot ids spliced in id space, never re-encoded) → actor
  recompute. Slot counts are cross-checked against the feature counts.
- **Remote-code shims** (`verl_omni/models/transformers/minicpm_o.py`) — one
  file for all HF remote-code fixes, each documenting its own trigger.
- **`MiniCPMORLHFDataset`** — `RLHFDataset` subclass that reads the media blocks
  verl builds from the `<image>`/`<audio>` markers with no Qwen dependency.
- **vLLM plugin** (`vllm.general_plugins` entry point) — two engine-process
  workarounds for vllm-omni#7384 (input) and #7517 (output); both delete at the
  pin bump.
- **Recipe** —
  `examples/gspo_trainer/minicpm/run_minicpmo_4_5_thinker_gspo_lora_avqa_v1.sh`
  plus a README section: the Qwen3-Omni AVQA recipe with model-specific deltas
  only (checkpoint path, `trust_remote_code`, pipeline name, dataset class, LoRA
  exclusions, HF config overrides).
- **Docs** — the integration guide now documents
  `get_fsdp_ignored_module_names`, the hook this PR adds to `OmniModelBase`,
  plus the NCCL-desync pitfall it prevents; `docs/api/pipelines.rst` lists it
  as a member of the base class.

## Design notes — MiniCPM-specific deviations

- **Frozen towers stay FSDP2-unsharded** (`apm`/`vpm`/`resampler`): their
  forwards are skipped for media-free micro-batches and media presence is not
  DP-balanced (verl balances token counts only). FSDP2 issues collectives per
  managed module — `vpm.embeddings.position_embedding` alone is a
  blanket-wrapped `nn.Embedding` group — so a sharded data-dependent tower
  desyncs the NCCL stream across ranks. Unsharded the skips are
  collective-neutral (~1.2 GB/rank replicated, mitigated by `param_offload`),
  and the engine fails closed if an ignored parameter requires grad.
- **Media towers run per sample / per clip**: the remote towers are
  batch-sensitive (the audio padding mask compares mel-frame lengths against
  post-conv2 positions, ~2× under-mask; batched vpm/resampler bf16 kernels flip
  reduction order at small batch sizes, which the LLM amplifies into interior
  logprob deviations). The patched paths run one sample's slice group / one
  exact-length clip at a time, re-splitting the packed pseudo-row.
- **Audio splice, unpacked path only**: the remote `get_omni_embedding` dedents
  its audio splice out of the batch loop, so only the last row of a bs>1 batch
  is written. Inert on the packed path the recipe ships; keeps the adapter's
  supported unpacked branch correct.
- **Reward decode keeps the answer tags** (stock `naive` manager): the reward
  loop builds its own tokenizer and decodes with `skip_special_tokens=True`,
  while the checkpoint registers `<answer>`/`</answer>` as special tokens, so
  the tags were stripped and every choice-reward score was 0 with the answers
  correct. `OmniRewardLoopWorker` now demotes the two tags on the tokenizer its
  manager decodes with; a `TODO (mike)` records that reaching into the manager's
  tokenizer is not the final design. No-op when the tags are already plain.

## Deviation from the RFC

**AVQA-R1-6K instead of GSM8K** (image+audio → text), authorized by the RFC
author. Consequences, all deliberate: encoders stay **present-but-frozen**
rather than stripped (RFC §1); `use_remove_padding` stays **ON**;
`flash_attention_2` only, since sdpa broke train/rollout consistency in
Qwen3-Omni experiments; sampling inherits verl's T=1.0/top_p=1.0/top_k=-1
defaults, because verl's training side divides logits by the temperature in
bf16 while the engine casts to fp32 first, so any T≠1.0 shows up directly in
the parity metric.

## Status and known numerical floor

Training loop, rollout, reward, and weight sync work end-to-end; a fresh
170-step run with these fixes tracks the reference run's score exactly.

**Known floor: engine↔actor parity pearson sits below the 0.99 gate at trained
states.** Root cause established as cross-implementation bf16 numerics, not a
defect: with bitwise-verified identical inputs the engine's prefill is faithful
(HF on the engine's own embeds agrees to ~1.0 at L0/L6), fp32-vs-fp32 agrees
deep, and a single one-ulp input perturbation reproduces the L35 state collapse
through MiniCPM-o's massive-activation stack (~30× Qwen3-Omni's magnitude, whose
live pearson holds 0.996+ with the same ulp seed). Every wiring layer — slots,
block tables, attention metadata, decode staging — was audited clean on the
exact divergent forwards. Training math is unaffected; the divergence feeds only
the parity metric. Evidence: the e35/e36/e37 records in the PR thread.

## Not a duplicate

Closest work is #550 (offline DPO, same model family, different algorithm) and
the Qwen3-Omni AVQA recipes (different architecture; hyperparameters only).

**Three PRs currently create `verl_omni/pipelines/minicpm/`**: this one, #550
(offline DPO, open, not draft), and #576 (simplex OPD, draft). The objectives
are distinct — DPO vs GSPO on AVQA vs online teacher-scored OPD — but the
layout collides file-for-file, and the registries are plain dicts, so whichever
lands second must reconcile. Against #550 the shared paths are
`models/transformers/minicpm_o.py`,
`pipelines/minicpm/{__init__,thinker_training_adapter}.py`,
`pipelines/model_base.py`, `utils/fsdp_utils.py`,
`workers/engine/fsdp/omni_impl.py`, and the test files
`test_minicpm_o_on_cpu.py`, `test_minicpm_adapter_on_cpu.py`,
`test_omni_fsdp_engine_on_cpu.py`. #576 additionally collides on
`omni_rollout_adapter.py` and registers the same keys —
`("MiniCPMO", "thinker")` and `minicpmo_4_5` — under the same
`MiniCPMThinkerAdapter` name. The training-adapter base here was vendored from
#550 at `25a0248e`, with @wtomin as commit co-author.

Sequencing should be settled with @wtomin and @NancyFyong before merge; the
materially different training objective is not in question, only who owns the
package layout.

## Test plan
gpu result

螢幕截圖 2026-09-17 下午12 04 37
螢幕截圖 2026-09-17 下午12 05 00
螢幕截圖 2026-09-17 下午12 05 14

cpu test

```
pytest -q --asyncio-mode=auto tests/     # full suite, local CPU
→ 1880 passed, 2 skipped   (skips are the qwen-tts / qwen-omni-utils
                            optional-dependency guards)
pytest -q --asyncio-mode=auto tests/{the 13 files this PR touches}
→ 160 passed, 1 skipped
```

- **108 new or changed test functions**: 99 in the nine new CPU test files
  (prompt_parity 33, adapter 18, minicpm_o 15, vllm_plugin 9, media_inputs 6,
  answer_tags 5, reward_decode 5, rl_dataset 5, launcher 3), plus 9 added to
  existing files (omni_fsdp_engine 27→31, vllm_omni_lora 1→4,
  thinker_adapter 6→7, audio_frame_parity 3→4).
- Sanity gates run locally, all green: `check_device_api_usage`,
  `check_dataproto_usage`, `check_license`, `validate_structure`, `compileall`,
  PR-title modules against `check_pr_title.py`.
- **Remote GPU full-suite re-run still owed** before un-drafting; the counts in
  the earlier revision of this description predate the review commits.

## Reviewer routing

`examples/gspo_trainer/README.md` and `verl_omni/utils/dataset/` changed —
CODEOWNERS / governance path owners: @chenyingshu @NancyFyong. `docs/` is owned
by @wtomin, who is also the #550 author in the layout collision below.

## AI assistance

Implementation, tests, and this description were drafted with ZCode (GLM); a
ZCode-driven self-review pass against the repo rubric preceded the final
cleanup. The human submitter reviewed every changed line. The training-adapter
base vendored from #550 keeps @wtomin as co-author on `25a0248e`.


wtomin:
Could you please extract fsdp2 `ignored_params` related code into a seperate PR? I would like to take a look at this feature alone, which is relatively indepdent and model-agnostic.

SamitHuang (changes requested):
## Review (Files changed)

I went through the MiniCPM adapters, FSDP2 ignore path, reward decode, and the media collation helpers on `4cbbfcf`. The GSPO recipe and CPU tests cover a lot of the wiring, but a few correctness/layering issues should be fixed before merge. GitHub's new Files changed hover-`+` control did not open a line composer here, so the findings are anchored by path:line below (same places an inline thread would land).

**Verdict: request changes.**

### 1. `sample_pixel_slices` treats a CHW tensor as N slices (`media_inputs.py:117-118`)
`if pixel_values.ndim >= 3: return [t for t in pixel_values]` iterates dim 0. That is correct for `(n_slices, 3, H, W)` and wrong for a single slice `(3, H, W)` — you get three `(H, W)` tensors and the vision scatter / span-count check will disagree with the ids. The CPU tests only pass list-of-`(3, H, W)` packs. Guard on `ndim == 4` (or `shape[0] == 3 and ndim == 3` → one slice) and add that case.

### 2. Object-ndarray batches are flattened into one sample (`thinker_training_adapter.py:327-331`)
`_normalize_media_containers` iterates per sample only for `list`/`tuple`. This PR's own comments say DataProto collates ragged media into object-dtype ndarrays. For a padded `bs>1` batch that array takes the `else` branch, `sample_pixel_slices` concatenates every row, and `_require_span_count` can still pass on totals while `get_vllm_embedding` writes every image onto row 0. Unwrap to a sample list first (`_unwrap_collated` already exists). The shipped recipe is rmpad-only, so this is silent until someone turns padding back on.

### 3. MiniCPM-specific decode patch lives on the shared reward worker (`reward_loop.py:77-85`)
`OmniRewardLoopWorker._init_reward_fn` now imports `minicpm_o` for every omni run. The TODO is right: this belongs on tokenizer setup / an adapter hook, not a shared worker. `actor_registers_special_answer_tags` also treats unknown architectures as apply, so a non-MiniCPM omni job pays for a tokenizer rewrite. Please keep the demotion, but gate it behind a model-base hook (same pattern as `get_fsdp_ignored_module_names`).

### 4. FSDP2 `ignored_params` should land as its own PR (`omni_impl.py:261-272`, `fsdp_utils.py:80-132`)
Agreeing with @wtomin. The fork is no longer MiniCPM-only: `diffusers_impl.py` now imports this `apply_fsdp2`, so pin bumps have to re-diff a MiniCPM-motivated wrap path against verl for diffusion too. Extract the hook + wrap fork, then rebase this recipe on it.

### Also
- `patch_minicpm_get_vllm_embedding` (`minicpm_o.py:208-210`) `torch.stack`s per-span `arange`s. Unequal `[start, end)` lengths (different slice token counts) raise at scatter time. `torch.cat` the indices, or fail closed with the span lengths. No CPU test for mixed span lengths.
- Layout collision with #550 / #576 on `verl_omni/pipelines/minicpm/` is already called out; do not merge the second copy of those registries without a reconcile plan.
- Unrelated diffs in this PR: qwen-tts archive URL + `examples/grpo_trainer/qwen3_tts/README.md`, and `ray_omni_trainer.py` `master_port_range`. Those should not ride a MiniCPM GSPO feature unless they are required to green CI; say so if they are.

verl_omni/pipelines/minicpm/media_inputs.py
```diff
@@ -0,0 +1,218 @@
...
+    Returns:
+        Flat list of slice tensors; empty when the sample has no images.
+    """
+    pixel_values = _unwrap_collated(pixel_values)
+    if pixel_values is None:
+        return []
+    if isinstance(pixel_values, torch.Tensor):
+        if pixel_values.numel() == 0:
+            return []
+        if pixel_values.ndim >= 3:
```
SamitHuang, line 117: `ndim >= 3` iterates dim 0. That is correct for `(n_slices, 3, H, W)` and wrong for a single CHW slice `(3, H, W)` — you get three `(H, W)` tensors and the vision scatter / span-count check will disagree with the ids. The CPU tests only pass list-of-`(3, H, W)` packs. Guard on `ndim == 4` (or treat `ndim == 3` as one slice) and add that case.

verl_omni/pipelines/minicpm/thinker_training_adapter.py
```diff
@@ -0,0 +1,466 @@
...
+        batch_audio_feature_lens,
+        normalize_audio_features,
+        sample_pixel_slices,
+        sample_tgt_sizes,
+    )
+
+    device = data["input_ids"].device
+    # Per-sample slice lists, including the collated single-batch form.
+    pixel_values = data["pixel_values"]
+    if isinstance(pixel_values, (list | tuple)):
```
SamitHuang, line 328: `_normalize_media_containers` only walks samples for `list`/`tuple`. This PR’s own comments say DataProto collates ragged media into object-dtype ndarrays — those take the `else` branch, `sample_pixel_slices` concatenates every row, and `_require_span_count` can still pass on totals while `get_vllm_embedding` writes every image onto row 0. Unwrap with `_unwrap_collated` (already exists) before the per-sample loop. Hidden by the rmpad recipe until someone turns padding back on.

verl_omni/reward_loop/reward_loop.py
```diff
@@ -74,6 +74,15 @@ def __init__(
...
     def _init_reward_fn(self):
         super()._init_reward_fn()
+        # TODO (mike): refactor later — the worker should not reach into the reward
+        # manager's tokenizer; the demotion belongs with the checkpoint's tokenizer setup.
+        from verl_omni.models.transformers.minicpm_o import (
+            actor_registers_special_answer_tags,
+            patch_minicpm_answer_tags,
+        )
+
+        if actor_registers_special_answer_tags(self.config):
```
SamitHuang: MiniCPM-specific answer-tag demotion should not live on the shared `OmniRewardLoopWorker`. Every omni reward process now imports `minicpm_o`, and `actor_registers_special_answer_tags` treats unknown architectures as apply. Keep the demotion, but gate it behind an adapter/tokenizer hook (same pattern as `get_fsdp_ignored_module_names`) so non-MiniCPM runs stay untouched.

verl_omni/workers/engine/fsdp/omni_impl.py
```diff
@@ -247,3 +257,122 @@ def _build_lora_module(self, module):
                     submodule.cast_input_dtype_enabled = False
 
         return module
+
+    def _build_fsdp_module(self, module):
```
SamitHuang, line 261: Agree with @wtomin: extract the FSDP2 `ignored_params` path (`get_fsdp_ignored_module_names` + forked `apply_fsdp2` + this `_build_fsdp_module` override) into a separate PR. `diffusers_impl.py` already imports the fork, so pin bumps re-diff a MiniCPM-motivated wrap path for diffusion too. Land the model-agnostic hook first, then rebase this GSPO recipe on it.

SamitHuang (changes requested):
## Request changes (4 inline)

The four threads cover the blockers:
1. `media_inputs.py` — CHW `ndim >= 3` slice split
2. `thinker_training_adapter.py` — object-ndarray batch flatten
3. `reward_loop.py` — MiniCPM demotion on shared worker
4. `omni_impl.py` — extract FSDP2 ignore (agree with @wtomin)

Also: layout collision with #550/#576; unrelated qwen-tts / `master_port_range` diffs should not ride this PR unless needed for CI.

zhtmike:
let me extract the `ignored_params` part first.

verl_omni/reward_loop/reward_loop.py
zhtmike: worker side hook can not be propagate to driver side. I do not have a concrete idea now, let us keep a TODO here temporarily.

zhtmike: I moved the patch to a better place now.

verl_omni/workers/engine/fsdp/omni_impl.py
zhtmike: extracted.

zhtmike:
> let me extract the `ignored_params` part first.

Done.

verl_omni/pipelines/minicpm/media_inputs.py
zhtmike: done.

verl_omni/pipelines/minicpm/thinker_training_adapter.py
zhtmike: done.

verl_omni/pipelines/minicpm/thinker_training_adapter.py
```diff
@@ -0,0 +1,469 @@
...
+        data["pixel_values"] = [sample_pixel_slices(sample) for sample in pixel_values]
+    else:
+        data["pixel_values"] = [sample_pixel_slices(pixel_values)]
+    tgt_sizes = unwrap_collated(data["tgt_sizes"])
+    if isinstance(tgt_sizes, (list | tuple)):
+        data["tgt_sizes"] = [sample_tgt_sizes(sample, device=device) for sample in tgt_sizes]
+    # Mel features stack per clip; the lens must stay 1-D tensors for the tower's hstack.
+    data["audio_features"] = normalize_audio_features(data.get("audio_features"))
+    if data["audio_features"] == []:
```
SamitHuang: `normalize_audio_features` returns a `(n_clips, n_mels, frames)` tensor whenever the batch has audio, and `[]` only when it does not. This `== []` is then a tensor comparison: PyTorch compares elementwise, so a real mel tensor raises `RuntimeError` (shape mismatch) instead of falling through to `batch_audio_feature_lens`.

AVQA micro-batches always have audio, and the wrapped `forward` calls `split_minicpm_forward_kwargs` again on that already-normalized tensor, so the first audio step dies here — before the per-clip tower path. The CPU coverage only passes `[[], []]`.

Use an explicit empty check (the same `None` / `len == 0` test `get_audio_embedding` already uses; do not `== []` a value that may be a tensor). Please add a CPU test that feeds a `(n_clips, 80, frames)` tensor through `split_minicpm_forward_kwargs`.

verl_omni/models/transformers/minicpm_o.py
```diff
@@ -0,0 +1,443 @@
...
+        if llm_config is not None and hasattr(llm_config, "scale_emb"):
+            vllm_embedding = vllm_embedding * llm_config.scale_emb
+
+        vision_hidden_states = [
+            i.type(vllm_embedding.dtype) if isinstance(i, torch.Tensor) else i for i in vision_hidden_states
+        ]
+
+        rows = []
+        batch_size = len(data["input_ids"])
+        image_bound = data.get("image_bound") or [[] for _ in range(batch_size)]
```
SamitHuang: `data.get("image_bound") or [...]` still bool()s the value. A collated `(N, 2)` tensor raises `RuntimeError: Boolean value of Tensor with more than one value is ambiguous`. The scatter tests pass a list that wraps the tensor (`[tensor]`), so they never evaluate this `or`. `get_audio_embedding` below already refuses `or` for that reason.

The training path currently re-derives bounds as lists, so the AVQA recipe may not hit this on the happy path. Please still use an explicit `None` / empty-list check, and add a CPU test that passes a bare 2-D `image_bound` tensor into `get_vllm_embedding`.

Same pattern a few lines down: `data.get("audio_feature_lens") or []` (and `_has_pixel_slices`'s `pixel_values == []`). A tensor there fails the same way.

SamitHuang (changes requested):
## Review (request changes)

Inline comments:

1. **Blocking.** `_normalize_media_containers` compares `audio_features == []` after `normalize_audio_features` has already turned a real batch into a mel tensor. That comparison raises, and every AVQA micro-batch has audio. The wrapped `forward` runs the split again, so the first audio step dies before the per-clip tower. CPU tests only pass `[[], []]`.
2. **Same tensor-truthiness pattern** in `get_vllm_embedding` (`image_bound or ...`), `audio_feature_lens or []`, and `_has_pixel_slices` (`pixel_values == []`). The training path currently stores lists, so the recipe may not hit it; please still use the explicit empty check `get_audio_embedding` already uses, and cover a bare 2-D tensor.

I did not re-run the CPU suite. The disclosed engine/actor pearson floor is unchanged by these two defects.

Package layout still collides file-for-file with #550 (`minicpm_o.py`, `pipelines/minicpm/`, the thinker adapter). Sequencing with @wtomin should be settled before merge.

verl_omni/pipelines/minicpm/thinker_training_adapter.py
zhtmike: done. 

`first audio step dies here` -> false alarm.

verl_omni/models/transformers/minicpm_o.py
zhtmike: done.

SamitHuang:
@wtomin @NancyFyong PTAL, thx

verl_omni/pipelines/minicpm/thinker_training_adapter.py
```diff
@@ -0,0 +1,471 @@
...
+        unwrap_collated,
+    )
+
+    device = data["input_ids"].device
+    # Unwrap DataProto's ragged collation first: a padded bs>1 batch arrives as an
+    # object-dtype ndarray, which the isinstance checks below would otherwise treat
+    # as one sample and flatten every row's slices onto batch row 0.
+    pixel_values = unwrap_collated(data["pixel_values"])
+    if isinstance(pixel_values, (list | tuple)):
+        data["pixel_values"] = [sample_pixel_slices(sample) for sample in pixel_values]
```
SamitHuang: The list branch still treats every element as a sample. A flat list of `(C, H, W)` slices (common processor output for one image at `bs==1`) becomes N samples with one slice each, while `input_ids` stays one row. Span counts can still match (one span per slice) while `get_vllm_embedding` only scatters row 0 and the remaining vision tokens are dropped. The object-array tests use nested per-sample lists, so they miss this.

If `pixel_values[0]` is a 3-D tensor and `len(pixel_values) != batch_size`, wrap as one sample (`[sample_pixel_slices(pixel_values)]`); keep the per-sample map when the lengths match or the elements are already nested. Same missing `else` on `tgt_sizes` two lines down — a bare `(n_slices, 2)` tensor is never wrapped.

SamitHuang (changes requested):
## Review (request changes)

Re-checked `aff00b3` against the previous threads. The earlier blockers are addressed: FSDP2 ignore landed as #623, `minicpm_naive` is on the V1 reward path, CHW/`ndim==4` slice split, object-ndarray per-sample unwrap, mixed-span `torch.cat`, and the tensor-truthiness family (`== []` / `or []`).

**Remaining (inline on `_normalize_media_containers`).** The list branch still maps every element as a sample, so a flat list of `(C, H, W)` slices at `bs==1` is split into N samples. Span counts can still pass while vision embeddings for slices 1..N-1 never scatter. `tgt_sizes` still has no tensor `else`. Add a CPU test with a flat 3-slice pack and `input_ids` shape `(1, seq)`.

Package layout still collides with #550 / #576 on `verl_omni/pipelines/minicpm/`; settle sequencing with @wtomin before merge. I did not re-run the CPU suite.

verl_omni/pipelines/minicpm/thinker_training_adapter.py
zhtmike: done.

Issue #232, opened 2026-07-03
[RFC] Formal Omni-Model Support in verl-omni


zhtmike:
## 1. Motivation

verl-omni currently provides first-class support for **diffusion models** (Qwen-Image, BAGEL, SD3.5, Wan2.2) via a clean plug-in architecture: `DiffusionModelBase` (training) + `VllmOmniPipelineBase` (rollout), registered by `(architecture, algorithm)` pair, with a dedicated Hydra entrypoint (`main_diffusion`), diffusion-specific agent loop, trainer, and config hierarchy.

For **omni models** (Qwen3-Omni and future models with thinker/talker/code2wav multi-stage architecture), support is _ad-hoc_:

- **No entrypoint**: training goes through upstream verl's `main_ppo` directly, which works but lacks omni-specific ergonomics (no `main_omni` wrapper, manual `external_lib` wiring).
- **No model adapter**: Qwen3-Omni integration relies on heavy monkey-patching of `AutoModelForCausalLM`, `peft.get_peft_model`, `verl.utils.tokenizer.hf_processor`, and `verl.utils.tokenizer.hf_tokenizer`.
- **No omni-specific trainer or agent loop**: uses verl's stock `AgentLoopWorkerTQ` and V1 `PPOTrainerSync` (with `rollout.mode: async`) — functional but without multimodal-aware preprocessing hooks.
- **Empty placeholder**: `verl_omni/trainer/omni/__init__.py` exists but is unused.
- **High patch maintenance burden**: every monkey-patch is a potential breakpoint when verl or transformers upgrades.

This RFC proposes formalizing omni-model support with the same plug-in rigor as diffusion models, **built on top of verl's V1 trainer infrastructure** (`PPOTrainer`, `@register_trainer`, `AgentLoopManagerTQ`, `TransferQueue`), maximally reusing existing verl and verl-omni infrastructure.

---

### Goals

1. **Unified entrypoint** (`main_omni`) — a thin wrapper around verl's `main_ppo` that provides omni-specific defaults and ergonomics, **not** a reimplementation of the V0 `TaskRunner` pattern.
2. **Pluggable model adapters** (`OmniModelBase`) for training, registered by `(architecture, stage)` — algorithm-agnostic, covering thinker-only, talker-only, and joint training.
[...]

PR #573, opened 2026-09-09
[omni, worker, cfg] feat: support Qwen3-Omni Thinker training with VeOmni

FoolPlayer:
### What does this PR do?

Add a shared Omni VeOmni engine using the existing architecture/stage adapter registry, with Qwen3-Omni Thinker text/image training as its first supported adapter. Training uses the V1 omni trainer with VeOmni FSDP2 and expert parallelism. The actor trains the full text backbone, keeps the vision/audio towers frozen, and streams per-expert weights to vLLM-Omni. Pin the optional VeOmni installs in CI, Docker and documentation to **0.1.12**, the latest PyPI release on 2026-09-09.

### Checklist Before Starting

- [x] Reviewed #232 and its discussion, including the request for VeOmni integration. Searched open PRs for [VeOmni + Qwen3](https://github.com/verl-project/verl-omni/pulls?q=is%3Apr+is%3Aopen+VeOmni+Qwen3), [VeOmni + thinker](https://github.com/verl-project/verl-omni/pulls?q=is%3Apr+is%3Aopen+VeOmni+thinker), and [232 in:body](https://github.com/verl-project/verl-omni/pulls?q=is%3Apr+is%3Aopen+232+in%3Abody). The related open PRs add DAPO recipes/rewards or async rollout; none implements this VeOmni training backend. PR #231 is already merged and uses native FSDP.
- [x] Title follows the repository convention.

### Design & Code Changes

- Register one CUDA `omni_model/veomni` engine (`OmniVeOmniEngine`) and add `OmniVeOmniActorConfig`, reusing verl's model loading, FSDP2/EP, optimizer, checkpoint and weight-export machinery. Both Omni FSDP and VeOmni resolve the existing `OmniModelBase` registry by `(architecture, model_stage)`; adding another model does not require a new engine registration.
- Add adapter hooks for VeOmni setup, packed inputs and trainable parameters. The existing replay-input hook runs for both backends. Adapters must explicitly opt into VeOmni before model loading, and optional backend imports are lazy. Qwen3-specific helpers live in `pipelines/qwen3_omni/veomni.py`.
- In the Qwen3 adapter, derive modality masks from packed prompt/response boundaries. An image placeholder sampled into the response stays an ordinary token, instead of consuming a nonexistent image embedding.
- The Qwen3 VeOmni adapter rejects unsupported model stages, audio/video feature inputs, LoRA and Ulysses SP before those paths can silently train with incomplete inputs. Before optimizer construction it also rejects a loaded graph with `has_talker=True` or any `talker` / `code2wav` / `code_predictor` module, using the same excluded-module names as the FSDP adapter.
- Expand fused gate/up and down expert tensors to the Hugging Face per-expert format used by vLLM-Omni, including EP rank offsets. Adapt VeOmni's obsolete `cache_position` causal-mask keyword. This shim is specific to VeOmni 0.1.12's generated GPU model, validated with Transformers 5.14.1, and must be rechecked when either dependency changes. A compatible signature is left unchanged; other unknown arguments still raise.
- Align the launcher and both smoke tests with VeOmni 0.1.12 GPU defaults for Qwen3 operators: FlashAttention 2, explicit `fused_triton` MoE, Liger CE/RMSNorm/SwiGLU/RoPE, and Triton load-balancing loss. Reference selectors inherit actor values, including CLI overrides. Keep verl's existing `use_fused_kernels` → `return_log_probs` conversion; operator selection remains independent.
- Provide a V1 MMK12 GSPO launcher, CPU tests, an EP=2 backend smoke test and a V1 full-weight rollout smoke test. Keep both GPU scripts available for manual validation, outside the required omni smoke group until both pass through normal package initialization on the pinned vLLM stack.

### Test

CPU tests against the repository's pinned verl commit `fefb080262e1c015a0ea05f958822a6a512dc795`:

```bash
TORCH_COMPILE_DISABLE=1 TORCHINDUCTOR_DISABLE=1 python -m pytest -q \
  tests/workers/test_omni_veomni_engine_on_cpu.py \
  tests/pipelines/test_qwen3_omni_veomni_on_cpu.py \
  tests/pipelines/test_qwen3_omni_veomni_prompt_region_on_cpu.py \
  tests/workers/config/test_omni_veomni_config_on_cpu.py \
  tests/workers/test_omni_fsdp_engine_on_cpu.py \
  tests/special_sanity/test_gpu_smoke_selector_on_cpu.py
# 83 passed

pre-commit run --all-files --show-diff-on-failure
# All hooks passed, including generated configs, formatting, mypy and sanity checks.
```

CPU coverage includes a second test architecture using the same engine, external adapter loading, rejection of adapters without VeOmni support, input/replay hook dispatch, and native FSDP regression tests. `tests/workers/test_omni_fsdp_engine_on_cpu.py` is an existing regression suite run against this change, not a new file in this PR. Five additional cases reject speech-generation modules or `has_talker=True` before the base optimizer is called, and another checks that the causal-mask shim leaves compatible signatures untouched. Configuration tests cover launcher/smoke defaults, agreement with the installed VeOmni defaults, actor-to-reference override propagation and explicit reference overrides.

GPU backend validation used **VeOmni 0.1.12 from the PyPI wheel**, Transformers 5.14.1, PyTorch 2.11.0+cu130, and two GB200 GPUs:

- The committed backend check now exercises the real `use_fused_kernels` batch flag through `prepare_model_inputs` and `prepare_model_outputs`, instead of passing `return_log_probs` manually. On each rank, fused log probabilities and entropy matched the logits path at temperatures 1.0 and 0.8 (atol=0.03, rtol=0.01, comparing reductions in FP32). Log probabilities also matched direct log-softmax/gather. Backward and optimizer updates passed with finite gradient norms; the vision tower had no gradients.
- The tiny checkpoint has `enable_audio_output=True`, `enable_talker=True` and extra `talker.*`, `talker.code_predictor.*`, and `code2wav.*` tensors. Both strict state-dict loading and actual checkpoint loading passed; the resulting graph has no speech-generation modules, and the optimizer parameter identities exactly match the trainable Thinker parameters. This checks the contract with a synthetic checkpoint, not the full 30B Instruct weights.
- Export followed by the PyPI checkpoint converter restored all four fused expert tensors of the tiny model.
- The shared engine, selected through the real Qwen3 training adapter, completed two FSDP2/EP=2 optimizer updates with images on rank 0 and text on rank 1 using the aligned default operators (FlashAttention 2, fused Triton MoE, Liger CE/norm/MLP/RoPE, Triton load-balancing loss). The updated backend check passed again on 2026-09-18. All 165 exported weight tensors matched exactly across both ranks.

The backend check's normal invocation is:

```bash
torchrun --standalone --nproc_per_node=2 tests/special_e2e/check_qwen3_omni_veomni_backend.py
```

On the validation host it was executed through a small module loader that bypasses unrelated `verl_omni` package initialization, because `vllm_omni` is not installed there; the model, VeOmni engine, sharding, optimizer and export implementations were real. Actual invocation: `python -m torch.distributed.run --standalone --nproc_per_node=2 /tmp/pr573-unified-backend-loader.py`. The loader supplies parent package paths, imports the real Qwen3 training adapter to populate its registry, and executes the committed script with `runpy`.

`bash tests/special_e2e/run_gspo_qwen3_omni_thinker_veomni_smoke.sh` **was not run locally**. Both VeOmni scripts have been removed from the required `ci-e2e-omni` group and remain manual validation tools. End-to-end V1 rollout on the current vLLM 0.28 / PyTorch 2.13 stack remains to be validated.

### 16-GPU training experiment

A 152-step GSPO run on MMK12 completed successfully using the experimental **VeOmni source backend** (2 nodes × 8 H100 SXM 80GB). These curves describe that source-checkout run; full V1 rollout validation of this PR's PyPI VeOmni 0.1.12 stack remains pending. The exact runtime VeOmni source commit was not captured.

| Setting | Value |
|---|---|
| Model and training | Qwen3-Omni-30B-A3B Thinker; full text backbone; vision/audio towers frozen |
| Backend | VeOmni FSDP2, EP=8; V1 omni trainer; rollout TP=2 |
| Data / budget | MMK12 image/text; 4,909 training and 935 validation examples after filtering; 152 steps |
| Batching | 32 prompts × 8 rollouts; PPO mini-batch 32, micro-batch 1/GPU; one PPO epoch |
| Sequence limits | Prompt 1,024 / response 4,096 tokens |
| Optimizer | AdamW, LR 2e-6, weight decay 0.1; GSPO clip 3e-4 / 4e-4; reference KL coefficient 0.001 |
| Observed LR schedule | Cosine decay to zero, with no effective warmup; the declared 10 warmup steps were ignored by the pinned engine's ratio-only scheduler path |
| Sampling | Training: temperature 0.8, top-p 0.9. Validation: temperature 1.0, top-p 0.7, n=1; this V1 path did not apply the declared `do_sample=false` |

| Validation metric | Step 0 | Step 152 |
|---|---:|---:|
| Correctness, including partial credit | 69.68% | 69.84% |
| Format reward (max 0.3) | 0.2296 | 0.2735 |
| Composite reward | 0.7126 | 0.7476 |

Composite reward is `(accuracy + format_reward) / 1.3`; about 96.5% of its increase came from the format component. This single stochastic-validation run does not establish a substantial correctness gain or equivalence with PR #231. The script exited with code 0 after 21 h 53 min; final model/optimizer/state files were present for all 16 ranks. All fetched metrics were finite; four overlong prompts were evicted during training.

**Validation.** All eight raw evaluation points are shown without smoothing; dashed lines mark the step-0 baseline. Correctness includes partial credit.

16-GPU validation: correctness, format reward and composite reward

**Training dynamics.** Faint lines show all 152 raw observations; dark lines show trailing 10-step means with at least one available point. Learning rate is unsmoothed; final labels are raw values.

16-GPU training: reward, entropy, reference KL, clipping, gradients and learning rate

**Rollout diagnostics.** Rollout/actor probability agreement, response length and the fraction reaching the response-length limit, using the same raw/mean convention.

16-GPU rollout diagnostics

**Runtime and memory.** Mean step/generation/actor-update times were 458.6/374.7/38.7 seconds. Step timing excludes separately logged validation; throughput counts total training tokens per second per GPU. CUDA memory covers the actor, with parameter and optimizer offload enabled.

16-GPU runtime and actor memory

The figures were generated with OpenAI Codex assistance from 34 exported metrics, with no server downsampling or outlier removal. Every final metric value was checked against Tracking's summary.

### API and Usage Example

Install VeOmni 0.1.12 on all nodes as described in `docs/start/install.md`, then launch on an existing two-node Ray cluster with MMK12 parquet data:

```bash
NUM_GPUS=8 NNODES=2 ACTOR_EP=8 \
bash examples/gspo_trainer/qwen3_omni/run_qwen3_omni_thinker_gspo_veomni.sh
```

The recipe supports text and images, full-parameter Thinker training, EP, and packed inputs. Vision/audio encoders are frozen. Audio/video training, Talker, LoRA and Ulysses SP are outside this change. Additional Hydra overrides go last. Warmup uses the supported ratio setting and evaluation explicitly sets temperature to zero.

- [x] Applied the full pre-commit suite.
- [x] Updated documentation and usage examples.
- [x] Added CPU coverage and manual GPU smoke scripts; required GPU registration is deferred pending full-stack validation.
- [ ] Human submitter has reviewed every changed line and run the relevant tests.

AI assistance from **OpenAI Codex** was used to organize, implement and test this change and draft this description. This PR is a draft pending human review and full V1 rollout validation.


examples/gspo_trainer/qwen3_omni/run_qwen3_omni_thinker_gspo_veomni.sh
```diff
@@ -0,0 +1,92 @@
...
+    data.val_files="${VAL_FILE}" \
+    data.train_batch_size=32 \
+    data.max_prompt_length=1024 \
+    data.max_response_length=4096 \
+    data.filter_overlong_prompts=true \
+    data.truncation=error \
+    actor_rollout_ref.model.path="${MODEL_PATH}" \
+    actor_rollout_ref.model.lora_rank=0 \
+    actor_rollout_ref.model.use_remove_padding=true \
+    actor_rollout_ref.model.use_fused_kernels=true \
```
SamitHuang, line 35: `OmniFSDPEngine._build_module` rejects `use_fused_kernels=True` for omni models, and `engine_workers.py` still copies this flag onto every train/infer batch. This recipe turns it on with no VeOmni-side check. Either the backend actually supports fused LM kernels for Qwen3-Omni (then document that and cover it in the GPU smoke), or this will crash / miscompute logprobs — default it to false until that is proven.

verl_omni/workers/engine/veomni/omni_impl.py
```diff
@@ -0,0 +1,60 @@
...
+        model_inputs, output_args = super().prepare_model_inputs(micro_batch)
+        model_inputs = self.model_adapter_cls.prepare_model_inputs(model_inputs, micro_batch, self.model_config)
+        if not isinstance(model_inputs, dict):
+            raise TypeError(
+                f"OmniModelBase.prepare_model_inputs must return a dict, got {type(model_inputs).__name__}."
+            )
+        return model_inputs, output_args
+
+    def _build_optimizer(self, module):
+        self.model_adapter_cls.configure_veomni_trainable_params(module, self.model_config)
```
SamitHuang, line 59: This hook only freezes towers after VeOmni has already built the module. Unlike `OmniFSDPEngine._build_module`, nothing here calls `adapter.configure_model` / `get_strip_modules`, so `talker` / `code2wav` / `code_predictor` are never deleted and are not frozen either. On the real Instruct checkpoint those parameters would enter the optimizer. The GPU backend check asserts `not module.has_talker` on a tiny config that never had a talker, so it does not cover this. Fail closed in `setup_veomni` / after load if those modules exist, or strip them before optimizer creation.

tests/gpu_smoke/run_gpu_smoke_omni_e2e.sh
```diff
@@ -30,4 +30,12 @@ run_test 3 "Qwen3-TTS Talker full-parameter GRPO e2e" \
     env CUDA_VISIBLE_DEVICES="${CUDA_DEVICE_LIST}" NUM_GPUS=2 \
     bash tests/special_e2e/run_qwen3_tts_grpo_smoke.sh "${omni_trainer_args[@]}"
 
+run_test 4 "Qwen3-Omni Thinker VeOmni EP=2 image training and export" \
+    env CUDA_VISIBLE_DEVICES="${CUDA_DEVICE_LIST}" \
+    torchrun --standalone --nproc_per_node=2 tests/special_e2e/check_qwen3_omni_veomni_backend.py
+
+run_test 5 "Qwen3-Omni Thinker VeOmni GSPO full-weight e2e (V1)" \
```
SamitHuang, line 37: These two `ci-e2e-omni` jobs are now on the required GPU smoke path, but the PR says the V1 GSPO smoke was not run locally and the backend check was invoked through a `/tmp` loader that bypasses `verl_omni` package init because `vllm_omni` was missing. Per `docs/contributing/ci_cd.md` / `gpu_smoke_tests.md`, L2 should prove the tiny trainer actually finishes. Also, `check_qwen3_omni_veomni_backend.py` defaults to `--moe-implementation fused_triton` while this recipe/smoke uses `MOE_IMPL=fused` / `eager` — CI will not exercise the documented launcher default. Run both scripts on the vLLM 0.28 stack (or keep them out of the omni group until they pass).

verl_omni/pipelines/qwen3_omni/veomni.py
```diff
@@ -0,0 +1,276 @@
...
+            masks[f"{modality}_mask"] = mask
+        else:
+            if prompt_mask is not None and bool(mask.any()):
+                raise ValueError(f"Prompt contains {modality} placeholder tokens but no {modality} features.")
+            masks[f"{modality}_mask"] = torch.zeros_like(input_ids, dtype=torch.bool)
+
+    return masks
+
+
+def freeze_modality_towers(module: torch.nn.Module) -> tuple[str, ...]:
```
SamitHuang, line 143: This freezes `visual` / `audio_tower` and then claims both backends optimize the same parameter set. FSDP Thinker training still goes through `configure_model` → `get_strip_modules()` (`talker`, `code2wav`, `code_predictor`). Those modules are not in this list, and a missing tower raises while a present talker is ignored. If VeOmni materializes the full Instruct graph, GSPO updates talker weights. Either strip/freeze those modules here, or fail in `setup_backend` when `has_talker` is true.

SamitHuang (changes requested):
Requesting changes on the Thinker training contract and on the GPU evidence, not on the shared `OmniVeOmniEngine` + adapter-hook direction.

**Blocking**
1. The VeOmni engine never calls `configure_model` / `get_strip_modules`. FSDP strips `talker` / `code2wav` / `code_predictor` before wrap; this path only freezes `visual` / `audio_tower`. On the Instruct checkpoint those extra modules can enter the optimizer. The tiny-model `has_talker` assert does not cover that.
2. The MMK12 launcher sets `use_fused_kernels=true`. `OmniFSDPEngine` rejects that flag for omni models, and `engine_workers.py` still copies it onto every batch. Prove VeOmni fused LM kernels work here, or leave the default false.
3. Tests 4 and 5 are wired into `ci-e2e-omni`, but the PR says the V1 GSPO smoke was not run and the backend check used a `/tmp` loader that skipped package init. `fused_triton` (backend default) also does not match `MOE_IMPL=fused` / smoke `eager`. Per the GPU smoke guide, keep them out of the omni group until they pass on the pinned vLLM 0.28 stack.

**Non-blocking:** the 16-GPU curves are a different VeOmni source checkout; they do not validate this PyPI 0.1.12 + V1 rollout path. The body also cites `test_omni_fsdp_engine_on_cpu.py`, which is not in the diff. The `create_causal_mask` monkeypatch should be called out as pin-sensitive.

CPU coverage for masks, EP expert mapping, Hydra compose, and fail-closed missing `setup_veomni` is in the right place. Keep this draft until the two GPU scripts actually complete.

verl_omni/workers/engine/veomni/omni_impl.py
FoolPlayer: Addressed in [the latest fix](https://github.com/verl-project/verl-omni/commit/6811a485a212224aebae0bac12a294a630097e34). The loaded Qwen3 graph is now checked before the base optimizer is called: `has_talker=True` or any `talker`, `code2wav`, or `code_predictor` module raises. The exclusion names come from the existing adapter's `get_strip_modules`, including nested-module checks; the shared engine remains model-independent.

One clarification from checking the pinned [VeOmni 0.1.12 model](https://github.com/ByteDance-Seed/VeOmni/blob/fd99abfda9ef4d9d485f0dae14841de88d30963d/veomni/models/transformers/qwen3_omni_moe/generated/patched_modeling_qwen3_omni_moe_gpu.py#L2934): it unconditionally constructs only the Thinker and sets `has_talker=False`, even for speech-enabled configs. The runtime guard makes that dependency assumption explicit.

Five CPU cases cover the rejection paths. The backend smoke now enables audio output/Talker in its config, saves extra Talker/Codec checkpoint tensors, checks strict state-dict loading and actual engine loading, and verifies that optimizer parameter identities exactly match the trainable Thinker parameters. That synthetic checkpoint check passed on two GB200 GPUs; it is not a full 30B Instruct checkpoint run.

examples/gspo_trainer/qwen3_omni/run_qwen3_omni_thinker_gspo_veomni.sh
FoolPlayer: Covered in [the latest fix](https://github.com/verl-project/verl-omni/commit/6811a485a212224aebae0bac12a294a630097e34). The backend smoke now sets the real batch `use_fused_kernels` flag and runs the inherited `prepare_model_inputs` / `prepare_model_outputs` methods, instead of manually passing `return_log_probs`. It verifies the conversion to `return_log_probs=True`, pre-shifted labels and temperature, then compares log probabilities and entropy against the logits path at temperatures 1.0 and 0.8, followed by backward and optimizer updates.

This passed with image inputs on rank 0 and text on rank 1 on two GB200 GPUs, using PyPI VeOmni 0.1.12, Transformers 5.14.1 and PyTorch 2.11. BF16/FP32 reductions are compared in FP32 with atol=0.03 / rtol=0.01. The native Omni FSDP guard is not executed by the VeOmni engine; the README now explains the separate output-protocol and operator selectors.

The default remains true with this backend evidence. Full V1 validation on the pinned vLLM stack is still pending; the scripts are no longer on the required smoke path.

tests/gpu_smoke/run_gpu_smoke_omni_e2e.sh
FoolPlayer: Addressed by removing both VeOmni registrations from `ci-e2e-omni` in [the latest fix](https://github.com/verl-project/verl-omni/commit/6811a485a212224aebae0bac12a294a630097e34); the scripts remain available for manual validation. Existing main-branch omni tests are preserved.

The launcher, reference and both scripts now agree on explicit VeOmni defaults, including `fused_triton` and Liger. The strengthened backend script passed on two GB200 GPUs, but still used the disclosed package loader on PyTorch 2.11 because that host lacks `vllm_omni`. This is not a pass through normal package initialization or the vLLM 0.28 / PyTorch 2.13 stack. Both normal script invocations must pass there before required registration is restored. The PR remains Draft.

verl_omni/pipelines/qwen3_omni/veomni.py
FoolPlayer: Addressed in [the latest fix](https://github.com/verl-project/verl-omni/commit/6811a485a212224aebae0bac12a294a630097e34): `configure_veomni_trainable_params` now validates the loaded graph using `get_strip_modules()` before freezing towers or delegating optimizer construction. A present speech module is rejected even when `has_talker=False`, and nested `code_predictor` is covered. The freeze helper's docstring no longer claims parameter-set equivalence by itself. See the engine-thread reply for the speech-enabled checkpoint regression and its validation limits.

FoolPlayer:
Rebased this branch onto current main (`2905603`) and pushed [the review fixes](https://github.com/verl-project/verl-omni/commit/6811a485a212224aebae0bac12a294a630097e34). Replies on the inline threads describe the Thinker guard, fused-output regression and deferred GPU registration.

For the non-blocking notes: the 16-GPU figures and their source-checkout/PyPI caveat are preserved; `test_omni_fsdp_engine_on_cpu.py` is now explicitly identified as an existing regression suite run against this change; and the README/helper docstring call out the causal-mask shim's VeOmni 0.1.12 / Transformers 5.14.1 validation scope and need to recheck dependency upgrades.

Validation after rebase: 83 relevant CPU tests and all pre-commit hooks passed. The expanded two-GB200 backend check passed, including two updates and exact agreement of all 165 exported tensors. Full V1 rollout on the pinned stack and human review remain pending, so this stays Draft. OpenAI Codex assisted with these changes and validation.

docs/start/install.md
```diff
@@ -107,7 +107,7 @@ python -c "import veomni; print('veomni', veomni.__version__)"
 python -c "from veomni.distributed.offloading import load_model_to_gpu, load_optimizer, offload_model_to_cpu, offload_optimizer; print('VeOmni offloading helpers OK')"
 ```
 
-VeOmni's torch pin has not been validated against torch 2.13 yet — the `--no-deps` install above is expected to work for import/offloading, but a full VeOmni-engine training run on the vLLM 0.28 stack is still pending GPU validation. If you want VeOmni's full `[gpu,dit]` extras (flash-attn variants, liger-kernel, cuda-python, etc.), install them in a separate environment not pinned to vllm 0.28.0; verl-omni does not need them.
+VeOmni's torch pin has not been validated against torch 2.13 yet — the `--no-deps` install above is expected to work for import/offloading, but a full VeOmni-engine training run on the vLLM 0.28 stack is still pending GPU validation. If you want VeOmni's full `[gpu]` extras (flash-attn variants, liger-kernel, cuda-python, etc.), install them in a separate environment not pinned to vllm 0.28.0; verl-omni does not need them.
```
SamitHuang: This sentence still says verl-omni does not need VeOmni `[gpu]` extras (FlashAttention, Liger, etc.), but `run_qwen3_omni_thinker_gspo_veomni.sh` now defaults to `flash_attention_2` + `liger_kernel` + `fused_triton`. Anyone following this `--no-deps` recipe and then launching the documented GSPO script will miss those kernels. Either list the extra packages this Thinker path actually needs (without replacing torch 2.13), or keep the recipe on eager until that stack is the documented install.

.github/actions/gpu-smoke-prepare/action.yml
```diff
@@ -27,10 +27,10 @@ runs:
         uv pip install --system --break-system-packages --no-deps --reinstall "verl @ git+https://github.com/verl-project/verl.git@$(cat .github/verl_pin.txt)"
         uv pip install --system --break-system-packages TransferQueue==0.1.9
         uv pip install --system --break-system-packages --no-deps -e .
-        # veomni's [gpu] extra pins torch 2.9 and would conflict with the vLLM stack.
+        # veomni's [gpu] extra pins torch 2.11 and would conflict with the vLLM stack.
         # See docs/start/install.md "Optional engine backends".
         # TODO: rm --no-deps when VeOmni supports the vLLM torch pin (torch 2.13 as of vLLM 0.28)
-        uv pip install --system --break-system-packages veomni==0.1.11 --no-deps
+        uv pip install --system --break-system-packages veomni==0.1.12 --no-deps
```
SamitHuang, line 33: This pin bump is shared by GPU smoke, L3 nightly, Docker, and type-coverage — every VeOmni consumer, including diffusion. This PR’s Qwen3 scripts are still off the required omni smoke path and were not run through normal package init on the pinned vLLM 0.28 / torch 2.13 stack. Unless diffusion VeOmni jobs were re-run on 0.1.12, keep the version bump in a dedicated env PR (#646) or cite a nightly/diffusion pass on this wheel.

SamitHuang (changes requested):
Re-review of 6811a48.

The previous blocking items on this draft are addressed in code: `configure_veomni_trainable_params` fail-closes on `talker` / `code2wav` / `code_predictor` (and `has_talker`) before the optimizer; the backend check now drives the real `use_fused_kernels` batch flag; both VeOmni GPU scripts are out of `ci-e2e-omni`. CPU coverage for masks, EP mapping, Hydra compose, and adapter rejection is in the right place. Those earlier threads can be resolved.

Still requesting changes:
1. `docs/start/install.md` still tells users they do not need VeOmni `[gpu]` extras, while the GSPO launcher defaults to FlashAttention 2 + Liger + `fused_triton`.
2. The 0.1.11 → 0.1.12 bump is shared (GPU smoke, L3 nightly, Docker). Qwen3 GPU evidence is still a `/tmp` loader on PyTorch 2.11, not the pinned vLLM 0.28 / torch 2.13 stack, and does not cover diffusion VeOmni on this wheel.

Keep this Draft until `tests/special_e2e/run_gspo_qwen3_omni_thinker_veomni_smoke.sh` completes through normal package init on the pinned stack. The 16-GPU curves remain a different VeOmni source checkout.

.github/actions/gpu-smoke-prepare/action.yml
SamitHuang: I think can leave it in https://github.com/verl-project/verl-omni/pull/646

docs/start/install.md
FoolPlayer: Addressed in [this update](https://github.com/verl-project/verl-omni/commit/25023fddb5815f769e4bcf489b9333d9d7ec3dcd). The install guide now separates the base VeOmni/offloading install from the Thinker kernel requirements: local FA2, Liger, and the Triton bundled with the existing torch stack. It explicitly notes that the shared CI/Docker install does not provide local FA2, and removes the blanket claim that GPU extras are unnecessary.

The instructions constrain installed packages before adding dependencies, describe a matching-wheel or source-build FA2 install, and include native `apply_ops_config` / FA2 extension imports plus normal `vllm_omni` and `verl_omni` package checks. The launcher retains the native defaults. Kernel imports passed in the existing torch 2.11 environment; this is explicitly not claimed as validation of torch 2.13 or complete V1 rollout. The latter remains pending and the PR stays Draft.

.github/actions/gpu-smoke-prepare/action.yml
FoolPlayer: Rebased onto main `1c5e9e9`, which includes merged #646. The shared GPU-smoke, L3-nightly, Docker and type-coverage pin changes have disappeared from this PR's diff; 0.1.12 now comes entirely from main. The PR description has been updated accordingly.

The rebase also preserves #623's FSDP adapter hooks. Validation after rebase: 89 relevant CPU tests and all pre-commit hooks passed. This does not add a diffusion-runtime validation claim. Full Qwen3 V1 validation through normal package initialization on the pinned stack remains pending.

docs/start/install.md
FoolPlayer: Follow-up validation after rebasing onto main (including #646): both committed GPU scripts now pass through normal package initialization with PyPI VeOmni 0.1.12, Torch 2.13.0+cu130, vLLM 0.28.0 and the pinned vLLM-Omni commit, on two GB200 GPUs.

- Backend check: fused log-prob/entropy comparisons at two temperatures, two optimizer steps, and EP=2 export of 165 matching tensors all pass.
- Full V1 smoke: two GSPO steps, final validation and repeated actor-to-rollout full-weight sync completed; exit 0, no traceback. The tiny synthetic run checks plumbing, not convergence.
- This exposed two fixes: align the tiny MoE dimension for SM100 with TP=2; use vLLM's native layerwise reload when updating already-packed AR weights. Incoming tensors are cloned because deferred layer loading can outlive a reused IPC bucket. A CPU regression checks changed layouts, two reload rounds, values and preserved kernel-storage pointers.
- 104 CPU tests and the full pre-commit suite pass. Installation docs now list the local FA2/Liger/Triton requirements and the tested stack. The smoke allows 1800s for cold rollout kernel builds; GPU scripts remain manual, and the PR remains a draft for human review.

The PR description has the commands, results and scope limitations.

zhtmike (commented):
nice, finally support EP

zhtmike:
Relaunched gpu test since `Cancelled after 90m`

examples/gspo_trainer/qwen3_omni/run_qwen3_omni_thinker_gspo_veomni.sh
```diff
@@ -0,0 +1,103 @@
...
+    actor_rollout_ref.model.enable_gradient_checkpointing=true \
+    actor_rollout_ref.actor.optim.lr=2e-6 \
+    actor_rollout_ref.actor.optim.lr_scheduler_type=constant \
+    actor_rollout_ref.actor.optim.lr_warmup_steps_ratio=0.05 \
```
zhtmike: your learning rate is different from your PR content, pls double check.

examples/gspo_trainer/README.md
```diff
@@ -600,6 +600,131 @@ binary `<answer>` exact-match reward): `critic/rewards/mean` rose from ~0.73 to
...
+```bash
+torchrun --standalone --nproc_per_node=2 tests/special_e2e/check_qwen3_omni_veomni_backend.py
+```
+
+The adapter rejects a loaded model containing `talker`, `code2wav`, or
+`code_predictor` (or reporting `has_talker=True`) before optimizer construction.
+It uses the same excluded-module names as the FSDP adapter. VeOmni 0.1.12's
+default modeling constructs only the Thinker even with speech enabled in the
+checkpoint config; the runtime check protects against a different backend or
+release changing that behavior.
```
zhtmike: keep it clean, no need to mention test scripts in gspo readme

examples/gspo_trainer/README.md
```diff
@@ -600,6 +600,131 @@ binary `<answer>` exact-match reward): `critic/rewards/mean` rose from ~0.73 to
...
+parallelism for the actor/reference, and vLLM-Omni for text rollout. Install
+VeOmni on every node following the [installation guide](../../docs/start/install.md#optional-engine-backends).
+The recipe uses the MMK12 parquet files prepared above.
+
+Both FSDP and VeOmni use `verl_omni.trainer.main_omni` and the same
+`OmniModelBase` registry keyed by `(architecture, model_stage)`.
+`model_engine=veomni` selects the shared `OmniVeOmniEngine`; the registered
+`Qwen3OmniThinkerAdapter` supplies model-specific behavior. Adding another
+architecture does not require a new engine class or engine registration.
+
```
zhtmike: keep it clean, it is already self-explained in the code.

examples/gspo_trainer/README.md
```diff
@@ -600,6 +600,131 @@ binary `<answer>` exact-match reward): `critic/rewards/mean` rose from ~0.73 to
...
+| --- | --- |
+| `setup_veomni(model_config, engine_config)` | Opt in, validate supported settings before model loading, and install backend integrations such as weight-export handlers. |
+| `prepare_veomni_inputs(model_inputs, micro_batch, model_config)` | Adapt packed inputs after verl's VeOmni transforms; defaults to passthrough. |
+| `configure_veomni_trainable_params(module, model_config)` | Set trainable parameters after parallelization and before optimizer creation; defaults to no-op. |
+
+The existing `prepare_model_inputs` replay hook runs for both backends.
+Keep optional VeOmni imports inside the backend hooks. Adapters that do not
+implement `setup_veomni` fail before model loading; selecting this backend does
+not imply that every registered architecture is supported. Qwen3's helpers,
+including prompt-region masking, live under `pipelines/qwen3_omni/veomni.py`.
```
zhtmike: this should be belong to veomni migrate guide, not here.

examples/gspo_trainer/README.md
```diff
@@ -600,6 +600,131 @@ binary `<answer>` exact-match reward): `critic/rewards/mean` rose from ~0.73 to
...
+Hydra overrides go last, for example `trainer.total_training_steps=2` or
+`--cfg job` to inspect the composed configuration. The learning-rate schedule
+uses `lr_warmup_steps_ratio`; evaluation sets `temperature=0.0` explicitly. This
+recipe does not claim numerical equivalence to PR #231 or the LoRA curves above.
+
+The two scripts below passed through normal package initialization on two
+GB200 GPUs with PyPI VeOmni 0.1.12 and the pinned vLLM 0.28 / PyTorch 2.13
+stack. They remain manual validation tools outside the required
+`ci-e2e-omni` group. These tiny-checkpoint checks validate training and weight
+transfer mechanics, not 30B-model convergence.
```
zhtmike: move them into `how_to_integrate_xxx` migrate guideline

zhtmike (changes requested):
majorly doc issue and may recheck gpu test.

examples/gspo_trainer/qwen3_omni/run_qwen3_omni_thinker_gspo_veomni.sh
FoolPlayer: Fixed in 8009e14. The base LR was already 2e-6, but the launcher's constant schedule with 5% warmup differed from the experiment's effective schedule. It now explicitly uses cosine, lr_warmup_steps_ratio=0.0, lr_min=0.0 and lr_decay_ratio=1.0. I checked the composed launcher through the pinned verl engine's scheduler with PyPI VeOmni 0.1.12: all 152 LR values exactly match the exported historical actor/lr values (max absolute error 0). The PR description now states these settings; the original experiment data/figures and its stochastic-validation caveat are unchanged. The launcher config suite passes (8 tests).

examples/gspo_trainer/README.md
FoolPlayer: Done in 8009e14. Removed the test scripts, validation narrative and backend implementation details from the GSPO README. It now contains installation/data prerequisites, the launch command, supported settings and LR/evaluation defaults. GPU commands and their expected checks are in docs/contributing/integrating_an_omni_model.md under “Validate the VeOmni integration”.

examples/gspo_trainer/README.md
FoolPlayer: Done in 8009e14. Removed the registry/shared-engine implementation explanation from the GSPO README. The recipe links to the existing omni integration guide for contributors who need the adapter/backend details.

examples/gspo_trainer/README.md
FoolPlayer: Moved in 8009e14 to docs/contributing/integrating_an_omni_model.md, “VeOmni backend (optional)”. It covers extending the existing OmniModelBase adapter, hook lifecycle, native operator fields and the Qwen3-specific integration constraints. The GSPO README now only links to that guide.

examples/gspo_trainer/README.md
FoolPlayer: Moved in 8009e14 to the existing omni integration guide, together with both GPU commands, expected checks and the tiny-checkpoint/convergence distinction. The install guide now points there too. Sphinx builds successfully with no new warning messages relative to the pre-edit baseline, the rendered links resolve, and all pre-commit hooks pass. The maintainer-relaunched GPU CI job is still in progress as of this update; the previously completed local PyPI backend/V1 checks remain documented separately in the PR body.

zhtmike (approved):
looks clear for me, may check the gpu smoke result.

zhtmike:
gpu smoke is broken

SamitHuang (approved):
LGTM

FoolPlayer:
Fixed in c9ba94b, on top of the maintainer's main merge 5f11164.

The timeout came from a regression introduced by my full-model layerwise reload change: the Qwen3-TTS loader copies speech-tokenizer encoder buffers and calls encoder.to() before reload finalization, so meta restoration broke it during weight sync. The fix reserves layerwise restoration for models containing vLLM RoutedExperts and keeps dense AR models on direct loading plus vLLM's native post-load processing. It uses the expert-layer type, without architecture-name special cases, and preserves the NPU path.

Validation on the pinned Torch 2.13 / vLLM 0.28 / PyPI VeOmni 0.1.12 stack:
- The new dense-omni regression fails with the original meta-tensor error on the pre-fix worker and passes after the fix. The packed-MoE cross-bucket test still passes. Full CPU subset: 105 passed; all pre-commit hooks passed.
- The previously hanging Qwen3-TTS GPU smoke completed both GRPO steps and both weight updates on two GB200 GPUs, exit 0, no traceback. Gradient norms were finite/nonzero; updates took about 3 seconds each.
- The Thinker VeOmni V1 GPU smoke also completed two GSPO steps, repeated packed-weight updates and final validation, exit 0, no traceback.

Commands and scope are recorded in the PR description. Re-triggering CI on the fixed head; the local tiny-model runs validate execution and weight transfer, not convergence or audio quality.

zhtmike:
All CI green.

PR #470, opened 2026-08-28
[diffusion, ckpt] fix: restore checkpointed async rollout queue

Wx727:
### What does this PR do?

Port the checkpoint-boundary async recovery path from [verl#7037](https://github.com/verl-project/verl/pull/7037) to the independent Diffusion V1 trainer.

Before this change, Diffusion V1 restored the actor and dataloader state but not TransferQueue. In `separate_async` mode, this could lose queued prompt groups and trajectories after a restart.

This PR saves and restores TransferQueue, re-issues checkpointed in-flight prompts, preserves finished trajectories, and restores the configured warmup depth without overfilling it after recovery.

### Checklist Before Starting

- [x] Search for similar PRs: [verl-omni PR search](https://github.com/verl-project/verl-omni/pulls?q=is%3Apr+TransferQueue+checkpoint) and [verl#7037](https://github.com/verl-project/verl/pull/7037).
- [x] Format the PR title as `[{modules}] {type}: {description}`.
  - Proposed title: `[diffusion, ckpt] fix: restore checkpointed async rollout queue`

### Test

Tested with TransferQueue 0.1.9.

CPU regression tests:

```text
pytest -q tests/trainer/diffusion/test_v1_checkpoint_recovery_on_cpu.py
7 passed
```

Coverage includes:

- TransferQueue checkpoint capability detection by version and API availability;
- trainer save/load integration;
- in-flight recovery: re-issue `pending` and `running` groups, remove their partial trajectories, and leave finished groups untouched;
- a real TransferQueue checkpoint save/load/re-issue round trip;
- separate-async warmup top-up based on restored in-flight prompt groups.

Diffusion V1 separate-async GPU smoke:

```text
FlowGRPO diffusion v1 separate_async e2e test passed
```

Checkpoint recovery GPU smoke:

- saved actor, dataloader, and TransferQueue state at `global_step_1`;
- restarted with `trainer.resume_mode=auto`;
- restored checkpointed TransferQueue entries;
- continued training successfully;
- saved `global_step_2`.

### API and Usage Example

There is no user-facing trainer API or configuration change.

Existing resume configuration continues to be used:

```yaml
trainer:
  resume_mode: auto
  save_freq: 1
```

TransferQueue is upgraded from 0.1.8 to 0.1.9 because 0.1.9 is the first stable release providing the public `save_checkpoint` and `load_checkpoint` APIs.

### Design & Code Changes

- Save TransferQueue after the actor and dataloader state for non-sync trainers.
[...]

PR #575, opened 2026-09-10
[trainer, cfg, tests, doc] feat: hybrid rollout switching for diffusion v1 separate-async

cr-gao:
### What

Ports verl's hybrid rollout switching (verl#7373, in the pinned verl `fefb0802`) into `PolicyGradientDiffusionTrainerV1SeparateAsync`: when the replay buffer is short for the next step, the colocated (hybrid) replicas are lent to generation at step end and reclaimed once `switch_threshold_ratio * train_batch_size` groups are sampleable. It helps when the trainer/rollout GPU split is fixed and the rollout side is the bottleneck (the trainer waits on `timing_s/gen` with the switch off); it does nothing useful when training is the bottleneck.

Off by default. Enable with
```bash
SYNC_COMPATIBLE=false bash examples/flowgrpo_trainer/sd35/run_sd35_medium_ocr_lora_v1_separate_async.sh \
    trainer.v1.separate_async.hybrid_rollout.enable_switch=true
```
`enable_switch` requires `sync_compatible=false` and `rollout.disaggregation.enabled=false`; the remaining knobs follow upstream `HybridRolloutSwitchConfig`.

Default-path change (switch off): the hybrid replicas start in rollout mode and registered with the balancer after init, and are reclaimed (removed from the balancer, aborted, slept) before the warmup batches are fed. Before, `_setup` marked them TRAINER while the init sync had already woken them, so they held rollout memory without serving requests. Validation still lends them and the next step reclaims them.

Diffusion-specific differences from upstream, pinned by CPU tests:
- The lend decision runs after the standalone weight sync: the standalone batch in flight at step end lands during the sync, and deciding before it lent against a phantom gap.
- Hybrid replicas join the balancer only after the naive sync has resumed them; vLLM-Omni's `AsyncOmni.generate()` rejects requests while a sleeping tag is set, so upstream's "register early, requests queue" does not hold.

Also: `prepare_step` / `_consume_sync_metrics` / `_get_n_gpus_for_throughput` on the diffusion base trainer (upstream parity).

### Not a duplicate

`gh pr list --state open --search` for `hybrid_rollout OR enable_switch`, `hybrid switch`, `separate_async hybrid`: no open PR touches this (#560 is the GC knob). #470 only edits `on_init_end`; #380 is the Omni separate-async trainer. Tracks the re-diff item in #445.

### Tests

CPU, on the GPU box at the pins (verl `fefb0802`, vllm `0.28.0`, vllm-omni `ded8934`):
```
pytest --asyncio-mode=auto tests/workers/rollout/rollout_vllm/test_vllm_omni_async_server_lifecycle_on_cpu.py \
  tests/trainer/diffusion/ tests/workers/test_checkpoint_engine_on_cpu.py tests/workers/rollout/**/*_on_cpu.py \
  tests/utils/test_rl_insight_on_cpu.py
# 429 passed on 9f4d791 (the later commits remove the two checkpoint-manager return tests and edit docs only). New: tests/trainer/diffusion/test_v1_hybrid_rollout_switch_on_cpu.py (35 test functions, ported from
# verl's test_separate_async_step_switch_on_cpu.py plus the diffusion ordering / init / validation / warmup-feed
# cases).
```
CI on `4f108c5` (`ready-for-ci`): `cpu_unit_tests` 1153 passed, `gpu_smoke`, `sanity`, `doc`, `pre-commit`, `type_coverage_check`, `secrets_scan`, `check_pr_title` all green.

In-flight lifecycle, 4× RTX PRO 6000 (96 GB), tiny random SD3 FlowGRPO with generation slowed (3000 denoising steps, 1024²) so every reclaim meets in-flight work:
- 1 hybrid + 1 standalone, switch on, 4 steps: 3 lend→reclaim round trips, reclaims aborted 11/11/8/7 in-flight requests, `switch_to_trainer` 1.1–6.5 s (waits for the executing batch), `switch_to_rollout` 1.7 s, all aborted samples retried whole; no errors.
- 2 hybrid + 2 standalone (FSDP, 2 actor ranks), switch on, 5 steps: 4 reclaims + 3 lends, both ranks abort/sleep/sync together, 0 NCCL/collective errors, `switch_to_trainer` 0.75–6.96 s, `switch_to_rollout` 1.76–1.85 s.
- Probe (per-process event logs): in the measured runs no hybrid execution overlapped actor training on the actor GPUs; each hybrid sleep ACK reports the rollout memory freed (2 MiB tiny model, 15.85 GiB SD3.5-M, 21.24 GiB Wan2.2-5B).

SD3.5-Medium OCR LoRA (`run_sd35_medium_ocr_lora_v1_separate_async.sh`, recipe settings: 384², 10 denoising steps, 2 hybrid + 1 standalone + 1 reward, `train_batch_size=16`, `parameter_sync_step=4`, n=8), 20 steps per run, steps 6–20 measured, all within one hour on the same box. Training is the bottleneck here, so the switch must not regress:

| arm | `timing_s/step` median (min–max) | `gen` wait median | hung requests | lends in steps 6–20 |
|---|---|---|---|---|
| `main` 298eb0c (no PR) | 24.46 (23.55–25.17) | 0.067 s | 0 | – |
| this PR, switch off, 2 runs | 24.26 (23.67–25.54), 25.35 (24.05–26.33) | 0.060 s | 0, 0 | – |
| this PR, switch on | 24.69 (23.58–25.63) | 0.064 s | 0 | 0 / 15 |

Step time is set by the actor phases in all arms and the PR heads sit inside the OFF pair's own 1.1 s spread; the post-sync decision sees 12–16 of 16 groups and lends nothing; rewards 0.11–0.13 → 0.20–0.31 in every run.

Wan2.2-TI2V-5B T2V DanceGRPO (`examples/dancegrpo_trainer/wan22/run_wan22_5b_t2v_hpsv3_v1.sh` run in `separate_async` mode; the first video run of v1 separate-async). Three arms × 20 steps, steps 4–20 measured (17 steps, 2176 training videos), one run per arm, consecutive on the same box; an earlier 8-step pair of the two 2+1 arms gave the same −19 %. With 2 hybrid + 1 standalone the geometry is rollout-bound: the standalone executes one request at a time (2.9 s per 9-frame 704×1280 video, ≈ 376 s per 128) against ≈ 250 s of training.

| | OFF, 2 hybrid + 1 standalone | ON, 2 hybrid + 1 standalone | OFF, 2 hybrid + 2 standalone |
|---|---|---|---|
| `timing_s/step` median (min–max) | 377.9 (336.0–418.8) | 307.8 (275.2–355.8) | 250.4 (241.6–252.8) |
| total step time, 17 steps | 6406 s | 5252 s (−18 %) | 4251 s (−34 %) |
| videos/s (2176 videos / total step time) | 0.340 | 0.414 (+22 %) | 0.512 |
| videos/s/GPU (3 / 3 / 4 GPUs) | 0.113 | 0.138 | 0.128 |
| sampling wait per step (`gen` + `switch_wait`), median / total | 130.4 s / 2200 s | 52.2 s / 882 s | 0.1 s / 2 s |
| lends / reclaims | 0 / 1 | 17 / 18 (every decision from step 3) | 0 / 1 |
| `switch_to_rollout` / `switch_to_trainer` | – | 9.1 s (naive sync of the 5B weights + resume) / 3.3 s | – |
| hybrid share: request attempts / completed deliveries | – | 39 % / 19 % (497 of 2616) | – |
| aborted attempts, retried whole (hybrid reclaims / standalone step-end syncs) | – / 2065 | 1838 / 1522 | – / 0 |
| `critic/rewards/mean`, steps 4–20 | 0.674 | 0.690 | 0.681 |
| full run wall, 20 steps incl. startup | 7898 s | 6511 s | 5420 s |

No obvious reward degradation in these short runs; every hybrid sleep frees 21.24 GiB; 0 hung requests, 0 errors in all arms. Reading: at a fixed 3-GPU budget the switch recovers about half of the trainer's idle time (−18 % time, +22 % throughput); a second rollout GPU removes the wait entirely and is faster in total, while the switch keeps ≈ 8 % higher throughput per GPU. The switch is not a substitute for sizing the rollout pool.

Wan2.2 reward curves

<details>
<summary>Wan2.2 reproduction command, deviations from the recipe, per-step rewards</summary>

```bash
TRAIN_BATCH_SIZE=16 MODEL_NAME=<Wan2.2-TI2V-5B-Diffusers> CUSTOM_REWARD_MODEL_PATH=<HPSv3/HPSv3.safetensors> \
bash examples/dancegrpo_trainer/wan22/run_wan22_5b_t2v_hpsv3_v1.sh \
    trainer.v1.trainer_mode=separate_async trainer.n_gpus_per_node=2 \
    actor_rollout_ref.rollout.nnodes=1 actor_rollout_ref.rollout.n_gpus_per_node=1 \
    actor_rollout_ref.rollout.agent.num_workers=2 actor_rollout_ref.rollout.mode=async \
    actor_rollout_ref.rollout.calculate_log_probs=True actor_rollout_ref.rollout.checkpoint_engine.backend=nccl \
    +actor_rollout_ref.rollout.engine_kwargs.vllm_omni.max_num_seqs=1 \
    actor_rollout_ref.actor.ppo_mini_batch_size=4 \
    actor_rollout_ref.rollout.log_prob_micro_batch_size_per_gpu=2 actor_rollout_ref.ref.log_prob_micro_batch_size_per_gpu=2 \
    trainer.v1.separate_async.num_warmup_batches=1 trainer.v1.separate_async.parameter_sync_step=4 \
    trainer.v1.separate_async.sync_compatible=false \
    trainer.v1.sampler.max_off_policy_threshold=8 trainer.v1.sampler.max_off_policy_strategy=drop \
    trainer.val_before_train=False trainer.test_freq=-1 trainer.save_freq=-1 trainer.total_training_steps=20 \
    trainer.v1.separate_async.hybrid_rollout.enable_switch=true   # false for the OFF arms; the 2+2 arm sets rollout.n_gpus_per_node=2
```
Deviations from the recipe: 8 colocated GPUs (sync) → 2 hybrid + 1 standalone `separate_async`; `train_batch_size` 64 → 16; `ppo_mini_batch_size` 16 → 4; `log_prob_micro_batch_size_per_gpu` 16 → 2; `max_num_seqs` explicit 1 (the Wan2.2 DanceGRPO adapter executes one request per batch); validation and checkpoints off. Unchanged: n=8, 704×1280, `num_frames=8` (9 frames after the adapter's VAE rounding), 10 denoising steps, `ppo_micro_batch_size_per_gpu=4`, full fine-tune with parameter/optimizer offload, HPSv3 as the custom reward function in the reward loop (it lands on the first actor GPU), prompts from `wan22_hpsv3.py` on the README's `Hao7/video_prompts`. AutoDL blocks `pidfd_getfd`, so the runs used a small platform plugin that forces the shm path for the naive sync. Adaptive `threshold_ratio` over the ON run (steps 1–20): 0.40 0.40 0.37 0.34 0.34 0.39 0.39 0.39 0.39 0.39 0.39 0.36 0.36 0.41 0.46 0.46 0.43 0.40 0.40 0.45.

Per-step `critic/rewards/mean`:

| step | 4 | 5 | 6 | 7 | 8 | 9 | 10 | 11 | 12 | 13 | 14 | 15 | 16 | 17 | 18 | 19 | 20 |
|---|---|---|---|---|---|---|---|---|---|---|---|---|---|---|---|---|---|
| OFF 2+1 | 0.611 | 0.749 | 0.655 | 0.650 | 0.740 | 0.579 | 0.659 | 0.719 | 0.705 | 0.652 | 0.669 | 0.693 | 0.578 | 0.713 | 0.677 | 0.772 | 0.646 |
| ON 2+1 | 0.622 | 0.750 | 0.673 | 0.692 | 0.747 | 0.523 | 0.732 | 0.758 | 0.679 | 0.695 | 0.645 | 0.673 | 0.632 | 0.730 | 0.697 | 0.786 | 0.702 |
| OFF 2+2 | 0.627 | 0.763 | 0.659 | 0.649 | 0.742 | 0.591 | 0.658 | 0.721 | 0.715 | 0.657 | 0.677 | 0.700 | 0.583 | 0.722 | 0.687 | 0.776 | 0.658 |

</details>

### Known limits

- Default off; enable only when `timing_s/gen` is clearly > 0 with the switch off.
- In rollout-bound runs the standalone step-end sync still waits for the executing batch and aborts its in-flight requests, which are retried as whole samples (the abort counts above); the switch does not change that path.
- Lending a 5B model costs a ≈ 9 s naive weight sync per lend; the cost gate accounts for it from the observed switch history.
- The `ENABLE_SWITCH=1` smoke knob is not wired into the CI matrix.

AI assistance (Claude Code) was used; every changed line was reviewed and the tests above were run by the submitter.

### DIFF
--- a/verl_omni/trainer/diffusion/v1/trainer_base.py
+++ b/verl_omni/trainer/diffusion/v1/trainer_base.py
@@ -240,6 +240,7 @@ class PolicyGradientDiffusionTrainerV1(ABC):
                     with marked_timer("save_checkpoint", self.timing_raw, color="green"):
                         self._save_checkpoint()
                 self.on_step_end()
+                metrics.update(self._consume_sync_metrics())
 
             if self.config.trainer.test_freq > 0 and (
                 is_last_step or self.global_steps % self.config.trainer.test_freq == 0
@@ -285,8 +286,7 @@ class PolicyGradientDiffusionTrainerV1(ABC):
         )
         sample_batch_size = train_batch_size // self.parameter_sync_step
 
-        with marked_timer("feed", timing_raw):
-            self._add_batch_to_generate()
+        prepare_metrics = self.prepare_step()
 
         metrics_aggregator = MetricsAggregator()
         metrics_aggregator.aggregation_rules["sum"].extend(
@@ -297,6 +297,8 @@ class PolicyGradientDiffusionTrainerV1(ABC):
                 "training/rollout_failure/refill_rounds",
             ]
         )
+        if prepare_metrics:
+            metrics_aggregator.add_step_metrics(prepare_metrics)
         prefetched_batches = None
         if self._should_prefetch_local_batches():
             prefetched_batches = []
@@ -327,6 +329,12 @@ class PolicyGradientDiffusionTrainerV1(ABC):
 
         metrics.update(metrics_aggregator.get_aggregated_metrics())
         return KVBatchMeta(partition_id=combined_partition_id, keys=combined_keys, tags=combined_tags)
+
+    def prepare_step(self) -> dict:
+        """Submit this step's prompt batch before any mini-batch is sampled."""
+        with marked_timer("feed", self.timing_raw):
+            self._add_batch_to_generate()
+        return {}
 
     def _step_once(self, metrics: dict, timing_raw: dict, sample_batch_size: int) -> KVBatchMeta:
         """Sample one mini-batch from the replay buffer and run the diffusion PG pipeline."""
@@ -610,10 +618,20 @@ class PolicyGradientDiffusionTrainerV1(ABC):
         """Called at the end of each training step."""
         return
 
+    def _consume_sync_metrics(self) -> dict:
+        """Weight-sync stats stashed by ``on_step_end``, merged into this step's logged metrics."""
+        metrics = getattr(self, "_pending_sync_metrics", None) or {}
+        self._pending_sync_metrics = {}
+        return metrics
+
     @abstractmethod
     def on_sample_end(self):
         """Called after sampling a batch from the replay buffer."""
         return
+
+    def _get_n_gpus_for_throughput(self) -> int:
+        """Return the total number of GPUs used for throughput normalization."""
+        return self.resource_pool_manager.get_n_gpus()
 
     def release_rollout_cache_for_weight_sync(self) -> None:
         """No-op for pure diffusion models (no KV cache)."""
@@ -1502,7 +1520,7 @@ class PolicyGradientDiffusionTrainerV1(ABC):
         data = diffusion_tq_batch_to_dataproto(batch_meta, pad_token_id=self.tokenizer.pad_token_id or 0)
         metrics.update({"training/global_step": global_steps, "training/epoch": epoch})
         metrics.update(compute_data_metrics_diffusion(batch=data))
-        n_gpus = self.resource_pool_manager.get_n_gpus()
+        n_gpus = self._get_n_gpus_for_throughput()
         num_images = (
             data.batch["advantages"].shape[0]
             if "advantages" in data.batch
--- a/verl_omni/trainer/diffusion/v1/trainer_separate_async.py
+++ b/verl_omni/trainer/diffusion/v1/trainer_separate_async.py
@@ -26,6 +26,13 @@ hook semantics, adapted to verl-omni diffusion rollout:
 3. Weight synchronization from actor to standalone rollout uses a non-naive
    checkpoint backend (nccl/nixl/mooncake/...); the colocated rollout still uses
    the naive in-place backend.
+4. Hybrid rollout switching (``separate_async.hybrid_rollout.enable_switch``)
+   lends the colocated replicas to generation at the end of a step when the
+   replay buffer is short for the next one, and reclaims them once enough
+   prompt groups are sampleable. The switch policy is ported method by method
+   from the upstream trainer. Unlike upstream, colocated replicas join the
+   load balancer only after the naive sync has resumed them, because the
+   vLLM-Omni engine rejects requests while any sleeping tag is set.
 
 Diffusion-specific compute (reward, old/ref log-prob, Flow-GRPO advantage,
 actor update, metrics, dumping) lives in ``PolicyGradientDiffusionTrainerV1``;
@@ -35,6 +42,8 @@ wiring.
 
 import logging
 import os
+import time
+from collections import deque
 from enum import Enum
 
 import ray
@@ -42,6 +51,7 @@ from omegaconf import DictConfig
 from transfer_queue import KVBatchMeta
 from verl import DataProto
 from verl.checkpoint_engine import CheckpointEngineManager
+from verl.trainer.config import HybridRolloutSwitchConfig
 from verl.trainer.ppo.utils import Role
 from verl.utils.config import omega_conf_to_dataclass
 from verl.utils.debug import marked_timer
@@ -70,19 +80,24 @@ class PolicyGradientDiffusionTrainerV1SeparateAsync(PolicyGradientDiffusionTrain
 
     Hook behavior:
 
+    - ``_setup``: colocated replicas start in rollout mode and registered with
+      the standalone load balancer, so warmup and validation use both pools.
     - ``on_init_end``: update weights on both the standalone and colocated
       checkpoint managers.
     - ``on_train_begin``: enqueue ``num_warmup_batches`` prompt batches.
-    - ``on_validate_begin``: switch to rollout mode if currently training.
-    - ``on_sample_begin``: switch to rollout mode if training and the switch
-      strategy says so.
-    - ``on_sample_end``: switch to trainer mode (abort + sleep colocated
-      replicas, remove them from the standalone load balancer). When
-      ``sync_compatible`` is True, also pause the standalone rollout.
+    - ``on_step_begin``: reclaim the colocated replicas (abort + sleep, remove
+      from the balancer) when switching is disabled or the replay buffer already
+      holds the switch threshold; otherwise ``prepare_step`` submits this step's
+      prompts and waits for the threshold before reclaiming.
+    - ``on_validate_begin``: switch to rollout mode if currently training; the
+      next ``on_step_begin`` reclaims.
+    - ``on_sample_begin`` / ``on_sample_end``: record replay-buffer shortfall
+      and wait time. When ``sync_compatible`` is True, ``on_sample_end`` also
+      pauses the standalone rollout.
     - ``on_step_end``: after ``parameter_sync_step`` local actor updates, push
-      actor weights into the standalone rollout replicas. Colocated replicas
-      stay slept during training (they share GPUs with the actor) and are only
-      synced at init and when switching to rollout mode. When
+      actor weights into the standalone rollout replicas. With switching
+      enabled, lend the colocated replicas to the next step's generation when
+      the estimated benefit exceeds the recent switch cost. When
       ``sync_compatible`` is True, resume standalone generation after weight
       sync.
     """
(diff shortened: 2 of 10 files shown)


verl_omni/workers/rollout/vllm_rollout/vllm_omni_async_server.py
zhtmike, line 1: I think we need to decouple the model pipeline clearly with the server side implementation. Not every pipeline needs to modify the server endpoint.
@knlnguyen1802 @NancyFyong PTAL.

cr-gao: This is actually not related to specific models, it is a race condition that might occur on any path that runs hybrid mode and only in hybrid mode (both switch ON and OFF) since this is the only mode that will abort a replica that may never be resumed again, which hang late requests forever. I also made sure the actual behavior change is constrained in the diffusion hybrid path. The concrete mechanism is described in PR, I think we may treat this as a bugfix for the newly added mode?

cr-gao: But if we don't want this change, I will remove it and test again, see if the race still happens. It is actually an extra guard, as moving the abort before warmup already avoids some race conditions.

zhtmike: I think we may wait @knlnguyen1802 's comment

NancyFyong: > I think we need to decouple the model pipeline clearly with the server side implementation. Not every pipeline needs to modify the server endpoint. @knlnguyen1802 @NancyFyong PTAL.

we should vllm-omni's fix, If this is a major bug, we could merge it with a TODO; alternatively, if it’s not a widely used feature, we could wait for vllm-omni support to avoid if-else piling up, We need to see @knlnguyen1802  perspective.

cr-gao: Reruned the experiments without the server-side change, no hang occured, so I reverted that change. There are indeed some inconveniences on the server side for diffusion separate async, and I will open a separate issue to discuss changes related to it.

verl_omni/workers/checkpoint_engine.py
```diff
@@ -31,7 +31,7 @@ async def update_weights(self, global_steps: int = None):
             peft_config = self._fetch_actor_lora_peft_config()
             self._lora_peft_config = peft_config
             await self._push_lora_peft_config_to_replicas(peft_config)
-        await super().update_weights(global_steps=global_steps)
+        return await super().update_weights(global_steps=global_steps)
```
zhtmike: no return for this func, do not overwrite the signature.

zhtmike (changes requested):
as comment

verl_omni/trainer/diffusion/v1/trainer_separate_async.py
```diff
@@ -70,19 +82,24 @@ class PolicyGradientDiffusionTrainerV1SeparateAsync(PolicyGradientDiffusionTrain
 
     Hook behavior:
 
+    - ``_setup``: colocated replicas start in rollout mode and registered with
+      the standalone load balancer, so warmup and validation use both pools.
```
knlnguyen1802: With `enable_switch=false` (default, including the SD3.5 recipe and gpu_smoke), `on_train_begin` reclaims hybrid before the warmup feed, so warmup is standalone-only. Validation still lends hybrid. This bullet reads like the switch-on path. Could you say both pools only when `enable_switch=true`, and that default-off reclaims first so warmup requests are not aborted?

verl_omni/trainer/diffusion/v1/trainer_base.py
```diff
@@ -1502,7 +1520,7 @@ def _compute_metrics(self, batch_meta: KVBatchMeta, metrics, timing_raw, global_
         data = diffusion_tq_batch_to_dataproto(batch_meta, pad_token_id=self.tokenizer.pad_token_id or 0)
         metrics.update({"training/global_step": global_steps, "training/epoch": epoch})
         metrics.update(compute_data_metrics_diffusion(batch=data))
-        n_gpus = self.resource_pool_manager.get_n_gpus()
+        n_gpus = self._get_n_gpus_for_throughput()
```
knlnguyen1802, line 1539: `_get_n_gpus_for_throughput()` now drives metrics here. On `separate_async` the subclass adds standalone rollout GPUs to the denominator for every run, not only when the switch is on, so throughput will look lower vs `main` even with the switch off. That matches upstream verl. Worth one line in `docs/start/diffusion_v1.md` calling this a default-path metric change.

verl_omni/trainer/diffusion/v1/trainer_separate_async.py
```diff
@@ -216,71 +269,197 @@ def _setup(self):
...
+
+        if should_switch:
+            switch_start = time.perf_counter()
+            with marked_timer("switch_to_rollout", self.timing_raw, color="cyan"):
+                logger.info("Switching hybrid engine to rollout mode for the next step")
+                self.switch_to_rollout()
+                self.clear_sticky_cache()
+            self._to_rollout_costs.append(time.perf_counter() - switch_start)
+
+    def _get_n_gpus_for_throughput(self) -> int:
```
knlnguyen1802, line 468: Same throughput note as `trainer_base.py` where this override is consumed: the denominator grows on all `separate_async` runs. A short doc callout would help anyone comparing dashboards to `main`.

tests/special_e2e/run_flowgrpo_qwen_image_v1_separate_async.sh
```diff
@@ -44,7 +46,12 @@ NUM_WARMUP_BATCHES=${NUM_WARMUP_BATCHES:-0}
 PARAMETER_SYNC_STEP=${PARAMETER_SYNC_STEP:-2}
 MAX_OFF_POLICY_THRESHOLD=${MAX_OFF_POLICY_THRESHOLD:-1}
 MAX_OFF_POLICY_STRATEGY=${MAX_OFF_POLICY_STRATEGY:-drop}
-SYNC_COMPATIBLE=${SYNC_COMPATIBLE:-1}
+ENABLE_SWITCH=${ENABLE_SWITCH:-0}
```
knlnguyen1802, line 49: The knob is wired, but gpu_smoke always leaves this at 0, so CI still only covers reclaim-before-warmup. Switch-on is on CPU tests plus your GPU runs. Fine to leave out of the matrix; not blocking on that.

verl_omni/workers/checkpoint_engine.py
cr-gao: Reverted.

verl_omni/trainer/diffusion/v1/trainer_separate_async.py
cr-gao: Revised.

verl_omni/trainer/diffusion/v1/trainer_base.py
cr-gao: Added one sentence.

verl_omni/trainer/diffusion/v1/trainer_separate_async.py
cr-gao: Added one sentence.

zhtmike (approved):
Looks good.

knlnguyen1802 (approved):
LGTM now, but I think this one can be more beneficial for Omni model, could you check relevant on that side too @cr-gao

cr-gao:
> LGTM now, but I think this one can be more beneficial for Omni model, could you check relevant on that side too @cr-gao

Sure, will do a follow up.

PR #583, opened 2026-09-11
[diffusion] fix: align Qwen-Image-Edit image tokens and features

Sky-Trigger:
## Summary

- Prefer the raw condition image used to build pre-tokenized Qwen-Image-Edit prompts.
- Fall back to the vLLM-Omni preprocessed condition image when no raw image is present.
- Add regression coverage for both selection paths.

## Problem

vLLM-Omni retains the raw image in multi_modal_data.image and also stores a derived resized image in additional_information.condition_images. Treating these values as equivalent aliases raises a conflict. Selecting only the resized image instead can make the existing image placeholder tokens disagree with the visual features (for example, 324 tokens versus 196 features).

This change keeps strict validation between raw input aliases while excluding the derived resized alias from prompt encoding whenever the raw image is available.

## Tests

- python3 -m pytest tests/pipelines/test_qwen_image_edit_on_cpu.py tests/pipelines/test_rollout_request_on_cpu.py -q (42 passed)
- ruff check on the changed files
- ruff format --check on the changed files
- git diff --check

### DIFF
--- a/verl_omni/pipelines/qwen_image_edit_flow_grpo/vllm_omni_rollout_adapter.py
+++ b/verl_omni/pipelines/qwen_image_edit_flow_grpo/vllm_omni_rollout_adapter.py
@@ -112,6 +112,17 @@ def _validate_condition_image_sizes(condition_images, vae_image_sizes, target_si
         )
 
 
+def _condition_images_for_prompt_encoding(custom_prompt: dict) -> list[Any]:
+    """Use the raw image that was used to build the pre-tokenized prompt."""
+    raw_payload = {
+        key: custom_prompt[key] for key in ("images", "image", "multi_modal_data", "extra_args") if key in custom_prompt
+    }
+    raw_images = condition_images_from_payload(raw_payload)
+    if raw_images:
+        return raw_images
+    return condition_images_from_payload(custom_prompt)
+
+
 @VllmOmniPipelineBase.register("QwenImageEditPlusPipeline", algorithm="flow_grpo")
 class QwenImageEditPlusPipelineWithLogProb(QwenImageTokenIdPromptMixin, QwenImageEditPlusPipeline):
     """Qwen-Image-Edit-Plus rollout pipeline for FlowGRPO."""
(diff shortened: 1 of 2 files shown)


Sky-Trigger:
Does Qwen-Image-Edit not have any CI tests?

NancyFyong:
> Does Qwen-Image-Edit not have any CI tests?

yes, I suspect it wasn't included because its architecture doesn't differ significantly from Qwen-Image.

verl_omni/pipelines/qwen_image_edit_flow_grpo/vllm_omni_rollout_adapter.py
```diff
@@ -112,6 +112,17 @@ def _validate_condition_image_sizes(condition_images, vae_image_sizes, target_si
         )
 
 
+def _condition_images_for_prompt_encoding(custom_prompt: dict) -> list[Any]:
```
NancyFyong: I think it's best to avoid inline names—for the sake of good code style. What do you think?

knlnguyen1802:
> Does Qwen-Image-Edit not have any CI tests?

Could you help to add e2e test for Qwen-Image-Edit maybe by a different PR

Sky-Trigger:
> > Does Qwen-Image-Edit not have any CI tests?
> 
> Could you help to add e2e test for Qwen-Image-Edit maybe by a different PR

I see, i will do it later

Sky-Trigger:
ready for merge now
@knlnguyen1802  @NancyFyong

knlnguyen1802:
Maybe let's process this after https://github.com/verl-project/verl-omni/pull/584

Sky-Trigger:
> Maybe let's process this after #584

The issue is that before this PR is merged, the E2E tests for Qwen Image Edit cannot pass.

knlnguyen1802:
> > Maybe let's process this after #584
> 
> The issue is that before this PR is merged, the E2E tests for Qwen Image Edit cannot pass.

Ok I see

knlnguyen1802 (approved):

PR #574, opened 2026-09-09
[recipe, diffusion, trainer, tests] feat: add LTX-2.3 FlowGRPO v1 sync recipe

SamitHuang:
### What does this PR do?

Closes #389 (LTX-2.3 FlowGRPO v1 sync task).
  Adds the v1 synchronous recipe and infrastructure adaptation for **LTX-2.3 text-to-audio-video FlowGRPO** using **TransferQueue + ReplayBuffer**:
  - Adds `examples/flowgrpo_trainer/ltx2/run_ltx2_3_t2av_lora_v1.sh`.
  - Resolves circular import cycle in `ltx2_flow_grpo` and registers `LTX2DiffusionSingleTurnAgentLoop` in `verl_omni.agent_loop`.
  - Fixes `sampler` parameter forwarding in `LTXForwardContext` (`vllm_omni_rollout_adapter.py`).
  - Fixes non-tensor metadata forwarding (including `audio_sample_rate`) in `DiffusionAgentLoopWorkerTQ`.
  - Refactors video and audio validation logging, artifact dumping, and explicit logger termination in `PolicyGradientDiffusionTrainerV1`.
  - Adds CPU unit test coverage for `LTX2DiffusionSingleTurnAgentLoop` registration.

### Checklist Before Starting

- [ ] Search for similar PRs. Paste at least one query link here: ...
- [ ] Format the PR title as `[{modules}] {type}: {description}` (This will be checked by the CI)
  - `{modules}` include `fsdp`, `vllm_omni`, `rollout`, `trainer`, `ci`, `training_utils`, `recipe`, `ray`, `worker`, `single_controller`, `misc`, `perf`, `model`, `algo`, `env`, `tool`, `ckpt`, `doc`, `data`, `cfg`, `reward`, `diffusion`, `omni`, `tests`, `docker`
  - If this PR involves multiple modules, separate them with `,` like `[diffusion, doc]`
  - `{type}` is in `feat`, `fix`, `refactor`, `chore`, `test`
  - If this PR breaks any API (CLI arguments, config, function signature, etc.), add `[BREAKING]` to the beginning of the title.
  - Example: `[BREAKING][diffusion, fsdp] feat: new rollout scheduler`

### Test

image

image

image

The v1 Trainer is -44.25s (-2.86% faster) than v0 on LTX2.3

### API and Usage Example

  Launch LTX-2.3 FlowGRPO with the V1 trainer:
  ```bash
[...]

PR #585, opened 2026-09-11
[diffusion, data, recipe, doc, tests] feat: add LTX-2.3 TI2VA FlowGRPO

NancyFyong:
### What does this PR do?

Add LTX-2.3 single-first-frame text-and-image-to-audio-video (TI2VA) LoRA FlowGRPO on the V0 trainer while preserving the shared text-only T2AV path.

- The shared `RLHFDataset` explicitly builds structured image/video/audio messages without requiring a processor for media transport. There is no LTX-specific dataset, fake processor, shallow-copy bypass, or installed-dependency patch.
- The LTX agent loop transports the first-frame image separately from Gemma-3 text tokens. Rollout keeps the first latent frame fixed, stores only generated video plus audio in each policy transition, and transports the fixed frame once for Actor replay.
- Include the single-image JSONL-to-parquet converter, production TI2VA launcher, documentation, and CPU regression coverage.
- The TI2VA launcher retains the previously validated long-run settings: rollout TP=1, batch 32 × n=16, 121-frame 10-step 256×384 rollouts, FSDP2 micro-batch 16, three non-contiguous SDE steps from `[0, 10)`, 50-step validation, FA3 on actor and rollout, and checkpoints every 20 steps. Training and validation now both explicitly set `task=ti2va`.

### Checklist Before Starting

- [x] Checked [open LTX TI2VA PRs](https://github.com/verl-project/verl-omni/pulls?q=is%3Apr+is%3Aopen+LTX+TI2VA) and [PRs referencing #585](https://github.com/verl-project/verl-omni/pulls?q=is%3Apr+is%3Aopen+585+in%3Abody). The TI2VA query returned only this PR; the reference query returned none. This updates the existing PR rather than creating duplicate work.
- [x] #574 is complementary: it adds a **V1-sync T2AV** recipe and trainer plumbing, not V0 single-first-frame TI2VA support.

### Design & Code Changes

Reviewer follow-up:

1. **Explicit dataset implementation:** replace the temporary `copy.copy(self)` / `processor=object()` workaround with an explicit `_build_messages` implementation adapted from pinned Verl. Preserve its media payload formats and placeholder-count checks, keep the actual processor unchanged, and keep processor-free text-only negative prompts independent of positive condition images. Initialize the negative-prompt key before upstream construction, where prompt filtering may call `_build_messages`.
2. **T2AV Actor replay:** `prepare_condition` always returns a non-empty dict. Missing T2AV condition metadata becomes a zero-row tensor with the trajectory's batch/width/device/dtype. Tests drive the public `pipelines.utils.prepare_model_inputs` dispatch, including positive/negative CFG inputs, missing/empty/first-frame conditions, and fail-closed TI2VA missing-image behavior.
3. **Shared agent loop:** accept both `images=None` and `images=[]` for text-only prompts, still reject multiple images and reference video/audio. TI2VA's required-image check remains in rollout `run_phase`, and its Actor replay rejects zero condition rows.
4. **Validation configuration:** explicitly set `actor_rollout_ref.rollout.val_kwargs.pipeline.task=ti2va` so the default `None` cannot overwrite the training task. A test executes the launcher with a stub Python executable and checks the actual emitted arguments. Align the model table and README with TI2VA's TP=1 default.
5. **Attention-mask contract:** retain `encoder_attention_mask=None` and `audio_encoder_attention_mask=None` on Actor replay, with an explanatory comment and regression test. This matches pinned vLLM-Omni, rather than dropping a mask used by its DiT:
   - [`ltx2_denoise.build_transformer_kwargs` at pin `ded8934`](https://github.com/vllm-project/vllm-omni/blob/ded8934626aaad1a3e816c3a1d9d742efc012d93/vllm_omni/diffusion/models/ltx2/ltx2_denoise.py#L430-L475) explicitly discards both text masks: the connectors replace padding tokens with learned registers, making all output context tokens valid.
   - [`ltx2_conditioning` at the same pin](https://github.com/vllm-project/vllm-omni/blob/ded8934626aaad1a3e816c3a1d9d742efc012d93/vllm_omni/diffusion/models/ltx2/ltx2_conditioning.py#L338-L379) runs the connector before caching the context shipped to Actor replay. Gemma's earlier token mask is not a DiT mask.
   - The new CPU contract test uses native vLLM-Omni DiT-kwargs construction and CFG combination, a deliberately mask-sensitive mock transformer, and real CPS scheduler sampling/replay. First-transition `ratio_mean` is **1.0** with both CFG=1 and CFG=4. **This is a CPU boundary/numerical regression, not a real-weight GPU rollout/Actor parity result.**

### Test

Fresh validation for the reviewer follow-up, using an environment matching the branch's Verl and vLLM-Omni pins:

```bash
TORCH_COMPILE_DISABLE=1 TORCHINDUCTOR_DISABLE=1 python -m pytest -q \
  tests/pipelines/test_ltx2_flow_grpo_on_cpu.py \
  tests/utils/dataset/test_rl_dataset_on_cpu.py \
  tests/utils/test_omni_rl_datasets_on_cpu.py \
  tests/pipelines/test_processor_preparation_on_cpu.py \
  tests/utils/test_ltx2_ti2va_data_process_on_cpu.py \
  --junitxml=/tmp/pr585-review-regressions.xml -o junit_family=xunit1
```

**105 passed, 19 warnings.** The JUnit properties record `ratio_mean=1.0` for both T2AV mask-contract cases. Before the fixes, the focused new regressions produced **6 failed, 7 passed**, covering the rejected empty image list, missing condition dict, missing validation task, and implicit upstream dataset delegation.

```bash
pre-commit run --files \
  docs/start/models.md \
  examples/flowgrpo_trainer/ltx2/README.md \
  examples/flowgrpo_trainer/ltx2/run_ltx2_3_ti2va_lora.sh \
  tests/pipelines/test_ltx2_flow_grpo_on_cpu.py \
  tests/utils/dataset/test_rl_dataset_on_cpu.py \
  verl_omni/pipelines/ltx2_flow_grpo/agent_loop.py \
  verl_omni/pipelines/ltx2_flow_grpo/diffusers_training_adapter.py \
  verl_omni/utils/dataset/rl_dataset.py
bash -n examples/flowgrpo_trainer/ltx2/run_ltx2_3_ti2va_lora.sh
git diff --check
```

All targeted pre-commit hooks and shell/diff checks passed. Additionally:

```bash
TORCH_COMPILE_DISABLE=1 TORCHINDUCTOR_DISABLE=1 python -m pytest -q tests/special_sanity
```

**3 passed.**

**Historical real-weight validation, before this reviewer follow-up:** the original submission reported 140 stable TI2VA training steps on 8×H800 with a 27,687-sample FLUX first-frame dataset, combined CLAP+ImageBind reward increasing from 0.16 to 0.63, and error-free FA3 execution. The plots below are retained as that earlier experiment's evidence; they are not a rerun of the current patch.

Earlier TI2VA training run
Earlier TI2VA training metrics

**Real-weight T2AV attention-mask follow-up on the current patch:** a standalone diagnostic used the full `dg845/LTX-2.3-Diffusers` checkpoint at revision `01fc4e67e5c88bf8c44f5217b9669b7e496fb985`, native vLLM-Omni rollout with the real Gemma-3 encoder and learned connectors, and an independently loaded two-rank PyTorch FSDP Diffusers Actor replay. Both roles used unchanged weights. The diagnostic ran one text-only sample each at CFG=1 and CFG=4, 256×384, 81 frames, 1024 text positions, four denoising steps, and three recorded CPS transitions.

The real Gemma positive attention mask contained 13 valid positions out of 1024, while the post-connector mask was all-valid (1024/1024). Every normal native-rollout and Actor DiT call received `encoder_attention_mask=None` and `audio_encoder_attention_mask=None`, matching the pinned connector contract.

| CFG | First ratio | Three-transition mean | Maximum absolute error from 1 |
|---|---:|---:|---:|
| 1 | 0.9999960065 | 0.9999824365 | 3.8564e-5 |
| 4 | 0.9999791384 | 0.9996541937 | 8.7225e-4 |

As a negative control, incorrectly forwarding the raw Gemma padding masks at the Actor DiT boundary changed the first-transition ratios to 0.9881081581 (CFG=1) and 0.8415279984 (CFG=4). This real-weight result supports retaining `None` masks and satisfies the requested first-transition check. It does **not** establish strict all-transition parity: CFG=4 transitions 1 and 2 exceeded the recipe's `clip_ratio=1e-4`, so that residual numerical difference remains explicitly unclassified rather than being attributed to the attention-mask contract.

This was a real-weight distributed Actor-replay diagnostic, not a full Ray training E2E: no optimizer update, LoRA synchronization, reward evaluation, VAE decode, or production 24-step run was performed. Existing training jobs were not interrupted. Local validation still does not imply that GitHub CPU/GPU CI has run on the current head.

### API and Usage Example

```bash
python3 examples/flowgrpo_trainer/ltx2/prepare_ti2va_data.py \
  --input_dir ./dataset/ltx2_ti2va \
  --output_dir ./data/ltx2_ti2va

DATA_DIR="$PWD/data/ltx2_ti2va" \
  bash examples/flowgrpo_trainer/ltx2/run_ltx2_3_ti2va_lora.sh
```

Existing text-only T2AV launchers require no new condition fields or custom dataset configuration.

### AI assistance and human review

AI assistance was used: CodeBuddy Code for earlier long-run configuration folding and merge work, and Pi with OpenAI for the processor-free dataset work, reviewer-follow-up implementation, tests, and PR-description update. The original submission recorded human review of the earlier changes. The human submitter remains responsible for understanding and reviewing every changed line and rerunning relevant tests; human review of this latest follow-up and any required fresh GPU evidence remain prerequisites to merge.


verl_omni/pipelines/ltx2_flow_grpo/diffusers_training_adapter.py
```diff
@@ -134,16 +132,73 @@ def prepare_model_inputs(
...
+        cls,
+        micro_batch: TensorDict,
+        latents: torch.Tensor,
+        step: int,
+    ) -> dict[str, torch.Tensor] | None:
+        """Read the fixed first-frame latent captured by rollout."""
+        del latents, step
+        image_latents = micro_batch.get("condition_image_latents")
+        if image_latents is None:
+            return None
```
SamitHuang: `LTX23FlowGRPO` now subclasses `DiffusionI2IModelBase`, so `prepare_model_inputs` in `verl_omni/pipelines/utils.py` rejects `prepare_condition(...) is None`. Returning `None` when `condition_image_latents` is missing will crash the existing T2AV recipe, not only a misconfigured TI2VA run.

Always return a dict (e.g. a zero-row `image_latents` tensor) for T2AV, or keep T2AV on `DiffusionModelBase`.

verl_omni/pipelines/ltx2_flow_grpo/diffusers_training_adapter.py
```diff
@@ -106,21 +101,24 @@ def prepare_model_inputs(
...
             "num_frames": latent_frames,
             "height": latent_height,
             "width": latent_width,
             "fps": frame_rate,
             "audio_num_frames": audio_latents.shape[1],
             "return_dict": False,
+            "_require_image_condition": getattr(model_config.pipeline, "task", None) == "ti2va",
         }
         model_inputs = {
             **common,
```
SamitHuang, line 120: Actor replay now hard-codes `encoder_attention_mask` / `audio_encoder_attention_mask` to `None` for both T2AV and TI2VA. Rollout still encodes with masks in `_encode_token_ids`.

If the DiT uses those masks, this is a silent log-prob mismatch on the existing T2AV recipe. Please either keep the masks, or cite the vLLM-Omni connector behavior and show step-1 `ratio_mean ≈ 1.0` on T2AV after this drop.

SamitHuang, line 120: This shared agent loop still serves T2AV. `if not isinstance(images, list) or len(images) != 1` rejects `images is None` and `images == []`, so text-only T2AV rows will raise here.

Gate the single-image check on `task == "ti2va"` (or equivalent), and treat missing/empty images as the T2AV path.

SamitHuang, line 120: Train sets `rollout.pipeline.task=ti2va`, but validation only overrides height/width/frames/steps here. `_config_to_sampling_dict(val_kwargs.pipeline)` still carries `task=None` and overwrites the train value, so validation can skip the TI2VA missing-image check in `run_phase`.

Add `actor_rollout_ref.rollout.val_kwargs.pipeline.task=ti2va`. Also, `docs/start/models.md` lists TP=2 while this launcher defaults `ROLLOUT_TP=1`.

SamitHuang (changes requested):
## Request changes

TI2VA first-frame conditioning is directionally right, but this PR regresses the shared T2AV path in a few places:

1. `prepare_condition` returning `None` under `DiffusionI2IModelBase` will fail closed in `prepare_model_inputs` for T2AV.
2. Shared agent loop requires exactly one image (rejects `None`/`[]`), so T2AV text-only rows break.
3. Actor drops `encoder_attention_mask` to `None` while rollout still encodes with masks — needs proof T2AV ratio stays ~1.0, or keep the masks.
4. Val kwargs omit `pipeline.task=ti2va`, so validation can overwrite the train task with `None`.

verl_omni/utils/dataset/rl_dataset.py
```diff
@@ -49,6 +50,27 @@ def __init__(self, *args, config: DictConfig, **kwargs):
...
+            example = {key: example[key]}
+        if not any(example.get(media_key) for media_key in (self.image_key, self.video_key, self.audio_key)):
+            return super()._build_messages(example, key)
+
+        # Upstream only checks processor presence here. A shallow view reuses its
+        # parsing/validation without changing the real processor used for filtering.
+        transport_dataset = copy.copy(self)
+        transport_dataset.processor = object()
+        return _UpstreamRLHFDataset._build_messages(transport_dataset, example, key)
+
```
zhtmike, line 140: simply overwrite the entire method `_build_messages` for this.

The monkey patch is silent, and will be broken after verl pin upgrade.

Never do any monkey-patch unless you have a clear confidence.

zhtmike (changes requested):
As comment

verl_omni/pipelines/ltx2_flow_grpo/diffusers_training_adapter.py
NancyFyong: Addressed at `9713164` and followed up with a real-weight T2AV diagnostic.

The Actor intentionally retains `encoder_attention_mask=None` and `audio_encoder_attention_mask=None`, matching pinned vLLM-Omni: Gemma uses its token mask before the learned connectors, while the connectors replace padding positions with valid learned registers and native `build_transformer_kwargs` drops both masks at the DiT boundary. `test_ltx2_t2av_first_step_ratio_matches_native_mask_contract` covers this contract for CFG=1 and CFG=4.

I additionally replayed native vLLM-Omni trajectories from the full `dg845/LTX-2.3-Diffusers` checkpoint with an independently loaded two-rank FSDP Diffusers Actor, without updating weights. The real Gemma positive mask had 13/1024 valid positions, the post-connector mask was 1024/1024 valid, and every normal rollout/Actor DiT call saw both masks as `None`.

First-transition ratios were:

- CFG=1: `0.9999960065`
- CFG=4: `0.9999791384`

As a negative control, incorrectly forwarding the raw Gemma padding masks changed those ratios to `0.9881081581` and `0.8415279984`, respectively. This supports retaining `None` and answers the requested step-1 check.

Caveat: this was a standalone real-weight rollout/FSDP Actor-replay diagnostic, not a full Ray training E2E. Across all three captured transitions, CFG=4 reached a maximum absolute ratio error of `8.7225e-4`, so I am not claiming strict all-step parity or attributing that residual difference to the mask contract. The PR description now records the scope and full measurements.

zhtmike (approved):

SamitHuang:
pls fix the ci

SamitHuang (approved):
LGTM

Issue #563, opened 2026-09-08
minimax h3 t2va diffusionnft reward issue


charon-agi:
According to https://github.com/verl-project/verl-omni/pull/383, i try to reproduce minimax h3 t2va diffusionnft training, use the default config in example: examples/diffusionnft_trainer/minimax_h3/run_minimax_h3_t2va_lora.sh, at 50 step validation,the reward exhibits a slight drop, at 60 step, a large drop, I use 8 h100 gpus，data and env are based on https://github.com/verl-project/verl-omni/blob/main/examples/diffusionnft_trainer/minimax_h3/README.md

image bind reward:
<img width="441" height="315" alt="Image" src="https://github.com/user-attachments/assets/6b369d46-87a9-4ebb-9f07-22fe85d4b7e3" />

clap reward:

<img width="425" height="298" alt="Image" src="https://github.com/user-attachments/assets/a02e8ce0-bf30-4b38-9569-1366f636fe54" />

SamitHuang:
@NancyFyong PTAL thx

NancyFyong:
Thank you very much for using it. It is a bit strange—did you modify any of the parameters? Could you provide more details on the training results?

charon-agi:
@NancyFyong I didn't use minimax h3 text encoder due to gpu memory, so I extract the prompt embedding before training, the vllm omni engine read the embedding directly, this is the only modification I made.

training script:
```
#!/usr/bin/env bash
# MiniMax-H3 text-to-audio-video (t2va) DiffusionNFT LoRA recipe.
set -x

export WANDB_MODE=${WANDB_MODE:-offline}

H3_LOAD_TEXT_ENCODER=${H3_LOAD_TEXT_ENCODER:-False}
export VERL_OMNI_H3_LOAD_TEXT_ENCODER=$H3_LOAD_TEXT_ENCODER

WORKSPACE=${WORKSPACE:-$HOME}
MODEL_PATH=${MODEL_PATH:-}
DATA_DIR=${DATA_DIR:-$WORKSPACE/data/vid_prompt/verl_omni}
NUM_GPUS=${NUM_GPUS:-8}
ROLLOUT_TP=${ROLLOUT_TP:-2}
ROLLOUT_N=${ROLLOUT_N:-2}
MAX_NUM_SEQS=${MAX_NUM_SEQS:-8}
TOTAL_TRAINING_STEPS=${TOTAL_TRAINING_STEPS:-1000}
[...]

PR #589, opened 2026-09-12
[rollout, recipe, tests, doc] fix: preserve diffusion text-encoder TP

NancyFyong:
### What does this PR do?

Fix silent loss of MiniMax-H3 text-encoder tensor parallelism (ETP) in the shared diffusion rollout path used by both DiffusionNFT and FlowGRPO.

Related to the ETP bug reported in #563; this does not claim to resolve the separate reward/convergence discussion there.

The pinned vLLM-Omni CLI accepts `text_encoder_tp_size`, but `OmniEngineArgs.from_cli_args` filters it out because it is not a declared engine/orchestrator field. The fused diffusion engine then defaults to ETP=1. This PR restores the value after CLI conversion, without switching to the AR strategy or adding an environment-variable patch.

The upstream bug is filed as [vllm-omni#7564](https://github.com/vllm-project/vllm-omni/issues/7564). The downstream compatibility shim is tracked by [verl-omni#445](https://github.com/verl-project/verl-omni/issues/445) and should be removed after the upstream fix enters the pin.

### Checklist Before Starting

- [x] Searched for duplicates: [#563 references](https://github.com/verl-project/verl-omni/pulls?q=is%3Apr+is%3Aopen+563+in%3Abody), [text_encoder_tp](https://github.com/verl-project/verl-omni/pulls?q=is%3Apr+is%3Aopen+text_encoder_tp), and text-encoder parallelism keywords. Read #563 and its comments; no matching open fix PR found.
- [x] PR title follows the validated format.

### Test

**CPU routing and regression tests: 98 passed.**

```bash
  tests/workers/rollout/rollout_vllm/test_diffusion_text_encoder_tp_on_cpu.py \
  tests/workers/rollout/rollout_vllm/test_vllm_omni_strategy_on_cpu.py \
  tests/workers/rollout/rollout_vllm/test_vllm_omni_run_server_engine_args_on_cpu.py \
  tests/workers/rollout/rollout_vllm/test_vllm_omni_deploy_config_on_cpu.py \
  tests/workers/config/test_diffusion_config_on_cpu.py
```

Before the fix, the routing regression reproduced requested TP=4/ETP=4 becoming ETP=1. Tests now cover the real CLI conversion and `run_server` boundary through default-stage construction to `OmniDiffusionConfig.parallel_config`, legacy overrides, conflicts, invalid groups, and explicit parallel configurations.

**Tiny-model GPU validation: both passed.** Two H20 GPUs, TP=2/ETP=2, T2VA, 160×288, 97 frames, four inference steps:

| Algorithm | Training | Exit code |
| --- | --- | --- |
| FlowGRPO | 2/2 steps, actor update and weight sync | 0 |
| DiffusionNFT | 2/2 steps, actor update and weight sync | 0 |

Both ranks had encoder group size 2 and embedding shards `[256,32]` from the full `[512,32]` table. Each shard matched its checkpoint slice exactly; each rank completed 16 real encoding calls in each run. This checks actual sharding, not just a config print.

These runs used a local diagnostic harness around the existing tiny runner's `_hydra_overrides`, with NFT-specific overrides for the second run (`python <local-harness>/run.py flowgrpo` / `nft`). The harness and instrumentation are not included in this PR. They use the existing tiny text-width workaround and lightweight VAE/reward fixtures; instrumentation asserts ETP but never assigns it. Tested with vLLM-Omni `0.28.0rc2.dev68+gded893462`, vLLM `0.28.0`, and torch `2.13.0+cu129`.

Also verified:
- All nine H3 launchers pass `bash -n`; 18 stubbed argument-capture checks passed for default ETP=TP and explicit ETP=1.
- All 12 pre-commit hooks passed, including generated-config verification. Full-repository checks used an explicit `git ls-files` list because the local Git lacks the `--deduplicate` flag required by `pre-commit --all-files`.
- The broader rollout CPU suite timed out after 600 seconds: **not counted as passing**. Production-weight memory savings, FL2VA/Ref2VA GPU runs, V1, and NPU execution were not revalidated here.

### API and Usage Example

Use the existing typed field, without `+`:

```bash
actor_rollout_ref.rollout.tensor_model_parallel_size=4 \
actor_rollout_ref.rollout.text_encoder_tp_size=4
```

H3 launchers now consistently forward `TEXT_ENCODER_TP`, defaulting to `ROLLOUT_TP`. The NFT T2VA and FL2VA launchers previously omitted ETP and therefore ran with ETP=1; this is an intentional recipe-default change to full-TP sharding. Set `TEXT_ENCODER_TP=1` to preserve the previous unsharded behavior. The pinned backend accepts only ETP=1 or exactly the configured rollout TP; intermediate subgroups are rejected. ETP is independent of CPU/layerwise offload.

The legacy `+actor_rollout_ref.rollout.engine_kwargs.vllm_omni.text_encoder_tp_size` remains accepted. It overrides the typed default of 1; conflicting non-default typed values fail rather than being silently discarded.

### Design & Code Changes

- Restore and validate ETP in `DiffusionStrategy.prepare_engine_args`, shared by NFT and FlowGRPO.
- Preserve explicit mapping/dataclass parallel configurations and unrelated fields, reject conflicting ETP values, and avoid mutating the input configuration.
- Update nine H3 launchers, the tiny FlowGRPO runner, and configuration/recipe documentation to use the typed field.
- Add diffusion-specific routing regression coverage; leave AR strategy behavior unchanged.

- [x] Run pre-commit checks across all tracked files (explicit-file workaround described above).
- [x] Update documentation.
- [x] Add CPU regression tests under the existing `*_on_cpu.py` discovery convention; GPU validation was manual, not added to CI.
- [x] Human submitter has reviewed every changed line and rerun the relevant tests before requesting review.

### DIFF
--- a/examples/diffusionnft_trainer/minimax_h3/run_minimax_h3_t2va_lora.sh
+++ b/examples/diffusionnft_trainer/minimax_h3/run_minimax_h3_t2va_lora.sh
@@ -9,6 +9,7 @@ MODEL_PATH=${MODEL_PATH:-}
 DATA_DIR=${DATA_DIR:-$WORKSPACE/data/vid_prompt/verl_omni}
 NUM_GPUS=${NUM_GPUS:-8}
 ROLLOUT_TP=${ROLLOUT_TP:-2}
+TEXT_ENCODER_TP=${TEXT_ENCODER_TP:-$ROLLOUT_TP}
 ROLLOUT_N=${ROLLOUT_N:-16}
 TOTAL_TRAINING_STEPS=${TOTAL_TRAINING_STEPS:-1000}
 HEIGHT=${HEIGHT:-256}
@@ -105,6 +106,7 @@ python3 -m verl_omni.trainer.main_diffusion \
     actor_rollout_ref.rollout.max_num_seqs=1 \
     actor_rollout_ref.rollout.rollout_attn_backend=$ROLLOUT_ATTN_BACKEND \
     actor_rollout_ref.rollout.tensor_model_parallel_size=$ROLLOUT_TP \
+    actor_rollout_ref.rollout.text_encoder_tp_size=$TEXT_ENCODER_TP \
     actor_rollout_ref.rollout.n=$ROLLOUT_N \
     actor_rollout_ref.rollout.seed=42 \
     actor_rollout_ref.rollout.agent.num_workers=$((NUM_GPUS / ROLLOUT_TP)) \
--- a/tests/workers/rollout/rollout_vllm/test_diffusion_text_encoder_tp_on_cpu.py
+++ b/tests/workers/rollout/rollout_vllm/test_diffusion_text_encoder_tp_on_cpu.py
@@ -1,0 +1,127 @@
+# Copyright 2026 Bytedance Ltd. and/or its affiliates
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""Exercise the fused diffusion ETP path without starting workers or loading weights."""
+
+import os
+from argparse import Namespace
+from dataclasses import asdict
+from types import SimpleNamespace
+
+import pytest
+from vllm_omni.diffusion.data import DiffusionParallelConfig, OmniDiffusionConfig
+from vllm_omni.engine.arg_utils import OmniEngineArgs
+from vllm_omni.engine.async_omni_engine import AsyncOmniEngine
+
+from verl_omni.workers.config import DiffusionRolloutConfig
+from verl_omni.workers.rollout.vllm_rollout import vllm_omni_async_server as server_module
+from verl_omni.workers.rollout.vllm_rollout import vllm_omni_diffusion_strategy as strategy_module
+
+
+def _prepare(monkeypatch, *, algorithm="diffusion_nft", typed=1, cli=None, parallel=None):
+    monkeypatch.setattr(strategy_module, "import_external_libs", lambda _: None)
+    monkeypatch.setattr(strategy_module.VllmOmniPipelineBase, "get_pipeline_path", lambda **_: "test.Pipeline")
+    monkeypatch.setattr(
+        strategy_module.VllmOmniPipelineBase, "get_class", lambda **_: SimpleNamespace(supports_request_batch=True)
+    )
+    config = DiffusionRolloutConfig(tensor_model_parallel_size=4, text_encoder_tp_size=typed)
+    server = SimpleNamespace(
+        config=config, model_config=SimpleNamespace(architecture="MiniMaxH3Pipeline", algorithm=algorithm)
+    )
+    args = Namespace(tensor_parallel_size=4, text_encoder_tp_size=cli)
+    engine_args = asdict(OmniEngineArgs.from_cli_args(args))
+    if parallel is not None:
+        engine_args["parallel_config"] = parallel
+    strategy_module.DiffusionStrategy(server).prepare_engine_args(engine_args, args)
+    return engine_args
+
+
+@pytest.mark.parametrize("algorithm", ["diffusion_nft", "flow_grpo"])
+@pytest.mark.parametrize("typed,cli,expected", [(1, None, 1), (4, None, 4), (1, 4, 4), (4, 4, 4), (1, 1, 1)])
+def test_text_encoder_tp_reaches_fused_diffusion_config(monkeypatch, algorithm, typed, cli, expected):
+    engine_args = _prepare(monkeypatch, algorithm=algorithm, typed=typed, cli=cli)
+    stages = AsyncOmniEngine._create_default_diffusion_stage_cfg(engine_args)
+    parallel = stages[0]["engine_args"]["parallel_config"]
+    od_config = OmniDiffusionConfig(parallel_config=parallel)
+    assert od_config.parallel_config.tensor_parallel_size == 4
+    assert od_config.parallel_config.text_encoder_tp_size == expected
+
+
+@pytest.mark.parametrize("cli", [0, 2, 8])
+def test_legacy_cli_rejects_invalid_encoder_groups(monkeypatch, cli):
+    with pytest.raises(ValueError, match="text_encoder_tp_size"):
+        _prepare(monkeypatch, cli=cli)
+
+
+def test_conflicting_typed_and_legacy_etp_fails(monkeypatch):
+    with pytest.raises(ValueError, match="Conflicting text_encoder_tp_size"):
+        _prepare(monkeypatch, typed=4, cli=1)
+
+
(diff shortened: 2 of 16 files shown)


verl_omni/workers/rollout/vllm_rollout/vllm_omni_diffusion_strategy.py
```diff
@@ -102,6 +103,34 @@ def worker_extension_cls(self, device_type: str) -> str:
...
+            nested_text_encoder_tp = parallel_config.get("text_encoder_tp_size")
+            if nested_text_encoder_tp is not None:
+                if nested_text_encoder_tp != text_encoder_tp and (
+                    cli_text_encoder_tp is not None or text_encoder_tp != 1
+                ):
+                    raise ValueError("Conflicting text_encoder_tp_size in rollout/engine_kwargs and parallel_config.")
+                text_encoder_tp = nested_text_encoder_tp
+            tp_size = parallel_config.get("tensor_parallel_size", tp_size)
+
+        if text_encoder_tp < 1 or text_encoder_tp not in (1, tp_size):
```
SamitHuang, line 127: `DiffusionRolloutConfig` already rejects anything other than ETP=1 or ETP=TP, and this repeats that. The new docs still say MiniMax-H3 supports ETP 1/2/4/8, which will send people to `TEXT_ENCODER_TP=4` with `ROLLOUT_TP=8` and a startup `ValueError`. Please make the recipe/docs match this check (allowed sizes are 1 or the actual TP), and drop the 1/2/4/8 list unless TP is one of those values.

examples/diffusionnft_trainer/minimax_h3/run_minimax_h3_t2va_lora.sh
```diff
@@ -9,6 +9,7 @@ MODEL_PATH=${MODEL_PATH:-}
 DATA_DIR=${DATA_DIR:-$WORKSPACE/data/vid_prompt/verl_omni}
 NUM_GPUS=${NUM_GPUS:-8}
 ROLLOUT_TP=${ROLLOUT_TP:-2}
+TEXT_ENCODER_TP=${TEXT_ENCODER_TP:-$ROLLOUT_TP}
```
SamitHuang, line 13: This script previously omitted ETP and therefore ran with the typed default of 1. Defaulting `TEXT_ENCODER_TP` to `$ROLLOUT_TP` (same on the FL2VA NFT launcher) changes existing NFT T2VA/FL2VA jobs from an unsharded encoder to full-TP sharding. That may be intended for #563, but it should be called out as a recipe default change, not only as a wiring fix. `TEXT_ENCODER_TP=1` remains the opt-out.

SamitHuang (changes requested):
Restoring `text_encoder_tp_size` in `DiffusionStrategy.prepare_engine_args` is the right place: the pinned CLI accepts the flag, `OmniEngineArgs.from_cli_args` drops it, and `launch_server` never copies the typed field the way it copies TP. Two follow-ups before this lands: docs/recipes still advertise H3 ETP 1/2/4/8 while this path (and `DiffusionRolloutConfig`) only allow 1 or full TP; NFT T2VA/FL2VA now default ETP to `ROLLOUT_TP` where they previously ran ETP=1.

verl_omni/workers/rollout/vllm_rollout/vllm_omni_diffusion_strategy.py
NancyFyong: Fixed in f62f812. The config reference and both MiniMax-H3 READMEs now state that ETP must be 1 or exactly the configured rollout TP; the misleading 1/2/4/8 list is removed. The 98 routing tests, config-doc check, and all pre-commit hooks pass.

examples/diffusionnft_trainer/minimax_h3/run_minimax_h3_t2va_lora.sh
NancyFyong: Fixed in f62f812. The NFT README and both T2VA/FL2VA launchers now explicitly call out that their default changes from the previous effective ETP=1 to full rollout TP, with TEXT_ENCODER_TP=1 documented as the compatibility opt-out. The PR body is updated as well.

tests/workers/rollout/rollout_vllm/test_diffusion_text_encoder_tp_on_cpu.py
zhtmike, line 1: this test can not do any regression protection to vllm-omni side bug. Raise the issue in vllm-omni side, thanks

verl_omni/workers/rollout/vllm_rollout/vllm_omni_diffusion_strategy.py
```diff
@@ -102,6 +103,34 @@ def worker_extension_cls(self, device_type: str) -> str:
...
+            raise ValueError(f"text_encoder_tp_size must be 1 or equal to tensor parallel size ({tp_size}).")
+        engine_args["text_encoder_tp_size"] = text_encoder_tp
+        if parallel_config is not None:
+            parallel_config["text_encoder_tp_size"] = text_encoder_tp
+            engine_args["parallel_config"] = parallel_config
+
```
zhtmike: fix in vllm-omni, and a TODO here to drop it later.

Tag #445

zhtmike (changes requested):
as comment.

SamitHuang:
Re-reviewed `f62f812`.

The previous request-changes items are addressed: recipe/docs now say ETP is 1 or full rollout TP (the 1/2/4/8 list is gone), and NFT T2VA/FL2VA call out the default change to `$ROLLOUT_TP` with `TEXT_ENCODER_TP=1` as the opt-out.

Remaining open item is @zhtmike's: this restore is a pin-compat workaround for `OmniEngineArgs.from_cli_args` dropping CLI-only `text_encoder_tp_size`. File that on vllm-omni, add a TODO here pointing at #445, and keep the CPU test as a verl-omni routing regression (it will not catch an upstream drop of the field).

knlnguyen1802 (commented):
It look like right now we hardcode quite a few of engine core param, we might need a proper way to control these param in some config @SamitHuang  @zhtmike

tests/workers/rollout/rollout_vllm/test_diffusion_text_encoder_tp_on_cpu.py
NancyFyong: Agreed. I filed the upstream bug as vllm-omni#7564. This CPU test is retained only as a verl-omni routing regression for the temporary compatibility shim; it does not claim to protect the vLLM-Omni implementation.

verl_omni/workers/rollout/vllm_rollout/vllm_omni_diffusion_strategy.py
NancyFyong: Addressed in 36da460. The upstream bug is filed as vllm-omni#7564, and the workaround now has a TODO linking that issue and the downstream patch tracker verl-omni#445 so it can be removed after the pin contains the upstream fix.

zhtmike:
> It look like right now we hardcode quite a few of engine core param, we might need a proper way to control these param in some config @SamitHuang @zhtmike

all these core params should be passed from engine_args, but currently vllm-omni is not. we need to fix them tougher in one PR in vllm-omni.

NancyFyong:
> > It look like right now we hardcode quite a few of engine core param, we might need a proper way to control these param in some config @SamitHuang @zhtmike
> 
> all these core params should be passed from engine_args, but currently vllm-omni is not. we need to fix them tougher in one PR in vllm-omni.

agree

zhtmike (approved):

knlnguyen1802 (approved):

SamitHuang (approved):
LGTM

Issue #396, opened 2026-08-15
[RFC] Support MiniMax H3 FL2VA with DiffusionNFT and FlowGRPO


NancyFyong:
## Summary

Add first-frame, last-frame, and first+last-frame conditioning for the MiniMax H3
FL2VA checkpoint under two training algorithms:

1. **DiffusionNFT** trains from the final clean generated video/audio target rows.
   Reference-frame condition rows remain fixed and are excluded from forward
   noising and loss.
2. **FlowGRPO** records separate stochastic video/audio reverse-SDE transitions.
   Reference-frame condition rows remain fixed and are excluded from trajectories
   and policy log-probabilities.

Both algorithms share the same official vLLM-Omni FL2VA request contract, packed
sequence semantics, token-ID-native multimodal prompt path, FSDP2 LoRA Actor,
and Actor-to-rollout weight synchronization.

## Motivation

MiniMax H3 jointly generates video and stereo audio, while its FL2VA checkpoint
also conditions generation on one or two keyframes. This is not ordinary I2V
conditioning: vLLM-Omni packs condition-video rows, generated-video rows, audio
rows, text embeddings, multimodal token tags, three-dimensional positions, and
per-row timesteps into one DiT sequence.

A correct RL integration must preserve all of those semantics during Actor
replay. Treating FL2VA as T2VA with extra images is incorrect because it can:

- update condition rows that should remain pinned to reference anchors;
- apply ordinary forward noise to reference rows;
- assign data timesteps instead of conditioning timesteps;
- lose vision-token modality tags or keyframe temporal positions;
- include condition rows in policy trajectories and log-probabilities;
- decode and re-tokenize prompts, changing multimodal presentation IDs;
- replay a different video/audio sigma schedule from rollout.

Supporting both DiffusionNFT and FlowGRPO also validates that one FL2VA transport
contract can serve a direct-preference objective and a policy-gradient objective
without conflating their different training targets.

## Relationship to existing work and non-duplication

- [RFC #347](https://github.com/verl-project/verl-omni/issues/347) proposes
  MiniMax H3 **T2VA FlowGRPO** on GPU and NPU and explicitly lists first/last-frame
  conditioning as a non-goal. This RFC is GPU-first and specifically owns the
  FL2VA conditioning/replay contract for both DiffusionNFT and FlowGRPO.
- Draft [PR #368](https://github.com/verl-project/verl-omni/pull/368) implements
[...]

PR #592, opened 2026-09-13
[diffusion, rollout, fsdp, perf, tests, doc] feat: add shared packed MiniMax H3 actor forward

NancyFyong:
### What does this PR do?

Add default packed Actor forward for MiniMax-H3 DiffusionNFT and FlowGRPO. The existing NFT training adapter installs sample-isolated packed execution on the standard Diffusers `AutoModel`; FlowGRPO reuses those helpers. No new production file, base-class hook, model-config class, or enable switch.

| Path | Main benefit |
| --- | --- |
| DiffusionNFT | Replace per-sample transformer calls with one packed call per microbatch/prediction: **8 → 1 at microbatch 8**. An earlier matched tiny trainer A/B observed **1.63× faster actor updates**; provenance and limits are below. |
| FlowGRPO | Allow different prompt lengths/reference layouts in one microbatch instead of requiring a shared layout. It already had dense batching, so the NFT speedup does not apply to it. |
| Both | Reuse sample-isolated DiT/text-refiner attention while preserving per-sample positions, timesteps, objectives and checkpoint names. |

This is model-side packing, **not** dynamic token-budget batching, DP workload balancing or rollout batching. Related context: #396; this PR does not close that broader RFC.

**Also included: two diffusion-rollout determinism fixes.** These are not H3 packing changes, but the packed GPU numerical checks now depend on them, so they ship together. Comparing packed against serial execution requires deterministic matmul, and enabling determinism was previously unusable on this path:

| Problem | Cause | Fix |
| --- | --- | --- |
| Rollout produced NaN latents, so every Actor update reported a non-finite `grad_norm` | `torch.use_deterministic_algorithms` also enables `fill_uninitialized_memory`, which poisons the `torch.empty` scratch buffers that hold the packed per-step latent trajectory | Disable that fill after enabling determinism |
| Worker aborted with `CUDA error: misaligned address` | The same fill corrupts vLLM custom all-reduce IPC buffers. vLLM disables that kernel itself under `VLLM_BATCH_INVARIANT`, but the vLLM-Omni diffusion worker builds its parallel state directly, so the guard never ran | Disable custom all-reduce under determinism, falling back to NCCL |

The rollout worker subprocess also restored no determinism at all, so it silently ran non-deterministically even when configured. Both are routed through one shared helper used by the server, the worker and the GPU test. If maintainers prefer, this can be split into its own PR; the packed GPU check would then need that PR first.

### Checklist Before Starting

- [x] Search for similar PRs: [MiniMax packed](https://github.com/verl-project/verl-omni/pulls?q=is%3Apr+is%3Aopen+MiniMax+packed), [batch forward](https://github.com/verl-project/verl-omni/pulls?q=is%3Apr+is%3Aopen+%22batch+forward%22), and `396 in:body`. No duplicate Actor-packing PR found; this updates the existing #592.

#### Correctness

Revalidated after test consolidation at `0994dd5`.

**CPU: 380 passed.** Coverage includes two Ref2VA prompt groups with `rollout.n=2`: visual/audio reference-row counts are `[4, 4, 8, 8]` / `[0, 0, 4, 4]`, so layouts match within each prompt group and differ only across groups. The former dense shared-layout path rejects the combined Actor micro-batch, while packed FlowGRPO executes it in one model call and matches four per-sample serial replays. The suite also covers outputs, actual losses/LoRA gradients, cross-sample isolation, default-config loading, checkpoint compatibility, exact-name FSDP wrapping, nested replay fields and mixed step groups.

```bash
export TORCH_COMPILE_DISABLE=1 TORCHINDUCTOR_DISABLE=1
python -m pytest $(find tests \( \
  -name '*minimax*on_cpu.py' -o -name '*diffusers*on_cpu.py' -o \
  -name '*diffusion_attention*on_cpu.py' -o -name 'test_diffusion_config_on_cpu.py' \
\) -print) tests/workers/rollout/rollout_vllm/test_vllm_omni_strategy_on_cpu.py -q
```

**GPU numerical: passed on both ranks, including the consolidated entry after relocation.** The packed regression is consolidated into the existing LoRA sync test, rather than a separate GPU script:

```bash
NUM_GPUS=2 bash tests/special_e2e/run_minimax_h3_lora_sync_tp2.sh \
  --check-packed-forward --attn-backend _flash_3_varlen_hub
```

This checks LoRA actor-to-rollout sync and encoder groups, then NFT/FlowGRPO losses and LoRA gradients over two updates, without/with recipe-style FSDP2 and checkpointing. FlowGRPO covers T2VA/FL2VA/Ref2VA; NFT includes old-adapter/reference predictions and sample isolation. Compatible FA3 v1 kernels were provisioned through `LOCAL_KERNELS`. Without `--check-packed-forward`, the original sync checks retain their existing scope and do not require FA3.

**Default-entry tiny trainers after the AutoModel refactor: NFT and FlowGRPO each passed 2/2 steps, exit 0.** Two H20 ranks, ordinary `DiffusionModelConfig`, microbatch 8, one model call per prediction, finite outputs/nonzero finite gradients; NFT copied the old policy at step 2 and FlowGRPO exercised varying prompt lengths. These runs validate the current AutoModel-loaded, adapter-installed path; they do not constitute a new timing run.

**Determinism fixes: validated separately.** Tiny FlowGRPO and NFT trainers each completed 2/2 steps, exit 0, with all three `full_determinism` flags enabled: zero NaN in every recorded Actor forward, finite grad norms on both ranks, and `rollout_corr/k3_kl` 1.59e-6 matching the non-deterministic baseline. Before the fix the same configuration failed with 1197056 NaN `all_latents` and a non-finite `grad_norm`. The cause was isolated by a 2x2 two-GPU all-reduce probe: `deterministic=1/fill=0/custom_ar=1` passes and `deterministic=0/fill=1/custom_ar=1` passes, while `deterministic=1/fill=1/custom_ar=1` fails with a misaligned address. A rollout-only arm reproduced the NaN with the Actor left non-deterministic, placing the fault entirely in the rollout path. The consolidated TP=2 GPU check now passes with determinism enabled, where it previously aborted with SIGABRT; reported gradient errors are byte-identical to the pre-change run (0.003283 / 0.003766 / 0.003219). Ordinary sync runs do not enable determinism and are unaffected.

**Why an earlier H3 matmul override was removed.** A prior revision selected cuBLASLt and disabled BF16 reduced-precision reductions in the H3 actor/rollout. Measured on a real-dimension rank-64 LoRA GEMM comparing per-sample against concatenated shapes, determinism alone already yields exact parity (`max_abs=0`), and the override changed only the reported backend/flags without changing any result. It was therefore deleted, together with the sanity-gate device-API whitelist entry and the version-dependent split-K API dependency it required.

**Pre-commit: all 12 hooks passed.** All tracked files supplied via `git ls-files -z` → `pre-commit run --files` because this host's Git lacks `--deduplicate`. Generated configs have no net diff. These are local results, not a claim that full GitHub CI ran.

#### How much faster? Separate NFT and FlowGRPO baselines

**NFT: complete earlier tiny actor-engine microbatch sweep (serial FA3 versus packed FA3):**

| Microbatch | Serial update | Packed update | Serial / packed | Peak allocated MiB, serial / packed |
| --- | ---: | ---: | ---: | ---: |
| 1 | 2.445 s | 2.621 s | 0.93× | 78.1 / 78.1 |
| 2 | 2.027 s | 1.209 s | 1.68× | 89.4 / 91.3 |
| 4 | 1.815 s | 0.750 s | 2.42× | 113.9 / 118.9 |
| 8 | 1.713 s | 0.406 s | 4.22× | 162.9 / 170.6 |

These are **actor-only measurements on the earlier September 12 implementation**, before the default-path follow-ups, not a fresh benchmark of the current head. Fixed 16 global samples and four timesteps per sample; two H20 ranks, the actual NFT engine/loss accumulation, LoRA, FSDP2 and gradient checkpointing. Six timed updates per configuration with all 48 samples retained. Timing includes preparation, old/default/reference predictions, loss/backward and optimizer work, but excludes Ray, rollout/reward, CPU parameter/optimizer offload, old-policy copies and weight sync. Peak allocation includes both resident tiny engines and is comparable only within this sweep.

The synthetic layout matches tiny rollout video/audio row counts but has different RoPE coordinates; this is a same-token-workload microbenchmark, not exact rollout replay. **The 4.22× at B=8 is not a whole-training speedup, nor a real-H3 result.** B=1 is slower. The separate full-trainer measurement below includes rollout and synchronization.

**NFT: earlier tiny full-trainer A/B, serial FA3 versus packed FA3, microbatch 8:**

| Metric | Serial | Packed | Observed speedup |
| --- | ---: | ---: | ---: |
| Actor update | 2.6643 s | 1.6313 s | **1.63×** (38.8% less time) |
| Whole training step | 8.4702 s | 6.5619 s | **1.29×** (22.5% less time) |

Both completed 6/6 steps, including weight sync and old-policy updates. Values are medians over steps 2–5; all per-step samples were retained. These measurements were taken on the earlier implementation on September 12, **before the default-path/consolidation follow-ups**, not remeasured at the current head. Unchanged weight-sync time also improved by 21%, so shared-host contention and the short window confound the whole-step delta: **do not attribute all of the 1.29× result to packing or extrapolate to real H3.**

**FlowGRPO: latest matched equal-layout tiny actor-engine comparison from `5f83dec`, against existing dense batching:**

| Microbatch | Dense actor-engine update | Packed actor-engine update | Packed overhead |
| --- | ---: | ---: | ---: |
| 1 | 714.70 ms | 727.43 ms | +1.78% |
| 2 | 367.49 ms | 373.33 ms | +1.59% |
| 4 | 187.74 ms | 192.92 ms | +2.76% |
| 8 | 97.42 ms | 103.55 ms | +6.29% |

Prepared-input model forward/backward alone at B=8 was **25.36 ms dense / 25.11 ms packed**; that separate diagnostic excludes preparation/replay/optimizer work.

The engine retains **~6.3% overhead**; model-only cost is comparable within shared-host noise. Both paths already use FA3 varlen for equal-length samples. This PR does **not** demonstrate a same-layout FlowGRPO speedup or a variable-layout throughput gain; its FlowGRPO benefit is broader batching support.

FlowGRPO controls: same tiny checkpoint/inputs/FA3/LoRA/FSDP2/checkpointing/loss accumulation; two ranks, 16 global samples, three transitions. All 48 engine-sweep and 96 model-only samples across B=1/2/4/8 were retained, with alternated execution order. Engine timing excludes Ray, rollout, weight sync and CPU offload; model-only timing additionally excludes preparation/replay/optimizer work. These absolute times are not comparable to the NFT full-trainer rows above.

The diagnostic trainer/profiling harness remains local and uncommitted; the numerical regression is included in the PR. Production-weight throughput, memory limits and convergence remain unverified.

### API and Usage Example

Existing H3 FSDP launchers keep the standard `AutoModel` loader. The H3 adapter installs packed execution automatically on the loaded instance; no custom `_target_` or packing flag is needed:

```bash
actor_rollout_ref.model.attn_backend=_flash_3_varlen_hub \
actor_rollout_ref.rollout.rollout_attn_backend=FLASH_ATTN_3_HUB \
actor_rollout_ref.actor.ppo_micro_batch_size_per_gpu=2 \
actor_rollout_ref.rollout.log_prob_micro_batch_size_per_gpu=2 \
actor_rollout_ref.ref.log_prob_micro_batch_size_per_gpu=2
```

FA3 consumes explicit per-sample sequence boundaries; `native` supplies padded batched SDPA with masks, paired with rollout `TORCH_SDPA`. `torch_varlen` is a manual numerical-test option, not a new trainer backend. Missing/unsupported kernels fail closed. Existing fixed-microbatch divisibility rules still apply; this does not automatically resize batches by token budget.

### Design & Code Changes

- **No new production module:** packed execution lives in the existing NFT Diffusers training adapter, and FlowGRPO imports its shared helpers.
- **Standard loading remains unchanged:** Diffusers `AutoModel` loads the original `MiniMaxH3Transformer3DModel`. On the first Actor forward, the H3 adapter installs instance-local packed forward/refiner methods and attention processors without replacing parameters or module classes. LoRA, FSDP2, checkpoint names, fp32 islands and the generic engine/base classes remain unchanged.
- **NFT:** pack sample-local layouts, split outputs and remove condition outputs before the unchanged target-only loss.
- **FlowGRPO:** stage metadata once; leave unrelated nested fields untouched; gather target rows, batch dual-scheduler replay by original step and restore sample order. Reference rows are not scored.
- **Determinism:** one shared rollout helper enables verl determinism and then undoes the two side effects vLLM cannot tolerate; the GPU test calls it before any CUDA/distributed work so the cuBLAS workspace setting still applies. Only the optional packed checks enable it.
- **Tests:** keep CPU fixtures and fold GPU numerical checks into `minimax_h3_lora_sync_tp2.py` behind an explicit option, preserving coverage without another standalone script.
- **Scope/limits:** generic config/YAML, attention validation and rollout strategies have no net change. Old serial/dense execution is test-only. NFT requires shared target geometry/FL2VA anchors; FlowGRPO requires equal target video/audio row counts. No mixed-checkpoint/task training promise. Sequence/context parallelism is rejected; VeOmni/NPU packed execution is not accepted here.

- [x] Apply [pre-commit checks](https://github.com/verl-project/verl-omni/blob/main/CONTRIBUTING.md#code-linting-and-formatting) to all tracked files.
- [x] Add CPU tests discoverable by the existing `*_on_cpu.py` workflow. FA3/FSDP2 checks remain manual because compatible GPU kernel provisioning is required; they are not added to offline GPU smoke CI here.
- [ ] Human submitter has reviewed every changed line and independently rerun relevant tests.
- [ ] Validate production weights, throughput, memory limits and convergence. **Not done:** all timing/correctness evidence above is tiny-model or synthetic-replay. Real-weight throughput, memory headroom and convergence are unverified.


verl_omni/pipelines/minimax_h3_diffusion_nft/diffusers_training_adapter.py
zhtmike, line 1: if possible, can we do it in veomni side.

zhtmike: it is a nice feature, but putting it under training engine will be more suitable, and may consider generalize the packed attention to other models.

NancyFyong: I’ll introduce it in weekly meeting, we can discuss it thoroughly.

Issue #266, opened 2026-07-15
[RFC] Qwen3-Omni Offline DPO Training Framework on FSDP


wtomin:
### Feature request

Training support for qwen3-omni using offline dpo algorithm.

### Motivation

Qwen3-Omni thinker-stage training in verl-omni today goes through upstream `verl.trainer.main_ppo` and existing integration paths (including `external_lib` hooks). Issue **#232** proposes formal omni support via `main_omni`, `OmniModelBase`, and native `pipelines/qwen3_omni/` adapters.

What is **missing** is the **training framework** for offline direct-preference optimization:

1. **FSDP** as the training backend (not VeOmni-only).
2. **Offline DPO** orchestration (no rollout, no reward engine).
3. **LoRA** with **ref-in-actor** (reference policy = base weights inside the actor worker).
4. **RFC \[RFC\] Formal Omni-Model Support in verl-omni #232** entrypoint and adapter layout (`main_omni`, `OmniModelBase`, native pipelines).

This RFC scopes **framework implementation only**. Any concrete dataset (parquet schema, multimodal preprocessing, collate) is assumed to be provided externally via verl's existing `data.custom_cls` hook and is not designed here.

**Backward compatibility:** Existing Qwen3-Omni training paths (GSPO, patch-based `external_lib` recipes, upstream `main_ppo`) must continue to work unchanged. The new framework is an **additive** code path, not a breaking migration.

The reference pattern is the diffusion offline DPO stack (`main_diffusion` \+ `DirectPreferenceRayTrainer` \+ FSDP engine).

## 2\. Goals

| #  | Goal                                                                                                                                                                                             |
| -- | ------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------ |
| G1 | **Unified entrypoint** — python -m verl\_omni.trainer.main\_omni supports offline DPO via algorithm.trainer\_type=direct\_preference \+ algorithm.sample\_source=offline.                        |
| G2 | **RFC-compliant model adapter** — provide OmniModelBase / Qwen3OmniThinkerAdapter as the preferred integration path for new offline DPO runs.                                                    |
| G3 | **FSDP training engine** — register model\_type=omni\_model on verl's FSDP backend with multimodal forward + token log-prob for DPO.                                                             |
| G4 | **Offline actor-only workers** — Role.Actor only; no rollout, reward, or vLLM-Omni servers.                                                                                                      |
[...]

PR #600, opened 2026-09-15
[tests] fix: use FP32 LoRA in offline DPO smoke recipe

0z5a:
### What does this PR do?

Fix the single-GPU Qwen3-Omni offline DPO smoke recipe failing with BF16 adapter parameters and FP32 FSDP2 gradients. Default `LORA_DTYPE` to `float32` and forward it to the existing model option. Related to #266; this is validation/fix of the existing framework, not a new DPO implementation.

### Checklist Before Starting

- [x] Checked [existing DPO PRs](https://github.com/verl-project/verl-omni/pulls?q=is%3Apr+is%3Aopen+DPO); no equivalent dtype fix found.
- [x] Title follows the project format.

### Test

One NVIDIA L20:50 real multimodal DPO/FSDP2 steps, actor-only reference computation. All metrics finite; gradient norms3.472–3.918. All24 LoRA tensors changed in each checkpoint interval10→20→30→40→50; all finite. Image/video/audio validation ran each step. Runtime298.41s including startup/cleanup. Prior two-step run also passed. This does not validate post-training judge generation, model quality, or multiple GPUs.

### API and Usage Example

`LORA_DTYPE=float32 NUM_GPUS=1 TOTAL_TRAINING_STEPS=50 bash tests/special_e2e/run_qwen3_omni_multimodal_offline_mllm_dpo_lora_smoke.sh trainer.total_epochs=100 trainer.save_freq=10`

See the new reproduction note for model/data paths and environment.

### Design & Code Changes

Only the smoke recipe and its reproduction note change. The existing `actor_rollout_ref.model.lora_dtype` handles parameter casting; no worker implementation changes are needed.

- [x] Shell syntax and diff whitespace checked.
- [x] Added reproduction and GPU results documentation.
- [ ] Full pre-commit suite not run for this shell/default change.
- [ ] GPU E2E run manually; no dedicated hardware CI job added.


wtomin:
@0z5a  Sorry for my late reply. I checked the smoke test code. The omni models' smoke tests are always running in multiple cards, so I think it's not necessary to change the smoke test script.

But I can replicate the error you mentioned. I try to trace the root cause in our repo:
```python
class DiffusersFSDPEngine(LoRAAdapterMixin, BaseEngine, ABC):
    def _build_fsdp_module(self, module):
        # TODO(ziheng): need to improve
        from torch.distributed.fsdp import CPUOffload, MixedPrecision
        from verl.utils.torch_dtypes import PrecisionType

        mixed_precision_config = self.engine_config.mixed_precision
        if mixed_precision_config is not None:
            param_dtype = PrecisionType.to_dtype(mixed_precision_config.get("param_dtype", "bf16"))
            reduce_dtype = PrecisionType.to_dtype(mixed_precision_config.get("reduce_dtype", "fp32"))
            buffer_dtype = PrecisionType.to_dtype(mixed_precision_config.get("buffer_dtype", "fp32"))
        else:
            param_dtype = torch.bfloat16
            reduce_dtype = torch.float32
            buffer_dtype = torch.float32
```

I think the `reduce_dtype` is the parameter dtype after `reduce-scatter` operations in fsdp/fsdp2. Therefore, the param dtype in cards>1 is always fp32, while in the single-card case, it is just a copy operation, and the param dtype stays to be `bf16`.

@cr-gao @zhtmike @NancyFyong  I know single-card training is rare, but shall we add a guard in the `DiffusersFSDPEngine`  mixed precision configuration: when the number of cards is 1, force the param_dtype to be the same as the gradient dtype?

zhtmike:
This PR is actually good. Better to keep LORA_DTYPE = fp32 when using PyTorch native optimizer (except torchao, etc)

tests/special_e2e/qwen3_omni_dpo_single_gpu.md
```diff
@@ -0,0 +1,22 @@
...
+
+Validated with the existing tiny checkpoint builder and image/video/audio datasets,
+FSDP2 ref-in-actor, one NVIDIA L20, torch2.13.0+cu130, Transformers5.14.1,
+and verl revision fefb080262e1c015a0ea05f958822a6a512dc795.
+All50 training steps completed; all logged metrics were finite and gradient norms
+were nonzero (3.472–3.918). Image/video/audio validation ran each step.
+Checkpoints at steps10,20,30,40,50 contained24 finite LoRA tensors;
+all24 changed between every adjacent pair. Runtime298.41s including initialization
+and cleanup. This is a tiny-model execution test, not a convergence or model-quality
+claim. The recipe's commented post-training judge generation was not exercised.
```
zhtmike: no need this.

0z5a: removed.

zhtmike: still there

zhtmike: .. still there.

0z5a: removed.

wtomin:
@0z5a Could you take a look at the CI error?

0z5a:
Fixed the two-GPU CI dtype mismatch in 103a44b: FP32 LoRA is now the single-GPU default, while multi-GPU runs retain the original dtype. I cannot reapply `ready-for-ci` with my permissions; could you add it to rerun the GPU suite? I am also validating the two-GPU E2E independently.

PR #606, opened 2026-09-16
[worker, fsdp, tests] fix: forward layered_summon and fsdp_layer_prefixes 

wtomin:
## Summary

Fix missing kwargs in the separate-async / NCCL weight-sync path so rollout settings actually reach LoRA collection:

- **`ActorRolloutRefWorker.update_weights`**: forward `rollout.layered_summon` into `get_per_tensor_param()` for both adapter-only and full-weight checkpoint sends (previously always defaulted to `False`).
- **`OmniFSDPEngine.get_per_tensor_param`**: forward `model_config.fsdp_layer_prefixes` as `layer_prefixes` to `collect_lora_params()` (matching `DiffusersFSDPEngine`; without this, omni models silently fell back to the DiT default `transformer_blocks.` prefix).

This is not a duplicate of an existing open PR: searched for `layered_summon`, `get_per_tensor_param`, and `fsdp_layer_prefixes` on `verl-project/verl-omni` (no matching open PR found).

## What changed

| File | Change |
| --- | --- |
| `verl_omni/workers/engine_workers.py` | Pass `layered_summon=self.layered_summon` in NCCL `update_weights` paths |
| `verl_omni/workers/engine/fsdp/omni_impl.py` | Pass `layer_prefixes=self.model_config.fsdp_layer_prefixes` to `collect_lora_params` |
| `verl_omni/utils/fsdp_utils.py` | More specific error when `collect_lora_params` returns 0 tensors |
| `tests/workers/test_omni_fsdp_engine_on_cpu.py` | Assert `layered_summon` / `layer_prefixes` forwarding in omni engine |
| `tests/workers/test_weight_sync_knobs_on_cpu.py` | Assert `update_weights(nccl)` forwards `layered_summon` for LoRA and full-weight sends |

## Test plan

- [ ] `pytest tests/workers/test_omni_fsdp_engine_on_cpu.py::TestAdapterNameForwarding -q`
- [ ] `pytest tests/workers/test_weight_sync_knobs_on_cpu.py -q`
- [ ] Re-run the CI job that failed on separate-async weight sync (if applicable)

```bash
pytest tests/workers/test_omni_fsdp_engine_on_cpu.py tests/workers/test_weight_sync_knobs_on_cpu.py -q

### CODE REVIEW

**Comment by wtomin** [comment]
@knlnguyen1802 This PR simply fixes some arguments passing issues, which didn't solve the CI errors. Suggestions are welcome.

File: verl_omni/workers/engine/fsdp/omni_impl.py [hunk 1]
```diff
@@ -86,6 +86,7 @@ def get_per_tensor_param(self, layered_summon=False, base_sync_done=False, **kwa
                     layered_summon=layered_summon,
                     base_sync_done=base_sync_done,
                     adapter_name=adapter_name,
+                    layer_prefixes=self.model_config.fsdp_layer_prefixes,
```
knlnguyen1802, line 89: I don't think `layer_prefixes` does anything on this path. `collect_lora_params` only threads `layer_prefixes` into the diffusers walker (the `is_diffusers` branch in `fsdp_utils.py`); the omni engine calls it without `is_diffusers`, so the non-diffusers branch uses verl's `layered_summon_lora_params`, which takes no prefix argument and is invoked as `layered_summon_fn(module)`. The only other consumer of `layer_prefixes` is the `RuntimeError` detail string when collection returns 0 params.

So the "omni models silently fell back to the DiT default `transformer_blocks.` prefix" described in the PR summary isn't quite right: nothing on the omni path ever consumed the prefix, and there is no prefix-based fallback to fix. The net effect of this line is the error message text only (and with the omni default `fsdp_layer_prefixes=[]`, that message would read `with prefixes=[]`).

Suggest dropping this line plus the `layer_prefixes` captures/asserts in the omni tests, so the PR stays purely about the `layered_summon` forwarding (which is a real fix, tested). If prefix parity with `DiffusersFSDPEngine` is intended, note that the diffusers twin only works because it also passes `is_diffusers=True` — making it real for omni would mean threading prefixes through the non-diffusers summon in `fsdp_utils.py` as well.

wtomin:
Close this PR because it's not a critical issue.

PR #609, opened 2026-09-16
[fsdp, diffusion] fix: sync a frozen LoRA adapter's weights during DiffusionNFT weight sync

### LINKED ISSUE #608: DiffusionNFT dual-adapter LoRA weight sync fails: collect_lora_params collects 0 parameters for frozen adapter
### System Info

Modal, 4x/2x/1x A10G, CUDA 12.8, Python 3.12.1, torch 2.11.0+cu130, vllm 0.28.0, vllm-omni 0.28.0rc2.dev68+gded893462, verl @ fefb0802 (pinned), verl-omni @ main (branch feat/boogu-image-diffusion-nft-training-adapter, includes PR #568).

### Information

- [x] The official example scripts

### Tasks

- [x] An officially supported task in the `examples` folder (such as QwenImage, ...)

### Reproduction

Run any DiffusionNFT smoke test that trains two LoRA adapters (`policy_state_adapters=["default","old"]`), e.g. the unmodified, already-merged:

```
bash tests/special_e2e/run_diffusionnft_boogu_image.sh
```

It fails at the very first weight sync:

```
RuntimeError: collect_lora_params collected 0 parameters with prefixes=['double_stream_layers.', 'single_stream_layers.', 'context_refiner.', 'noise_refiner.', 'ref_image_refiner.']. Check ``fsdp_layer_prefixes`` in the model config matches the model's FSDP layer naming (e.g. ``['transformer_blocks.']`` for DiT models).
```

This reproduces regardless of GPU count (1/2/4), `reward.reward_model.enable`, `algorithm.old_policy_update_interval` (1 or 2), and batch size. It also reproduces on the exact config that a prior PR comment reported as a passing 2-step/100-step run (PR #568), so it looks like a real regression or a latent bug that only shows up under certain conditions, not something introduced by that PR's changes.

I instrumented `verl_omni/utils/fsdp_utils.py`'s `collect_lora_params` to dump the live FSDP-wrapped module tree at the point of failure. With two adapters (`default` trainable, `old` frozen for rollout), `get_fsdp_wrap_policy`'s LoRA lambda policy (`verl/utils/fsdp_utils.py`) only wraps leaf modules with `weight.requires_grad=True`, so only the trainable `default` adapter's LoRA leaves get their own nested FSDP unit:

```
feed_forward.linear_3.lora_B.default  -> FullyShardedDataParallel (fsdp_version=1)
feed_forward.linear_3.lora_B.old      -> Linear                   (fsdp_version=0)
```

The frozen `old` adapter's LoRA weights are never individually wrapped. `_layered_summon_lora_params_diffusers` (in `verl_omni/utils/fsdp_utils.py`) then tries to summon the block containing them and extract `old`'s state via `get_peft_model_state_dict(..., adapter_name="old")`, but the aggregate result comes back empty across every prefix-matched block, and the caller raises `RuntimeError: collect_lora_params collected 0 parameters`.

### Expected behavior

`collect_lora_params` should be able to sync a frozen (non-trainable) LoRA adapter's weights to the rollout engine, since DiffusionNFT-style trainers (and any dual-adapter setup with a frozen snapshot adapter) rely on this. The wrap policy / summon logic needs to handle mixed trainable+frozen adapters within the same FSDP-wrapped block.

OnePunchMonk:
## Summary

Fixes #608. DiffusionNFT trains two LoRA adapters: `default` (trainable) and `old` (a frozen rollout snapshot, synced periodically per `algorithm.old_policy_update_interval`). Syncing the frozen `old` adapter's weights to the rollout engine fails every time with:

```
RuntimeError: collect_lora_params collected 0 parameters with prefixes=[...]
```

I bisected this by instrumenting `_layered_summon_lora_params_diffusers` to dump the FSDP-summoned state dict at the point of failure: `submodule.state_dict()` clearly contains real tensors for both adapters (e.g. `attn.to_q.lora_A.old.weight`), but `get_peft_model_state_dict(peft_model, state_dict=raw_state, adapter_name="old")` comes back with 0 keys. This reproduces on every DiffusionNFT run that uses two adapters, independent of GPU count, reward settings, or batch size (see #608 for the full repro and debug output).

## Fix

When `get_peft_model_state_dict` returns nothing for a given adapter despite the raw state clearly containing that adapter's `lora_` tensors, extract them directly (stripping the adapter-name suffix to match `get_peft_model_state_dict`'s normal key convention) instead of dropping the sync silently.

## Why this isn't a duplicate

Checked open issues/PRs for `collect_lora_params`, "frozen adapter", and FSDP LoRA sync -- nothing existing. Opened #608 first to document the repro before this fix.

## AI assistance disclosure

This PR was AI-assisted: **Claude** (Claude Code) diagnosed the root cause (by instrumenting the failing code path and inspecting the live FSDP-wrapped module tree on GPU), implemented the fix, and verified it. A human submitter (onepunchmonk) has reviewed the change.

## Test plan

- Added `tests/utils/test_fsdp_utils_on_cpu.py` covering the new extraction helper.
- `ruff check` / `ruff format --check` pass on changed files.
- Verified live on a Modal A10G: reran the exact previously-failing DiffusionNFT Boogu-Image smoke config (`tests/special_e2e/run_diffusionnft_boogu_image.sh`, 2 steps) -- fails with the RuntimeError above on `main`, completes both training steps successfully with this fix applied.


OnePunchMonk:
## Post-fix verification

Reran the exact same config on Modal (A10G) before and after the fix, unmodified `tests/special_e2e/run_diffusionnft_boogu_image.sh` with `NUM_GPUS=1`, `TOTAL_TRAIN_STEPS=2`.

**Before (main, ap-XKv58ZOavUw4MyZTSPBiTe):**

```
=== DEBUG _layered_summon_lora_params_diffusers: adapter_name='old' ===
    matched name='_fsdp_wrapped_module.double_stream_layers.0' type=FullyShardedDataParallel fsdp_version=1
      raw_state has 141 keys, 96 lora_ keys: [..., 'img_instruct_attn.to_out.0.lora_A.old.weight', ...]
      sub_lora_params has 0 keys: []
    ...
=== DEBUG END: matched_any=True total_lora_params=0 ===

RuntimeError: collect_lora_params collected 0 parameters with prefixes=['double_stream_layers.', 'single_stream_layers.', 'context_refiner.', 'noise_refiner.', 'ref_image_refiner.']
```

`raw_state` clearly holds the `old` adapter's tensors; `get_peft_model_state_dict` just returns nothing for it.

**After (this branch, ap-gxNn2M5qIrSqMiztMwAkwa):**

```
Training Progress:  50%|█████     | 1/2 [00:35<00:35, 35.48s/it]
Training Progress: 100%|██████████| 2/2 [00:46<00:00, 21.02s/it]
```

Both steps complete cleanly, no exception, weight sync to the rollout server succeeds for both the `default` and `old` adapters.

zhtmike:
@AndyZhou952 , pls take a look

zhtmike:
@knlnguyen1802 if you are interested to take a look the lora refactor..

AndyZhou952:
Seems a tentative fallback has been implemented and addressed in #607

wtomin:
@OnePunchMonk  Please close this PR if you confirm that the issues has been solved. Thanks!

OnePunchMonk:
Rebased onto main to pick up #607, which landed after this PR was opened and touched the same file.

#607 added `_lora_params_by_name`/`_lora_checkpoint_key` as a general named-parameters fallback for cases where `get_peft_model_state_dict` misses an adapter's tensors — exactly the class of bug this PR fixes, just in the non-layered/base-sync paths rather than `_layered_summon_lora_params_diffusers` (the layered diffusers path DiffusionNFT actually hits). Rather than keep my own duplicate `_extract_adapter_lora_state_dict` helper, I dropped it and wired the layered path through the existing `_lora_params_by_name` instead.

Note on test evidence: this dev environment can no longer import `verl_omni` at all for CPU tests (`vllm_omni` is a hard import in `verl_omni/pipelines/bagel_flow_grpo/vllm_omni_rollout_adapter.py` via `verl_omni/__init__.py`, and the GPU-only `vllm-omni` wheel isn't installed here) — that's pre-existing and unrelated to this change. I verified the fallback logic with a standalone repro of `_lora_params_by_name` against a two-adapter PEFT model (confirms it recovers the frozen adapter's tensors when the PEFT lookup is empty) and added `test_layered_diffusers_falls_back_for_frozen_non_default_adapter` to `tests/utils/test_fsdp_utils_on_cpu.py`, which should run fine in CI. Will confirm once the CPU-tests job runs.

Issue #285, opened 2026-07-20
[RFC] Save LoRA-only weights during training for inference usage


guozhihao-224:
# RFC: Save LoRA-only weights during training for inference usage

- **Author**: @guozhihao-224
- **Parent roadmap**: verl-project/verl-omni#97 (Q3 Roadmap → Easy-to-Use)
- **Status**: Draft
- **Scope**: `verl_omni` FSDP diffusion path; VeOmni and multi-adapter export are deferred.

## 1. Summary

This RFC adds an adapter-only checkpoint path for LoRA diffusion training. At configured save points, the FSDP diffusion engine gathers the active `default` adapter and publishes a standard PEFT directory:

```text
global_step_<N>/lora_adapter/
  adapter_config.json
  adapter_model.safetensors
  lora_train_meta.json
```

The adapter cadence is independent from full-checkpoint cadence. Full checkpoints, resume behavior, and online rollout weight sync remain unchanged.

1. Reuse `collect_lora_params` as the only FSDP-aware gather path.
2. Produce PEFT-canonical keys in the collector rather than normalizing them after gather.
3. Publish locally through a rank-0 temporary directory and an atomic rename.
4. Treat FSDP collective failures as process-group-fatal, while synchronizing post-gather validation and I/O failures.
5. Require content identity checks before reusing an already-published artifact for the same step.

### Motivation

LoRA training updates only a small set of adapter parameters, but the current trainer checkpoint contains the full actor state, optimizer state, extra state, and dataloader state. This makes frequent checkpoints expensive and leaves inference users with a sharded training artifact instead of a standard adapter directory.

The repository already provides the required building blocks:

- `verl_omni.utils.fsdp_utils.collect_lora_params` gathers LoRA parameters for FSDP1, FSDP2, and non-FSDP modules.
- `DiffusersFSDPEngine.get_per_tensor_param` uses that gather path for online trainer-to-rollout synchronization.
- `NonDiffusersModelBase.load_lora_adapter` and the vLLM-Omni LoRA path consume PEFT-style artifacts.

The missing capability is durable serialization of the gathered adapter.

### Goals

- Save `adapter_config.json` and `adapter_model.safetensors` in standard PEFT format.
- Decouple adapter save frequency from full-checkpoint frequency.
- Preserve existing full-checkpoint semantics for all users.
- Keep FSDP summon, offload, and rank discipline inside the engine.
- Provide an independent adapter discovery pointer without changing training resume semantics.

### Non-Goals

[...]

Issue #92, opened 2026-05-19
[RFC] VeOmni Backend Integration Plan for H2


Kirrito-k423:
## [RFC] VeOmni Backend Integration Plan for H2

---

## Motivation

Plan to integrate VeOmni as a training backend for H2, with two main parts:

1. **VeOmni needs to support typical models** — currently none are supported (priority: Qwen-Image, Wan2.2, BAGEL)
2. **Backend integration solution** — [TODO]

## Part 1: Model Support

VeOmni should support the following typical models for RL training:

| Model | Priority | Status | Notes |
|-------|----------|--------|-------|
| Qwen-Image | P0 | ✅ Supported | VeOmni support merged via [ByteDance-Seed/VeOmni#770](https://github.com/ByteDance-Seed/VeOmni/pull/770) |
| Wan2.2 | P0 | 🔄 TBD | No update yet |
| BAGEL | P0 | ❓ Unclear | [@zhtmike noted](https://github.com/verl-project/verl-omni/issues/92#issuecomment-4486564533) BAGEL may be officially supported in VeOmni — can you share the current status or link to the relevant VeOmni PR? |

## Part 2: Backend Integration

verl-omni ：[#104 - Support VeOmni as actor engine](https://github.com/verl-project/verl-omni/pull/104)（FoolPlayer）

## Timeline

- **H2**: Complete the integration plan

zhtmike:
FYI @princepride 

Seems BAGEL will be officially supported in veomni, perhaps we can switch to veomni implementations afterwards

panshaowu:
@Kirrito-k423 Would you please add the detail technical design in the RFC, so that other developers can refer to it.

FoolPlayer:
Qwen-Image support on VeOmni is ready https://github.com/ByteDance-Seed/VeOmni/pull/770, and I have also supported VeOmni backend for verl-omni, PR will be ready this week.

SamitHuang:
TODOs following #104 
1. Support LoRA
2. Support Ulysses SP

NancyFyong:
[...]

PR #624, opened 2026-09-17
[worker, tests] feat: add LoRA to the VeOmni diffusion engine

NancyFyong:
### What does this PR do?

Add dense, adapter-only LoRA support to the VeOmni diffusion actor engine, initially validated with Qwen-Image FlowGRPO.

> [!IMPORTANT]
> This draft depends on #573 for the standard VeOmni `0.1.12` installation in CI and `docs/start/install.md`. PR #573 already updates the two shared version lines, so this PR intentionally does not duplicate them. Until #573 lands, install the required version with `uv pip install veomni==0.1.12 --no-deps`.

The default full-parameter VeOmni path remains unchanged. Unsupported configurations fail closed: `lora.merge=true`, `target_modules=all-linear`, missing adapter injection, empty adapter export, and MoE expert LoRA reference-policy bypass.

This addresses the Qwen-Image diffusion part of the VeOmni LoRA TODO in #92. It does not add LTX-2.3 support from #365 or the LoRA-only checkpoint publishing proposed in #285/#581.

### Checklist Before Starting

- [x] Searched [open VeOmni LoRA PRs](https://github.com/verl-project/verl-omni/pulls?q=is%3Apr+is%3Aopen+%22VeOmni+LoRA%22), [Qwen-Image VeOmni LoRA PRs](https://github.com/verl-project/verl-omni/pulls?q=is%3Apr+is%3Aopen+%22qwen+image+veomni+lora%22), and PR bodies referencing #92/#285. No open PR implements diffusion LoRA in `VeOmniDiffusionEngine`.
- [x] Reviewed related drafts #365, #573, and #581. Their scopes are respectively LTX-2.3 VeOmni adaptation, Qwen3-Omni/full-parameter VeOmni plus the 0.1.12 pin, and FSDP2 trainable snapshots.
- [x] The title follows the repository convention.

### Test

Environment: VeOmni 0.1.12 installed with `--no-deps`, PyTorch 2.13.0+cu129, vLLM 0.28.0, and vLLM-Omni 0.28.0rc2.dev68+gded893462.

```bash
PYTHONPATH=$PWD TORCH_COMPILE_DISABLE=1 TORCHINDUCTOR_DISABLE=1 \
python -m pytest -q \
  tests/workers/test_veomni_diffusion_lora_on_cpu.py \
  tests/workers/test_diffusers_fsdp_merged_lora_on_cpu.py \
  tests/workers/test_weight_sync_knobs_on_cpu.py
# 23 passed

mapfile -d '' files < <(git ls-files -z)
pre-commit run --files "${files[@]}"
# All 12 hooks passed
```

Two-GPU tiny Qwen-Image FlowGRPO:

```bash
NUM_GPUS=2 TOTAL_TRAIN_STEPS=2 TOTAL_EPOCHS=2 BACKEND=veomni \
bash tests/special_e2e/run_flowgrpo_qwen_image_veomni_lora.sh
```

The run completed 2/2 steps. VeOmni injected 8 dense LoRA modules; each rollout update received 16 tensors (`A` + `B` for every module). The KL reference-policy path was exercised (`actor/kl_loss` finite and nonzero), actor gradient norms were finite and nonzero, and rollout-vs-train log-prob mean absolute differences were `1.30e-5` and `1.88e-5`.

Not validated: real Qwen-Image weights, long-run convergence, LoRA-only checkpoint publishing, MoE LoRA, or `lora.merge=true`.

### API and Usage Example

No new configuration keys are introduced. Use the existing diffusion LoRA fields with explicit targets:

```bash
uv pip install veomni==0.1.12 --no-deps

python -m verl_omni.trainer.main_diffusion \
  diffusion/model_engine=veomni_diffusion \
  actor_rollout_ref.actor.strategy=veomni \
  actor_rollout_ref.actor.veomni_config.strategy=veomni \
  actor_rollout_ref.model.lora_rank=8 \
  actor_rollout_ref.model.lora_alpha=16 \
  'actor_rollout_ref.model.target_modules=[to_q,to_k,to_v,to_out.0]'
```

VeOmni has no `all-linear` shorthand, so explicit target modules are required.

### Design & Code Changes

- Translate existing verl-omni LoRA config into VeOmni 0.1.12's native `VeOmniLoraConfig` and install adapters before FSDP parallelization.
- Export the initial plain base model and subsequent PEFT-compatible LoRA tensors through the existing rollout synchronization contract.
- Implement a dense-LoRA reference-policy context by temporarily selecting an inactive adapter and restoring the active adapter in `finally`.
- Keep the full-parameter path unchanged and fail before training when the requested VeOmni LoRA capability is unavailable or unsupported.
- Add CPU contract tests and register a two-GPU tiny Qwen-Image end-to-end smoke in the diffusion GPU suite.

- [x] Read the contribution and agent instructions.
- [x] Applied the full pre-commit suite.
- [x] Documented the required VeOmni 0.1.12 dependency here; the shared installation update is already in #573.
- [x] Added CPU and end-to-end GPU coverage.
- [x] Human submitter has reviewed every changed line and run the relevant tests.

AI assistance was used for this change. This PR remains a draft pending human line-by-line review and the dependency on #573.


verl_omni/workers/engine/veomni/diffusion_impl.py
```diff
@@ -584,13 +623,55 @@ def load_checkpoint(
...
+
+        VeOmni keeps LoRA in its own ``veomni.lora`` wrappers, whose module layout
+        and state-dict keys already match PEFT on disk, so the rollout side
+        consumes this exactly like the FSDP backend's PEFT export.
+        """
+        from veomni.lora import is_veomni_lora_model
+        from veomni.lora.state_dict import get_lora_state_dict
+
+        peft_model = getattr(self.module, "_fsdp_wrapped_module", self.module)
+        if not is_veomni_lora_model(peft_model):
```
OnePunchMonk, line 637: `self.module` is a `VeOmniLoraModel`, which never sets a `.peft_config` attribute (it uses `veomni_lora_config` instead). `engine_workers.py`'s `update_weights` detects LoRA via `hasattr(peft_module, "peft_config")` (lines 1067 and 1110). For this engine that's always False, so:
- with a non-naive `checkpoint_engine.backend` (async/disaggregated training), the actor always takes the full-base-weight branch (line 1084) and never ships the trained LoRA adapter to the rollout — training silently never updates the policy the rollout samples from.
- `use_lora_fast_path` (line 1113) is also always False, so the colocated fast path is skipped (harmless, just no perf win).

Worth exposing something `engine_workers.py` can check generically (e.g. a `has_lora` property on the engine) instead of relying on a PEFT-specific attribute.

verl_omni/workers/engine/veomni/diffusion_impl.py
```diff
@@ -606,10 +687,41 @@ def param_generator():
                     tensor = tensor.to(export_dtype, non_blocking=True)
                 yield f"transformer.{name}", tensor
 
-        return param_generator(), None
+        return param_generator(), peft_config_dict
 
+    @contextmanager
     def disable_adapter(self):
```
OnePunchMonk, line 693: `VeOmniDiffusionEngine` doesn't inherit `LoRAAdapterMixin` and only implements `disable_adapter`, not `copy_adapter`/`ema_update_adapter`. Those are called from `engine_workers.py` (`copy_adapter`/`ema_update_adapter` at lines 210-219) for 'old'-adapter algorithms, which `ray_diffusion_trainer.py` requires via `rollout_adapter=old` (e.g. lines 1522-1523, 1637, 1664).

Running one of those algorithms with this VeOmni LoRA backend will hit `AttributeError: 'VeOmniDiffusionEngine' object has no attribute 'copy_adapter'" the first time the trainer tries to snapshot the 'old' adapter — a hard crash rather than the fail-closed `NotImplementedError` this PR uses elsewhere for unsupported configs (merge, all-linear, MoE reference bypass). Worth adding the same guard at construction time when `policy_state_adapters` requests anything beyond default/reference.

verl_omni/workers/engine/veomni/diffusion_impl.py
```diff
@@ -148,6 +150,40 @@ def _get_veomni_model_paths(self) -> tuple[str, str]:
...
+                "VeOmni diffusion backend does not support target_modules='all-linear'; "
+                "list the modules explicitly, e.g. "
+                "actor_rollout_ref.model.target_modules=[to_q,to_k,to_v,to_out.0]."
+            )
+
+        lora_config = {
+            "rank": self.model_config.lora_rank,
+            "alpha": self.model_config.lora_alpha,
+            "lora_modules": target_modules,
+            "exclude_modules": self.model_config.exclude_modules,
```
OnePunchMonk, line 181: This dict doesn't forward `self.model_config.lora_dtype`. The FSDP/diffusers LoRA path (`lora_adapter_mixin.py`) casts LoRA params to `lora_dtype` when set, e.g. to keep the adapter in fp32 under bf16 training. Setting `actor_rollout_ref.model.lora_dtype=fp32` with this VeOmni backend is silently accepted and silently ignored — the adapter stays in whatever dtype `veomni.lora`'s `update_layer` picks from the base layer (i.e. the mixed-precision param dtype), with no error and no numerics warning. Given the PR fails closed for other unsupported knobs (merge, all-linear, MoE), consider doing the same here, or wiring `lora_dtype` through.

verl_omni/workers/engine/veomni/diffusion_impl.py
```diff
@@ -584,13 +623,55 @@ def load_checkpoint(
...
+        peft_model = getattr(self.module, "_fsdp_wrapped_module", self.module)
+        if not is_veomni_lora_model(peft_model):
+            raise RuntimeError(
+                "LoRA is configured but the VeOmni module is not a LoRA model; the adapter "
+                "was never injected, so a sync would ship frozen base weights."
+            )
+
+        lora_config = peft_model.get_lora_config(adapter_name)
+        if base_sync_done:
+            params = get_lora_state_dict(peft_model, adapter_name=adapter_name or "default", config=lora_config)
```
OnePunchMonk, line 645: `get_lora_state_dict` only strips the adapter-name infix (e.g. `.default.`), not the `base_model.model.` prefix — VeOmni's own docstring says these keys are deliberately `base_model.model.<fqn>` "byte-identical to a stock PEFT checkpoint" (see `veomni/lora/model.py`), unlike `peft.get_peft_model_state_dict` (used by the FSDP/diffusers sibling via `collect_lora_params`), which strips that prefix before returning.

After `get_per_tensor_param` prepends `"transformer."` (line 688), the wire key here is `transformer.base_model.model.transformer_blocks.0.to_q.lora_A.weight`, vs. `transformer.transformer_blocks.0.to_q.lora_A.weight` from the FSDP path for the same field.

On the rollout side, `vllm.lora.utils.parse_fine_tuned_lora_name` only strips `base_model.model.` when it is a **literal leading prefix**; here it's stuck after `transformer.` so it's never removed, giving `module_name = "transformer.base_model.model.transformer_blocks.0.to_q"`. `DiffusionLoRAManager._get_lora_weights` (vllm_omni) then tries three candidates against that name — full name, name minus first segment, and last-segment suffix — none of which equals the real pipeline module name (`"transformer.transformer_blocks.0.to_q"`). If no layer matches, `_bind_adapter_weights` raises `ValueError(..."applies to no layer"...)`.

Since VeOmni zero-inits `lora_B` (delta is exactly zero at step 0), a short smoke run can still show near-identical rollout/train log-probs even if every adapter bind after the first update silently fails — worth double-checking with an assertion on `bound_lora_names` (or a rollout-side log) rather than just log-prob deltas, and/or stripping `base_model.model.` here the way `peft`'s own export does.

verl_omni/workers/engine/veomni/diffusion_impl.py
SamitHuang, line 1: Looks fine to me. Did you run test to verifiy the e2e precision or convergence?

NancyFyong: I am currently running the validation, Results are expected today or tomorrow.

NancyFyong: image
image
this pink is fsdp2，yellow is veomni

verl_omni/workers/engine/veomni/diffusion_impl.py
```diff
@@ -584,13 +666,72 @@ def load_checkpoint(
...
+                f"rollout adapter {requested_adapter!r}. Old/EMA-policy rollouts require "
+                "actor_rollout_ref.actor.strategy=fsdp2."
+            )
+
+        # A second base sync with no adapter sync in between leaves the rollout on base weights.
+        if not base_sync_done:
+            if getattr(self, "_lora_base_synced", False) and not getattr(self, "_lora_adapter_synced", False):
+                raise RuntimeError(
+                    "VeOmni LoRA was asked for a second base-weight sync without any "
+                    "adapter sync in between; the rollout would keep serving the frozen "
```
SamitHuang: This fail-closed only fires on a *second* base-weight export. `engine_workers.py` still uses `hasattr(peft_module, "peft_config")`, which `VeOmniLoraModel` never sets, so the non-naive `update_weights` path keeps calling `get_per_tensor_param` with default `base_sync_done=False`. Step 1 ships frozen base weights; step 2 raises here. `get_lora_peft_config` / checksum also return `None`. Expose `_is_lora` (or a `peft_config` shim) on the engine and have the worker check that instead of a PEFT attribute.

verl_omni/pipelines/qwen_image_flow_grpo/vllm_omni_rollout_adapter.py
```diff
@@ -81,6 +81,27 @@ def __init__(self, *, od_config: OmniDiffusionConfig, prefix: str = ""):
...
+        """Map diffusers output-projection LoRA names to the vLLM layout."""
+        tensors = {name.replace(".to_out.0.", ".to_out."): tensor for name, tensor in lora_tensors.items()}
+        config = dict(peft_config)
+        targets = config.get("target_modules")
+        if isinstance(targets, list):
+            config["target_modules"] = [
+                target[:-2] if target == "to_out.0" or target.endswith(".to_out.0") else target for target in targets
+            ]
+        return tensors, config
+
```
SamitHuang: `_validate_diffusion_lora_binding` is unit-tested but never called at runtime. `VLLMOmniHijack.hijack__load_adapter` maps tensors then loads them; it never passes `bound_lora_names` into this helper, so a partial or empty bind can still leave rollout on base weights. Invoke this from the hijack (or the manager bind path) after `_bind_adapter_weights`.

verl_omni/pipelines/qwen_image_flow_grpo/vllm_omni_rollout_adapter.py
```diff
@@ -81,6 +81,27 @@ def __init__(self, *, od_config: OmniDiffusionConfig, prefix: str = ""):
             local_files_only=local_files_only,
         )
 
+    @staticmethod
```
SamitHuang: This mapper is on the shared Qwen-Image FlowGRPO rollout pipeline, so it rewrites every LoRA sync — including existing FSDP Qwen runs, not just VeOmni. If FSDP already bound `to_out.0` to the vLLM Sequential layout, this rename will miss those modules. Confirm FSDP Qwen still binds after the rewrite, or gate the map on the VeOmni key layout.

tests/gpu_smoke/run_gpu_smoke_diffusion_e2e.sh
```diff
@@ -71,4 +71,8 @@ run_test 9 "Diffusion OPD v1 separate_async standalone teachers e2e" \
     env CUDA_VISIBLE_DEVICES="${CUDA_DEVICE_LIST}" NUM_GPUS="${NUM_GPUS}" SMOKE=async \
     bash tests/special_e2e/run_diffusion_teacher_smoke.sh
 
+run_test 10 "Qwen-Image VeOmni LoRA FlowGRPO e2e" \
```
SamitHuang: This is in the required ci-e2e-diffusion group, but VeOmni 0.1.12 is not pinned in this PR (that lives in still-open #573) and core VeOmni GPU tests skip when the extra is missing. Without an importorskip/install guard this will fail the whole group on the current smoke image. Keep it out of the required suite until 0.1.12 is on the image, matching docs/contributing/gpu_smoke_tests.md.

SamitHuang (commented):
Colocated naive can still export adapters via `to_peft_dict()`, but `engine_workers.py` still detects LoRA with `hasattr(..., "peft_config")`. Non-naive sync ships base weights on step 1 and hits the new RuntimeError on step 2. The Qwen bind validator is never called at runtime, `map_lora_update_to_engine` rewrites all Qwen FlowGRPO LoRA (including FSDP), and the required GPU smoke depends on VeOmni 0.1.12 from still-open #573.

examples/flowgrpo_trainer/qwen_image/README.md
```diff
@@ -1,10 +1,31 @@
...
+```
+
+The recipe mirrors `run_qwen_image_ocr_lora.sh` and replaces the actor and reference backends with VeOmni.
+It keeps vLLM-Omni for rollout and uses explicit LoRA target modules because VeOmni does not support the
+`all-linear` shorthand.
+
+The tiny LoRA smoke is standalone until the CI image installs VeOmni 0.1.12:
+
+```bash
+uv pip install veomni==0.1.12 --no-deps
```
zhtmike: Bump veomni version first, in separate PR

zhtmike (changes requested):
As comment

examples/flowgrpo_trainer/qwen_image/README.md
NancyFyong: https://github.com/verl-project/verl-omni/pull/646

verl_omni/workers/engine/veomni/diffusion_impl.py
```diff
@@ -52,6 +52,115 @@
...
+            f"veomni_config.attn_implementation={attn_implementation!r} is not supported for Qwen-Image; "
+            f"use one of {sorted(_QWEN_IMAGE_DIFFUSERS_ATTENTION_BACKENDS)}. Local flash_attention_2/3 "
+            "map to diffusers varlen kernels that mishandle Qwen-Image's text padding."
+        )
+    from diffusers.models.attention_dispatch import _AttentionBackendRegistry
+
+    # set_attention_backend also switches diffusers' process-wide default; keep it for other models.
+    active_backend = _AttentionBackendRegistry._active_backend
+    model.set_attention_backend(backend)
+    _AttentionBackendRegistry.set_active_backend(active_backend)
```
NancyFyong: As discussed offline, we now need a patch to disable VeOmni's FA for all models except `qwen-image`, and we also need to gradually fix the issue where VeOmni's `attn_impl` setting is not taking effect.

verl_omni/workers/engine/veomni/diffusion_impl.py
```diff
@@ -52,6 +52,115 @@
...
+
+
+# TODO: Remove this shim, its call sites and tests once the required VeOmni includes
+# ByteDance-Seed/VeOmni#1214 and #1236. VeOmni 0.1.12 rejects Hub attention names and Qwen-Image
+# ignores attn_implementation. Only Hub varlen backends handle Qwen-Image's text padding correctly.
+_QWEN_IMAGE_DIFFUSERS_ATTENTION_BACKENDS = {
+    "eager": "native",
+    "flash_attention_2_hub": "flash_varlen_hub",
+    "flash_attention_3_hub": "_flash_3_varlen_hub",
+}
```
NancyFyong: will remove if veomni pr merged

verl_omni/workers/engine/veomni/diffusion_impl.py
```diff
@@ -52,6 +52,115 @@
...
+
+def _reject_veomni_moe_expert_lora(model: torch.nn.Module) -> None:
+    """Reject MoE expert LoRA, which ``disable_adapter`` cannot bypass for the reference policy."""
+    from veomni.lora.moe_layers import LoraIndependentExperts, LoraSharedExperts
+
+    if any(isinstance(module, LoraIndependentExperts | LoraSharedExperts) for module in model.modules()):
+        raise NotImplementedError(
+            "VeOmni diffusion backend supports dense LoRA only, not MoE expert LoRA; "
+            "the reference policy would reuse the actor's expert adapters."
+        )
```
zhtmike: move to veomni/lora_utils.py

verl_omni/workers/engine/veomni/diffusion_impl.py
```diff
@@ -52,6 +52,115 @@
...
+
+
+# TODO: Remove this shim, its call sites and tests once the required VeOmni includes
+# ByteDance-Seed/VeOmni#1214 and #1236. VeOmni 0.1.12 rejects Hub attention names and Qwen-Image
+# ignores attn_implementation. Only Hub varlen backends handle Qwen-Image's text padding correctly.
+_QWEN_IMAGE_DIFFUSERS_ATTENTION_BACKENDS = {
```
zhtmike: can we trim the `_QWEN_IMAGE` prefix?

verl_omni/workers/engine/veomni/diffusion_impl.py
```diff
@@ -52,6 +52,115 @@
...
+            "map to diffusers varlen kernels that mishandle Qwen-Image's text padding."
+        )
+    from diffusers.models.attention_dispatch import _AttentionBackendRegistry
+
+    # set_attention_backend also switches diffusers' process-wide default; keep it for other models.
+    active_backend = _AttentionBackendRegistry._active_backend
+    model.set_attention_backend(backend)
+    _AttentionBackendRegistry.set_active_backend(active_backend)
+
+
```
zhtmike: move to veromni/patch.py for now

verl_omni/workers/engine/veomni/diffusion_impl.py
```diff
@@ -258,6 +403,11 @@ def _build_model_optimizer(self):
         self.veomni_trainer = self._build_veomni_dit_trainer()
         veomni_base = self.veomni_trainer.base
         BaseTrainer._build_model(veomni_base)
+        _apply_qwen_image_attention_backend(veomni_base.model, self.engine_config.attn_implementation)
```
zhtmike: _apply_qwen_image_attention_backend -> _apply_attention_backend ?

verl_omni/workers/engine/veomni/diffusion_impl.py
```diff
@@ -52,6 +52,115 @@
...
+        )
+
+
+def _reject_veomni_moe_expert_lora(model: torch.nn.Module) -> None:
+    """Reject MoE expert LoRA, which ``disable_adapter`` cannot bypass for the reference policy."""
+    from veomni.lora.moe_layers import LoraIndependentExperts, LoraSharedExperts
+
+    if any(isinstance(module, LoraIndependentExperts | LoraSharedExperts) for module in model.modules()):
+        raise NotImplementedError(
+            "VeOmni diffusion backend supports dense LoRA only, not MoE expert LoRA; "
```
zhtmike: interesting, I think peft support moe LoRA but veomni does not?

zhtmike (commented):
minor comments. Overall good.

@knlnguyen1802 please take a look as it may related to lora refactor later

verl_omni/workers/engine/veomni/diffusion_impl.py
NancyFyong: Certainly, but on the VeOmni side, the logic for selecting the `attention_backend` varies across different models; I am starting with `qwen-image` here, and support for subsequent models will be rolled out gradually.

NancyFyong:
done

zhtmike (approved):

zhtmike:
Let us install veomni in cpu test instead, drop the skip

knlnguyen1802 (approved):
LGTM

PR #627, opened 2026-09-18
[omni, rollout, worker, trainer, cfg, recipe, doc, tests] feat: full-vocabulary on-policy distillation for Qwen3-Omni

wsc16:
### What does this PR do?

Adds **full-vocabulary on-policy distillation (OPD)** for Qwen3-Omni on the omni V1 sync trainer, on top of the existing OPD path (#375).

The existing omni OPD is limited to top-k / sampled teacher signals: the teacher scores the student rollout with prompt logprobs, so the KL is either truncated to `topk` entries or estimated from samples. This PR removes both approximations. The teacher instead serves its **last-layer hidden states**, and the actor reconstructs the teacher's full distribution locally as `h @ W.T` using a copy of the teacher's `lm_head`, walking the vocabulary in chunks with a single online-softmax pass. The result is an exact, untruncated, unsampled `KL(p_T ‖ p_S)` at memory that stays linear in the vocabulary.

This ports and adapts [verl-project/verl#6194](https://github.com/verl-project/verl/pull/6194) (Nitrobrew, Tilde Research — see the [Nitrobrew blog post](https://blog.tilderesearch.com/blog/nitrobrew) for the design) to the omni stack: the chunked KL kernel and its `@torch.compile` decorators are dropped so NPU runs stay eager, multi-teacher routing is added, and the hidden-state transport is re-implemented on vLLM-Omni's existing `multimodal_output` channel. Mechanism, config keys and constraints are documented in `docs/algo/omni_opd.md` (new "Full-Vocabulary OPD (nitrobrew)" section).

### Checklist Before Starting

- [x] The closest is [#375](https://github.com/verl-project/verl-omni/pull/375) (top-k / sampled omni OPD, merged) and [#576](https://github.com/verl-project/verl-omni/pull/576) (MiniCPM-o simplex OPD); neither consumes teacher hidden states or computes a full-vocabulary KL. This PR builds on #375's teacher loop rather than replacing it — the `loss_mode=kl` path and its recipe are untouched.
- [x] Format the PR title as `[{modules}] {type}: {description}`
  - Title: `[omni, rollout, worker, trainer, cfg, recipe, doc, tests] feat: full-vocabulary on-policy distillation for Qwen3-Omni`

### Test

**Unit tests (CPU, no NPU/GPU required)** — 7 new files, 50 passed:

```bash
python -m pytest \
  tests/agent_loop/test_nitrobrew_opd_agent_loop_on_cpu.py \
  tests/trainer/omni/test_nitrobrew_kernel_on_cpu.py \
  tests/trainer/omni/test_omni_distillation_losses_on_cpu.py \
  tests/utils/test_lm_head_loader_on_cpu.py \
  tests/workers/config/test_omni_distillation_config_on_cpu.py \
  tests/workers/rollout/test_process_hidden_states_on_cpu.py \
  tests/workers/test_teacher_unembed_distribution_on_cpu.py -q
```

Result: **50 passed**. The chunked kernel is checked against a naive full-vocabulary KL reference, forward and backward; the lm_head loader against sharded and single-file safetensors layouts; config validation against the `use_policy_gradient` rejection.

**End-to-end** — validated by a real training run on **16x Ascend 910C**:

- **Setup**: student is Qwen3-Omni-30B-A3B-Instruct with 25% noise added to its weights; teacher is the original (un-noised) model, served by `vllm_omni` in AR mode with the hidden-states opt-in.

Launch (`<head_ip>` / `<port>` from the head node):

1. On the master node: `ray start --head`
2. On the slave node: `ray start --address='<head_ip>:<port>'`
3. On the master node:
   `bash examples/gspo_trainer/qwen3_omni/run_qwen3_omni_thinker_gspo_lora_mmk12_v1_nitrobrew_opd_npu.sh`

### API and Usage Example

No breaking changes. A new `loss_mode` and one extra loss key, plus the agent-loop override the recipe already sets:

```bash
distillation.enabled=true \
distillation.teacher_models.teacher_model.model_path=<teacher_path> \
distillation.teacher_models.teacher_model.inference.name=vllm_omni \
distillation.teacher_models.teacher_model.inference.vllm_omni.output_mode="ar" \
distillation.teacher_models.teacher_model.inference.vllm_omni.pipeline_name="qwen3_omni_moe" \
distillation.teacher_models.teacher_model.inference.enable_chunked_prefill=false \
distillation.distillation_loss.loss_mode=nitrobrew \
distillation.distillation_loss.use_policy_gradient=false \
+actor_rollout_ref.rollout.agent.agent_loop_manager_class=verl_omni.agent_loop.nitrobrew_opd_agent_loop.NitrobrewOPDAgentLoopManagerTQ
```

`loss_mode=nitrobrew` is forward `KL(p_T ‖ p_S)`; `nitrobrew_reverse_kl` is `KL(p_S ‖ p_T)`. `kd_temperature` (default `1.0`) softens both distributions. Full runnable script: `examples/gspo_trainer/qwen3_omni/run_qwen3_omni_thinker_gspo_lora_mmk12_v1_nitrobrew_opd_npu.sh`.

### Design & Code Changes

Three pieces: get the teacher's hidden states out of the engine, get the teacher's unembedding onto every actor rank, and compute the KL without ever materializing `[N, V]`.

**Rollout side — hidden-state transport**

- **`verl_omni/workers/rollout/vllm_rollout/process_hidden_states.py`** (new): monkey-patches the AR model runner so the `hidden_states` tensor that `compute_logits` consumes is rerouted to the client over the **existing** `multimodal_output` channel under a new `teacher_hidden_states` key, rather than adding wire fields. Opt-in per request via `SamplingParams.extra_args["return_hidden_states"]`, which survives the engine's per-step request bookkeeping — a prompt dict `model_intermediate_buffer` entry does not, since the runner overwrites it from the engine's own `additional_information`. Asserts `enable_chunked_prefill=False`: hidden states are only emitted for tokens scheduled in the step that finishes prefill. Deliberately does not reuse vLLM-Omni's own `hidden_states` payload key, which is a nested `{"layers": ...}` dict of mid-layer pre-norm activations for staged thinker+talker runs.
- **`vllm_omni_ar_strategy.py`**: pops `return_hidden_states` off the sampling params and re-sets it on the policy stage's `SamplingParams.extra_args` (a list of stage params otherwise drops the flag); after generation, extracts the payload into `extra_fields["teacher_hidden_states"]`. Also forwards `engine_kwargs.vllm_omni.*` into the engine args, which is how the recipe sets `block_size`.
- **`vllm_rollout/utils.py`**: installs the patch from `vLLMOmniColocateWorkerExtension.__new__`. The patch has to live in the engine worker sub-process (`NPUARModelRunner` runs in its own proc), and an AR engine has a single `--worker-extension-cls` slot, which this class already owns for LoRA and weight sync.
- **`verl_omni/agent_loop/nitrobrew_opd_agent_loop.py`** (new): `NitrobrewOPDAgentLoopManagerTQ` / `NitrobrewOPDAgentLoopWorker` (a copy of verl's `AgentLoopWorkerTQ`, since Ray forbids subclassing actor classes) fetch the hidden states and lift them from `extra_fields` onto the TQ field dict, where collate turns them into a jagged nested tensor. Left in `extra_fields` they lose tensor identity.

**Trainer side — unembedding distribution**

- **`verl_omni/utils/lm_head.py`** (new): `load_lm_head_weight` reads the teacher's `lm_head` straight out of the safetensors checkpoint (`thinker.lm_head.weight`, falling back to a tied embedding) instead of loading a whole HF model, handling sharded (`model.safetensors.index.json`) and single-file layouts and resolving remote paths through `copy_to_local`.
- **`verl_omni/trainer/omni/ray_omni_trainer.py`**: `_ensure_teacher_unembeds` loads, converts to bf16 and caches one matrix per teacher, then pushes them to the actor worker group. `_update_actor` widens verl's logits-processor trigger from `distillation_use_topk` to `use_topk or use_hidden_states`, because the chunked kernel needs the full `student_logits` tensor. `_init_resource_pool_mgr` pins `role_worker_mapping` to verl_omni's `ActorRolloutRefWorker` — upstream builds that entry from its own import, and this is what makes loss dispatch a plain branch instead of a monkeypatch.
- **`verl_omni/workers/engine_workers.py`**: `set_teacher_unembeds` stores the per-teacher bf16 CPU matrices plus a sorted key vocabulary; `update_actor` factors them into each micro-batch as `NonTensorData` (`teacher_unembeds` / `teacher_key_to_id`). The loss-fn setup selects `omni_distillation_ppo_loss` when `loss_mode in HIDDEN_STATE_LOSS_MODES`, and the module-level import of `verl_omni.trainer.distillation.losses` fires the registry decorator.

**Loss**

- **`verl_omni/trainer/distillation/nitrobrew_loss.py`** (new): `_NitrobrewKL` / `_NitrobrewReverseKL` are `torch.autograd.Function`s reconstructing teacher logits as `z @ W.T` in vocabulary chunks of 1024 with a single online-softmax pass — peak extra memory `O(N * C)` per chunk instead of `O(N * V)` — with a custom backward that recomputes rather than stores. `@torch.compile` decorators dropped so NPU runs stay eager. `_grouped_nitrobrew_kl` routes tokens to their teacher by `teacher_key_ids`; a single teacher degrades to one group. A token whose teacher has no registered unembedding raises rather than silently dropping out of the loss.
- **`verl_omni/trainer/distillation/losses.py`** (new): `omni_distillation_ppo_loss` dispatches to the chunked kernel when the full `student_logits` are available and delegates to verl's `distillation_ppo_loss` unchanged otherwise, and leaves aggregation to verl by registering `nitrobrew` / `nitrobrew_reverse_kl` (no-padding, `clamp_min(0)`).
- **`verl_omni/workers/config/omni/distillation.py`**: adds `HIDDEN_STATE_LOSS_MODES` and `OmniDistillationLossConfig`, which adds `kd_temperature` and synthesizes `loss_settings` with `use_hidden_states=True` for these modes (they are absent from verl's registry, which only knows top-k and estimator losses), and rejects `use_policy_gradient=True` for `nitrobrew` — the whole-vocab signal is meant to be backpropagated directly.
- **`verl_omni/trainer/config/omni_trainer.yaml`**: `_target_` for the omni `distillation_loss`.

**Recipe and docs**

- **`run_qwen3_omni_thinker_gspo_lora_mmk12_v1_nitrobrew_opd_npu.sh`** (new): MMK12 GSPO + LoRA full-vocabulary OPD on Ascend NPU, with data prep and runtime deps (`math-verify`, `qwen-vl-utils`).
- **`docs/algo/omni_opd.md`** and **`examples/gspo_trainer/README.md`**: the new loss modes, `kd_temperature`, the mechanism, and what the mode changes outside the loss.
- **`run_..._{mmk12_v1_npu,mmk12_v1_opd_npu,npu_avqa_v1,npu_nextqa_v1}.sh`** + README: drop the dead `VLLM_ASCEND_ENABLE_NZ` export (recent vLLM-Ascend releases no longer read it) and pass `weight_nz_mode: 0` through `engine_kwargs.vllm_omni.additional_config` instead.

### Limitations

- The teacher engine must run with `enable_chunked_prefill=false`; a chunked prefill would return partial hidden-state tensors.
- `nitrobrew` is supervised only — `use_policy_gradient=false`. The reverse-KL variant is not rejected by that check.
- Teacher and student must share the same vocabulary, since the actor multiplies student hidden states by the teacher's `lm_head`.
- Validated on Ascend NPU only; the CUDA path is untested.

- [x] Apply [pre-commit checks](https://github.com/verl-project/verl-omni/blob/main/CONTRIBUTING.md#code-linting-and-formatting).
- [ ] Add unit or end-to-end test(s) to the [CI workflow](https://github.com/verl-project/verl-omni/tree/main/.github/workflows) to cover all the code. The 7 CPU test files above run under the existing CPU job; the end-to-end path is not added to CI since it requires a multi-node vLLM-Omni teacher + student cluster.

**AI assistance was used** for this change (Claude Code). All changed lines were reviewed by the submitter and the tests listed above were run locally.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


verl_omni/workers/rollout/vllm_rollout/process_hidden_states.py
SamitHuang: `_sample_with_hidden_states` peeks `execute_model_state[4]` as last-layer hidden states. That tuple layout is a private vLLM-Omni runner contract; a pin bump that reorders the tuple will attach the wrong tensor (or None) with no `ndim`/dtype check. Unpack a named field or assert a 2-D hidden tensor before slicing.

`_merge_into_multimodal_outputs` also drops payloads when the existing `multimodal_outputs` entry is neither `None` nor `dict`. The teacher request then fails late (or trains without hidden states). Raise there instead of skipping.

verl_omni/workers/rollout/vllm_rollout/utils.py
SamitHuang: `vLLMOmniColocateWorkerExtension.__new__` swallows every exception from `apply_hidden_states_patches()`. That function imports `GPUARModelRunner` *before* `_patch_npu_runner()`. On an NPU-only worker, a GPU-side ImportError/AttributeError means the NPU patch is never installed, the warning is easy to miss, and teacher generate fails only after a full prefill.

Isolate GPU vs NPU patch install, and fail closed when `return_hidden_states` is required instead of logging and continuing.

verl_omni/workers/config/omni/distillation.py
SamitHuang: `use_policy_gradient` is rejected only for `loss_mode in ("nitrobrew",)`. `nitrobrew_reverse_kl` is in `HIDDEN_STATE_LOSS_MODES` and takes the same early `return` that skips `super().__post_init__()`.

Either reject PG for both hidden modes (or document why reverse KL is exempt) and still run the parent validations you are skipping on that path. There is also no CPU test that constructs `OmniDistillationLossConfig(loss_mode="nitrobrew_reverse_kl", use_policy_gradient=True)`.

verl_omni/trainer/distillation/nitrobrew_loss.py
SamitHuang: The PR body says teacher logits are reconstructed in chunks so "neither side materialises a full [N, V] tensor". Forward already receives `student_logits` of shape `[N, V]`, and `_NitrobrewKL.backward` allocates `grad_s = torch.empty(N, V, ...)` before filling it by chunk.

Chunking only avoids a second teacher `[N, V]` reconstruction. Please correct the claim and report peak VRAM versus top-k OPD on Qwen3-Omni's real vocab, not a CPU `[N=4, V=32]` test.

SamitHuang (changes requested):
Requesting changes on the nitrobrew / full-vocab OPD path (head `3c043e5`).

The hidden-state wire is opt-in and the CPU kernel tests exist, but four issues block merge:
1. NPU hidden capture uses `execute_model_state[4]` with no shape/type check, and `_merge_into_multimodal_outputs` silently drops non-dict payloads.
2. `apply_hidden_states_patches()` imports `GPUARModelRunner` first; the colocated worker extension fail-opens, so an NPU-only import error skips `_patch_npu_runner()` entirely.
3. `OmniDistillationLossConfig` rejects `use_policy_gradient` only for `nitrobrew`, not `nitrobrew_reverse_kl`, and that branch skips `super().__post_init__()`.
4. The PR claims neither side materialises `[N, V]`; student logits and `grad_s` still do. Correct the write-up and add peak-VRAM vs top-k OPD on Qwen3-Omni vocab. CUDA is untested as noted.

Also worth a follow-up (not blocking this round): `prepare_engine_args` only copies `engine_kwargs.vllm_omni` keys already present on `OmniEngineArgs` — confirm `block_size` / `additional_config` (`weight_nz_mode: 0`) actually reach the teacher engine. `NitrobrewOPDAgentLoopWorker` is a full copy of `AgentLoopWorkerTQ` and will drift.

Reviewed with AI assistance.

PR #591, opened 2026-09-12
[recipe, tests, doc] feat: add Qwen-Image DiffusionNFT V1 sync recipe

0z5a:
### What does this PR do?

Adds the Qwen-Image + DiffusionNFT OCR recipe on V1 `sync`, as claimed in #389. The new entrypoint selects `main_diffusion_v1`, `trainer.use_v1=true`, and `trainer.v1.trainer_mode=sync`, and uses a separate experiment name. It preserves the V0 model, adapters, NFT loss, optimizer, rollout settings, and old-policy update schedule.

The change contains the recipe, documentation, and four CPU cases covering interleaved TransferQueue sample associations and old-policy copy/EMA ordering. Trainer implementations are unchanged. MiniMax H3 and asynchronous modes are outside this PR.

This draft now includes a resumed full-model V1 checkpoint chain reaching global step 100, reward/timing figures, and run logs. The September 16 V0 short-run chain now includes a successful step-3 restore → step-4 update, saved checkpoint, and final validation over all 32 prompts, with exit code 0. The prior OOM continuation and its checkpoint audit remain separately documented. A controlled matched V0/V1 performance comparison and full pre-commit validation remain outstanding.

### Checklist Before Starting

- [x] Checked #389 and its ownership confirmation, [open PRs referencing #389](https://github.com/verl-project/verl-omni/pulls?q=is%3Apr+is%3Aopen+389+in%3Abody), and [open NFT PRs](https://github.com/verl-project/verl-omni/pulls?q=is%3Apr+is%3Aopen+NFT) on September 12. No duplicate Qwen-Image / DiffusionNFT / V1 sync recipe was found. The BOOGU adapter and other algorithm migrations cover different paths.
- [x] Used the repository PR title format.

### Test

**Completed V1 training to global step 100 (September 15 update).**

The two-L20 run `ocr-v1-100-12` resumed from global step 23 and completed all 77 subsequent training steps (1,232 training images), final validation, and the `global_step_100` checkpoint. The process exited with code 0 at 13:08 UTC on September 15. This is a resumed run; it is not an uninterrupted run from step 0, and it is not concatenated with the separate historical 48-step series below.

Qwen-Image V1 training and validation reward

| Validation global step | Mean OCR reward (32 prompts) |
| --- | --- |
| 23 (after restore) | 0.4986821790 |
| 40 | 0.5800898186 |
| 60 | 0.8906557611 |
| 80 | 0.8364970256 |
| 100 | **0.9247436151** |

[...]

PR #263, opened 2026-07-14
[1/N][ci, tests, doc] test: add L1 CPU coverage reporting

wtomin:
## Summary

The first PR of #247.

- Adds coverage reporting to the CPU unit test workflow, including terminal output, `coverage.xml`, `pytest-coverage.txt`, and a GitHub job summary.
- Adds contributor docs for the L1-L4 CI/test taxonomy, L1 CPU test naming, placement rules, and coverage expectations.
- Adds CPU tests for offline DPO dataset utilities and JPEG compressibility rewards, and skips vLLM-Omni-dependent CPU tests when the optional package is unavailable.

## Test Plan

Suggested command:

```bash
pytest -s -x -v --asyncio-mode=auto --cov=verl_omni --cov-report=term-missing --cov-report=xml:coverage.xml tests/
```

The results:
```
======================================================= tests coverage =======================================================
______________________________________ coverage: platform linux, python 3.12.12-final-0 ______________________________________

Name                                                                                Stmts   Miss Branch BrPart  Cover   Missing
-------------------------------------------------------------------------------------------------------------------------------
verl_omni/__init__.py                                                                  15      3      2      1    76%   26-28
verl_omni/agent_loop/__init__.py                                                        3      3      0      0     0%   15-18
verl_omni/agent_loop/diffusion_agent_loop.py                                          182    182     52      0     0%   14-367
verl_omni/agent_loop/single_turn_agent_loop.py                                         28     28      4      0     0%   14-82
verl_omni/agent_loop/utils.py                                                          14     14      6      0     0%   15-38
... ...
verl_omni/workers/engine_workers.py                                                   472    472    132      0     0%   14-996
verl_omni/workers/rollout/__init__.py                                                   1      0      0      0   100%
verl_omni/workers/rollout/base.py                                                       2      0      0      0   100%
verl_omni/workers/rollout/replica.py                                                   19      2      0      0    89%   36-38
verl_omni/workers/rollout/vllm_rollout/npu_utils.py                                    85     85     22      0     0%   15-171
verl_omni/workers/rollout/vllm_rollout/utils.py                                        58     58     10      0     0%   14-166
verl_omni/workers/rollout/vllm_rollout/vllm_omni_async_server.py                      263    263    104      0     0%   14-517
verl_omni/workers/utils/losses.py                                                      56     33     18      0    39%   72-120
verl_omni/workers/utils/padding.py                                                     21      2      8      2    86%   58->62, 64-65
-------------------------------------------------------------------------------------------------------------------------------
[...]

PR #633, opened 2026-09-18
[5/N][tests,ci,doc] feat: add L4 convergence recipes (#247)

### LINKED ISSUE #247: [RFC] Layered L1-L4 CI/CD Plan for VeRL-Omni

0z5a:
**Related to** `verl-project/verl-omni#247` and the claim comment
https://github.com/verl-project/verl-omni/issues/247#issuecomment-5698860563.
This PR implements the `5/N` row only. It does **not** close #247.

## Why this is not duplicating an existing PR

Checked before implementing, at baseline `0971c25cf159c993ae1b92666c9efb37a39eb9d9`:

```bash
gh issue view 247 --repo verl-project/verl-omni --comments
gh pr list --repo verl-project/verl-omni --state open   --search '247 in:body'
gh pr list --repo verl-project/verl-omni --state merged --search '247 in:body'
gh pr list --repo verl-project/verl-omni --state open   --search 'convergence'
gh pr list --repo verl-project/verl-omni --state open   --search 'L4'
gh search prs --repo verl-project/verl-omni 'tests/convergence'
git ls-remote --heads origin | grep -iE 'converg|l4'
```

- `247 in:body`, merged: only **#263** (`[1/N][ci, tests, doc] test: add L1 CPU
  coverage reporting`) — the L1 slice of the same RFC, already merged, unrelated
  to convergence curves.
- `247 in:body`, open: none.
- No open, merged, or pushed branch creates `tests/convergence/`,
  `.github/workflows/l4_weekly_convergence.yml`, or any L4 runner/report. The
  `convergence` keyword hits are recipe PRs using the word for other reasons
  (e.g. #591, #576) and the merged bugfix #378 "Fix convergence issue of
  QwenImage".
- `L4` was still documented as "a planned layer" in
  `docs/contributing/ci_cd.md` before this change.

## Scope

The `5/N` row of the RFC's own PR breakdown: *"Add L4 real-model convergence
runners, release reports, and documentation for weekly/release convergence
validation."*

In scope: `tests/convergence/` (recipes, registry, curve extraction, comparison,
release report, runner), `.github/workflows/l4_weekly_convergence.yml`, the L4
documentation, and registering `convergence` as an exempt CI-layer test folder.

Out of scope: `verl_omni/` production code, L1/L2/L3 workflows and baselines, the
RFC's proposed `scripts/ci/` package, dashboards and publication (that is `4/N`),
changes to `ready-for-ci`, and any new GPU matrix.

## Design notes reviewers should look at

1. **Fail-closed, because the maintainer said this hardware is not L4-capable.**
   `wtomin` replied on #247 that L4 needs real checkpoints, real datasets, and
   replicating the real hyperparameters, that the current GPU CI has too little
   VRAM, and that NPU CI is coming. So recipes declare their own hardware and
   asset requirements, and the runner reports `skipped` with the exact unmet
   requirement instead of faking a result. The weekly gate is therefore expected
   to report `incomplete` (and exit non-zero) until a production cluster is
   attached — deliberately, rather than being green with no evidence.
2. **Evidence levels.** `static < run < compared`, and `ready` requires `passed`
   **and** `compared`. "The recipe is valid" can never be reported as "the model
   converged".
3. **A run never becomes its own baseline.** Baseline mode reports
   `baseline_created`, which the report counts as `incomplete`.
4. **Comparability is separate from tolerance.** A baseline stores a contract
   (weights, dataset, algorithm, precision, scoring window, step budget, GPU
   shape). A mismatching run is `incomparable`, never "no regression". The commit
   SHA is recorded but is not part of comparability, since the commit is the
   change under test.
5. **Tolerances are direction-aware and absolute-floored:**
   `|current - reference| <= atol + rtol * |reference|`, so an improvement never
   fails and a near-zero reference is not governed by an infinite relative error.
6. **Failure paths are explicit.** Non-zero launcher exit, timeout, missing
   required metrics, NaN/Inf, and too few scored points all fail the case; only
   the process group started by the runner is killed on timeout.
7. **No new privilege.** The workflow keeps `actions: read` + `contents: read`,
   has no `pull_request_target`, does not label its own PRs, and runs only on the
   `L4-convergence-ci` label, manual dispatch, or the weekly schedule.

## Tests run and results

CPU only; no checkpoint, dataset, or GPU is used by the tests.

```text
pytest -s -x --asyncio-mode=auto tests/convergence/        104 passed
python3 tests/special_sanity/validate_structure.py         [OK]
python3 tests/special_sanity/check_license.py --directories .   exit 0
python3 tests/special_sanity/check_docstrings.py           [OK]
python3 tests/special_sanity/check_docs_time_info.py       [OK]
ruff check / ruff format --check (changed files)           clean
```

The suite covers the negative semantics that matter for a release gate: missing
baseline, NaN/Inf, insufficient points, duplicate `case_id`, `tiny-random`
weights, unknown metric direction, all-optional metrics, an incoherent scoring
window, the `atol + rtol` boundary, a near-zero reference, contract mismatch,
non-zero launcher exit, timeout with process-group cleanup, secret redaction in
the run manifest, and the rule that a `passed` case with only `run` evidence is
not `ready`.

GPU validation on real weights + real data was performed separately on
submitter-provided hardware; the results are summarised below and in the
validation report attached to the PR conversation, not in the repository.

## Changes found by real-hardware validation

A GPU round on 8 x L20 (real Qwen-Image weights, real OCR dataset, real
Qwen3-VL reward) surfaced three harness defects and two bad recipe assumptions;
all are fixed in the second commit with regression tests:

- overrides only substituted a placeholder when it was the *whole* value, so
  `{case_dir}/checkpoints` reached the launcher as a literal brace path;
- a launcher that crashed before emitting any step record was reported `invalid`
  rather than `failed`;
- `run_all` rebuilt the release report from everything under `current/`, so a
  filtered run could fold a stale result into its verdict;
- `trainer.total_epochs=null` breaks the diffusion trainer's dataloader setup and
  must never be set (documented as a recipe pitfall);
- a too-low free-GPU-memory floor turns a clean `skipped` into a CUDA OOM at
  rollout wake-up.

A completed real convergence curve was **not** obtained. 18 launches across 7
recipe variants all died on shared-host resources - a diffusion worker rank
holding 35.78 GiB of a 44.40 GiB L20 at TP=2, a host-RAM ceiling hit at
294.78/295 GB, a reward resource pool that is disjoint and therefore needs a fifth
GPU, and GPUs 4/5 being `Exclusive_Process` - never on the L4 layer itself. No
baseline was written, so `convergence.min_improvement` stays at its `0.0`
placeholder rather than being given an invented floor; the auditable one-command
derivation (`tests/convergence/derive_thresholds.py`) and its refusal conditions
are part of this PR. The full attempt table, the measured requirements, and the
exact conditions needed to finish the run are in the validation report attached to
this PR conversation.

## AI assistance disclosure

AI assistance **was** used: an AI coding agent driven by the human submitter produced these recipes, harness, tests, workflow, and docs; the submitter reviewed the change and ran the validation. Commit trailers record the contributing tool.

## Limits and follow-ups

- **Full reward is pending.** The shipped recipes score with the GenRM-OCR VLM judge,
  which needs a GPU of its own. The validation host could not provide one, so the
  harness was exercised end-to-end with a **model-free reward proxy**
  (`jpeg_compressibility` via `MultiVisualRewardManager`, no reward engine) instead.
  Consequently no `baseline.json` was produced and
  `convergence.min_improvement` remains at its `0.0` placeholder in every recipe,
  rather than an invented floor. The proxy validates the harness plumbing and the
  curve reader; it does **not** claim OCR convergence. The full-reward run is
  unblocked by an 80 GiB-class GPU, a fifth usable card, or two genuinely idle L20s.

- The three shipped recipes are production-shaped and cannot run on the current
  GPU CI runner; they will report `skipped` until a production cluster exists.
  That is the intended behaviour, and it is documented in `ci_cd.md`.
- Absolute thresholds are conservative and should be tightened from the first
  reviewed baseline per recipe.
- L4's console/JSONL step-record reader overlaps L3's. They were not unified
  because L3's parser lives inside a standalone script executed by path from its
  own directory, so sharing it would change a working nightly's execution
  contract inside this PR. Unifying the two readers is a deliberate follow-up.
- `actionlint` is not installed locally; the workflow was validated by YAML
  parsing and by a structural comparison with `l3_nightly.yml`, not by a real
  GitHub Actions run.

## Paired validation metric (added in `9de913e0`)

A recipe's `protocol.json` can declare a primary metric that is not a per-step
curve — the OCR recipe declares:

> paired per-prompt score difference step48 - step0; report all values and
> prompt-bootstrap 95% CI; step24 diagnostic only

Nothing scored it. `tests/convergence/paired_validation.py` does, and
`tests/convergence/test_l4_paired_validation_on_cpu.py` covers it (33 tests).

It is **evidence, not a gate**: it does not change what `verify` scores, when a
recipe counts as converged, or whether `baseline.json` is written. This recipe's
protocol says *retain original short-run gate failure*, and changing gate
semantics is not something to land under a review comment.

A mean whose interval spans zero is reported `inconclusive`, not as a small win.
That distinction is the point: the tracked training curve is averaged over a
handful of sampled images per step, so it cannot tell the two apart.

### Test result

| item | value |
| --- | --- |
| new CPU tests | **33 passed** |
| `tests/convergence/` (regression) | **137 passed** (was 104) |
| verified against real runs | yes, both runs below, not only synthetic fixtures |

Scored on the exploratory OCR recipe at 48 steps, each run from its own
validation dumps:

| run | step0 mean | step48 mean | paired mean diff | 95% CI | improved / unchanged / worse | verdict |
| --- | --- | --- | --- | --- | --- | --- |
| baseline | 0.352414 | 0.490308 | **+0.137893** | **[-0.014768, +0.296725]** | 15 / 7 / 10 | `inconclusive` |
| verify | 0.349573 | 0.041426 | **-0.308148** | **[-0.436470, -0.185985]** | 2 / 13 / 17 | `signal_negative` |

step24 diagnostic only: baseline **+0.097794**, verify **-0.008964**.

### What that says about the gate

Neither run demonstrates convergence, and **the baseline's positive mean is not
statistically supported** — its interval crosses zero. The verify run is
**significantly worse**, not merely different. So `blocked` is the correct
verdict for the layer; what is wrong is the recorded reason, which compares
against a baseline that was never a valid reference and points at the verify run
rather than at the recipe.


0z5a:
## L4 gate reported `blocked`, and I believe that is run-to-run variance, not a regression

Recording this on the PR because the plan says the weekly gate is expected to
report `incomplete` until a production cluster is attached. What actually
happened on real hardware is a **`failed`** verdict from a single-baseline
comparison, and the evidence points somewhere different.

### The comparison was apples-to-apples

Both runs used the identical recipe and contract, so this is not an
`incomparable` result:

| | value |
| --- | --- |
| `recipe_sha256` (baseline run / stored baseline / verify run) | `58972e33…f51a` — all three identical |
| contract | identical (weights, dataset, algorithm, precision, scoring window, step budget, GPU shape) |
| verdict | `failed`, `evidence_level=compared` |

### Two runs of that recipe, 48 steps each

| metric | baseline | verify |
| --- | --- | --- |
| `critic/rewards/mean`, first window | 0.19263 | — |
| `critic/rewards/mean`, final window | **0.41287** | **0.02083** |
| in-run improvement | **+0.22025** | **−0.20354** |
| `actor/loss` improvement | −1.564 (worse) | +1.442 (better) |

Same seed (`data.seed=42`, `actor_rollout_ref.rollout.seed=42`), same 48-step
budget, same GPU shape. The final-window reward differs by ~20x between two runs
of the same recipe, and note the loss moves the *opposite* way to the reward in
both runs.

### Why I read this as variance

The fixed validation at **step 0** — base weights, the same 32 prompts, `n=1`,
`val_kwargs.seed=20260918`, `validation_shuffle=False` — is *nearly* identical
between the two runs:

| | value |
| --- | --- |
| prompts with an identical score | **31 / 32** |
| prompts differing | 1 / 32 |
| mean score, baseline vs verify | 0.352414 vs 0.349573 |
| max per-prompt difference | 0.0909 |

So the evaluation path is close to deterministic but **not bit-exact** — the
reward VLM runs at `GENRM_OCR_TEMPERATURE=0`, so what remains is batching-order
nondeterminism in its reductions. Training then runs at `data.train_batch_size=2`
with `actor_rollout_ref.rollout.n=2`: **4 images per step**, for 48 steps. A
perturbation that shows up as one prompt out of 32 at step 0 is enough to put
two trajectories on different paths, and by step 48 the reward means are 0.41 and
0.02.

### What this implies for the layer

A gate that compares **one** run against **one** stored baseline cannot separate
"the recipe stopped learning" from "this run drew a different trajectory". At 4
images per step the run-to-run variance is larger than any regression the gate is
meant to catch — so a `failed` verdict here carries no signal about the code.

Options, none of which I want to pick unilaterally since they change what the
gate means:

1. **Compare distributions, not points.** N runs per side, compare medians, and
   fail only on a separation the variance supports.
2. **Make the per-step estimate stable first.** Raise `train_batch_size` /
   `rollout.n` until a single run is reproducible enough to compare, then
   re-derive `min_improvement` from that baseline with `derive_thresholds`.
3. **Keep the single-run gate but refuse to score.** Detect that the step-0
   evaluation is not bit-exact between runs and report `inconclusive` instead of
   `failed` — the run artifacts are all present to do this check.

I have the run artifacts for both sides and can implement whichever is preferred.

### Separately: the CI still cannot be dispatched from here

`.github/workflows/l4_weekly_convergence.yml` runs on
`pull_request: types: [labeled]` gated on
`github.event.label.name == 'L4-convergence-ci'`. That label does not exist in
this repository, and the author has `pull` only, so the workflow cannot be
triggered from the PR. It needs a maintainer to create the label and apply it.

0z5a:
## Follow-up: computing the protocol's own primary metric changes the picture

My comment above scored the **training** reward curve, which is estimated from
`train_batch_size=2 x rollout.n=2` = 4 images per step. `protocol.json` declares
a different primary:

> paired per-prompt score difference step48 minus step0; report all values and
> prompt-bootstrap 95% CI; step24 diagnostic only

Both runs' `validation/{0,24,48}.jsonl` are intact, so I computed exactly that:
32 fixed prompts, `n=1`, greedy reward, paired by prompt, bootstrap over prompts
(20,000 resamples).

| run | step0 mean | step48 mean | paired mean diff | prompt-bootstrap 95% CI | improved / unchanged / worse |
| --- | --- | --- | --- | --- | --- |
| baseline (15:19-16:16) | 0.352414 | 0.490308 | **+0.137893** | **[−0.014768, +0.296725]** | 15 / 7 / 10 |
| verify (16:32-17:35) | 0.349573 | 0.041426 | **−0.308148** | **[−0.436470, −0.185985]** | 2 / 13 / 17 |

step24 diagnostic only: baseline **+0.097794**, verify **−0.008964**.

### What this says

**Neither run demonstrates convergence, and the baseline's positive mean is not
statistically supported.** Its CI crosses zero — with 32 prompts the instrument
can resolve a 0.31 drop but not a 0.14 rise, so `+0.1379` is inside the noise.
That matters beyond this case: `baseline.json` was written from a run whose
improvement the protocol's own primary metric cannot distinguish from zero, and
`min_improvement` would have been derived from it.

**The verify run is significantly worse, not merely different.** Its CI lies
entirely below zero and 17 of 32 prompts regressed. That is a real degradation
of the base model over 48 steps, not run-to-run scatter.

So the corrected reading of the two comments together:

1. The gate's current metric — the 4-image-per-step training reward — is too
   noisy to compare at all (my first comment).
2. The protocol's declared metric is a much better instrument, and it says the
   recipe **does not converge at this budget**, with one run actively degrading
   the model.
3. Therefore `blocked` is the *correct* verdict for the layer. What is wrong is
   the recorded reason: the gate reports a `failed` comparison against a
   baseline that was never a valid reference, and the message points at the
   verify run rather than at the recipe.

### Concrete change I would make

Score the **validation-based paired metric with a CI requirement** instead of the
training curve, and require the baseline itself to clear it:

- baseline mode refuses to write `baseline.json` when the step0→step48 CI
  crosses zero (report `inconclusive`, keep the artifacts);
- verify mode scores the same CI, so a verdict only ever means "the paired
  validation difference moved with confidence in the expected direction";
- the training curve stays as evidence, not as the scored metric.

Under that rule this recipe is `inconclusive` at 48 steps rather than `failed`,
which is the honest state, and the L4 layer stops being able to record a
baseline that does not demonstrate anything.

I can implement this against `tests/convergence/` with CPU tests — say the word
and I will push it onto this branch rather than open a competing PR.

### Reproducing the numbers

The computation is 30 lines: load `validation/0.jsonl` and `validation/48.jsonl`,
pair `score` by row index (the protocol fixes the prompt set and disables
shuffling), then bootstrap the mean of the per-prompt differences. Happy to add
it as `tests/convergence/` tooling if the gate adopts the metric.

0z5a:
## Pushed the scorer as in-tree tooling

Commit `9de913e0` adds the metric from the comment above, so the numbers are
reproducible from the repo instead of only from run artifacts:

- `tests/convergence/paired_validation.py` (271 lines) — pairs two validation
  dumps by row index, resamples prompts for the interval, reports
  `signal_positive` / `signal_negative` / `inconclusive`, with a CLI
  (`python -m tests.convergence.paired_validation --before 0.jsonl --after 48.jsonl`).
- `tests/convergence/test_l4_paired_validation_on_cpu.py` (250 lines, 33 tests).

**What it deliberately does not do:** it does not change what `verify` scores,
when a recipe counts as converged, or whether `baseline.json` gets written. The
protocol for this recipe says *retain original short-run gate failure*, and
changing gate semantics is your call, not something I should land under a
comment. This only makes the declared primary metric scoreable.

Running it on both runs reproduces the numbers exactly:

```
$ python -m tests.convergence.paired_validation --before <run>/validation/0.jsonl --after <run>/validation/48.jsonl

baseline: paired mean diff +0.137893  95% CI [-0.014768, +0.296725]  -> inconclusive   (exit 1)
verify  : paired mean diff -0.308148  95% CI [-0.436470, -0.185985]  -> signal_negative (exit 0)
```

The exit code distinguishes "inconclusive" from "a signal" without asserting a
direction, so a caller cannot mistake a wide interval for a pass.

**Tests.** 33 new CPU tests; the case that matters most is that a positive mean
with an interval spanning zero comes back `inconclusive` — if that ever
regressed to `signal_positive`, the layer would start reporting noise as a
result. `tests/convergence/` is **137 passed** (was 104), so nothing existing
moved. The tool is verified against both real runs above, not only synthetic
fixtures.

If you would rather this live behind the gate than beside it, the change I
sketched in the previous comment (baseline refuses to record when the interval
spans zero; verify scores the same interval) is the natural next step — say the
word and I will do it on this branch.

Issue #222, opened 2026-07-02
[RFC] Design and Implementation Plan for Lance Support


leohuang257:
### 1. Motivation

`Lance` is a 3B native unified multimodal model that unifies image and video understanding, generation, and editing within a single autoregressive (AR) and diffusion framework. It supports multiple generation modes including `t2i`, `t2v`, `i2v`, `image_edit`, `video_edit`, `x2t_image`, and `x2t_video`.

Integrating `Lance` into `verl-omni` extends coverage to mainstream unified image+video multimodal models.

### 2. Dev Plan & Validation

Lance is BAGEL-lineage — `vllm-omni`'s `LancePipeline` reuses BAGEL's MoT transformer and provides native single-stage inference implementations for the listed Lance modes on the pinned `vllm-omni==0.22.0`. Therefore, no upstream serving changes are planned for the pinned version; the RL integration lives in `verl-omni` as a new `verl_omni/pipelines/lance_flow_grpo/` package following the non-diffusers path ([`integrating_a_non_diffusers_model.md`](https://github.com/verl-project/verl-omni/blob/main/docs/contributing/integrating_a_non_diffusers_model.md)), registered as `OmniLanceForConditionalGeneration` and trained with LoRA on Lance's generation-expert pathway (the `*_moe_gen` linear projections). Since Lance spans several generation modes, we land it incrementally and use this RFC to track progress.

**Implementation Plan:**

- **M1 — `t2i` RL (LoRA, `Lance_3B`):** Implement `lance_flow_grpo/` (model module, training adapter, rollout adapter, shared utilities), with the PickScore reward following the existing BAGEL PickScore recipe. Lance dispatches `t2i` to the inherited `BagelPipeline.forward` image-generation path, which already supports recording denoising latents, timesteps, and scheduler log-probs — the cheapest way to exercise trainer-side actor construction, gen-expert LoRA injection, trainer-to-rollout LoRA weight sync, and rollout↔trainer log-prob agreement end to end. Before training, audit the released checkpoint layout and align the Lance sigma/timestep schedule with the pinned `vllm-omni==0.22.0` rollout implementation.
[...]

PR #636, opened 2026-09-19
[diffusion, rollout, trainer, tests, doc] feat: add Lance text-to-image FlowGRPO training

ruiling-smartbear:
Implements M1 of #222: text-to-image FlowGRPO training for Lance.

Lance is BAGEL-lineage. vllm-omni's `LancePipeline` inherits `BagelPipeline` and overrides only model construction, so the training side reuses `BagelForTraining` and the BAGEL adapters and overrides only what Lance changes:

- **Config** comes from `Lance_3B/llm_config.json` plus Lance's latent constants, since Lance ships no BAGEL-style `config.json` carrying `vae_config` / `latent_patch_size`. The image boundary token IDs come from `vision_start_token_id` / `vision_end_token_id` in that file, falling back to the tokenizer the way the rollout derives them.
- **Latent geometry** is Wan2.2: 48 channels, 16x spatial downsample, and no 2x2 latent patch, so `patch_latent_dim` is 48 and the position grid is 64x64.
- **Weights** come from `Lance_3B/model.safetensors` (sharded or not) under BAGEL's key layout; the LM head and the understanding tower are dropped. Shapes the release decides -- the extended vocabulary and the position table -- are taken from the tensors rather than the config, and a non-square position table (the 3-D video one) is rejected explicitly instead of being silently misread.
- **Sigma schedule** uses Lance's 3.5 shift, and carries one schedule point more than BAGEL's. `LanceBagel` sets `_denoise_schedule_extra_step`, so `Bagel.generate_image` samples `num_steps + 1` points and runs `num_steps` Euler steps, where BAGEL samples `num_steps` and runs one fewer. Taking only the shift leaves the trainer one sigma short of the recorded trajectory with every remaining sigma on a different grid.
- **Rotary positions** follow Lance too. BAGEL gives the whole image block a single position and lets `latent_pos_embed` carry the layout; Lance keeps Qwen2.5-VL's `rope_scaling` on the language model, and `LanceBagel.prepare_vae_latent` gives latent `(hi, wi)` the position `(P+1, P+1+hi, P+1+wi)`, with the markers at `P` and `P+max_hw+1`. Replaying under BAGEL's scalar would take the log-probs on a different rotary basis than the one that produced the trajectory, so `LanceForTraining` builds the same positions and `LanceRotaryEmbedding` assembles the same basis. The `mrope_section` is read from the checkpoint's `rope_scaling` rather than restated.

Three small changes outside the new package, each a hook whose default is BAGEL's existing behaviour: `BagelPipelineWithLogProb` reads its sigma shift from a class attribute and builds its schedule through an overridable classmethod, and `BagelForTraining` builds its RoPE positions in an overridable method.

This branch also fixes one pre-existing defect in the BAGEL rollout adapter. The scheduler adapter returns a log-prob only for in-window steps and vllm-omni drops the `None` entries, so the recorded list is already window-aligned; the adapter then sliced it again by the window's absolute offsets, which hands the trainer too few values for any window starting after step 0 and trips the rollout-correction shape check. It is invisible today because no shipped recipe requests rollout log-probs, and it is behaviour-preserving for a window that starts at 0. It arrives here because the Lance parity check below is what needs `calculate_log_probs`, and it is confined to `bagel_flow_grpo/vllm_omni_rollout_adapter.py` plus one test.

## Validation

**Checkpoint, against the released `bytedance-research/Lance`** (CPU, no GPU):

| Check | Result |
|---|---|
| Config parsed from `llm_config.json` | hidden 2048, 36 layers, 16 heads, 2 KV heads, vocab 151936 |
| Latent geometry | `patch_latent_dim` 48, position grid 64x64 |
| `vae2llm.weight` / `llm2vae.weight` | `(2048, 48)` / `(48, 2048)` |
| `latent_pos_embed.pos_embed` | `(4096, 2048)` |
| Tensors kept by the mapping | 1020 of 1021; only `language_model.lm_head.weight` dropped |
| `load_state_dict` | 0 missing, 0 unexpected |
| Parameters | 5.87B total, 2.77B in the generation expert |
| Forward | conditional and unconditional branches both finite and different |

**Serving.** `vllm-omni serve` with the deploy config in this PR, on one A100: the server comes up in 51s and returns a 1024x1024 image in 10.6s. The latent geometry, the 64x64 position grid, the 3.5 shift and the three-branch CFG all sit on that path, so a mistake in any of them shows up in the image.

**Against the rollout's own code, not a restatement of it.** The two places Lance departs from BAGEL are compared to the modules that will consume them: the positions against `LanceBagel._per_token_mrope_for_vae_latent` (exact, for the whole 32x32 latent block plus both markers), and the rotary basis against `BagelRotaryEmbedding` (exact, cos and sin). The sigma schedule is compared against a grid rebuilt from `generate_image`'s own formula rather than from the helper the implementation uses.

**Weight sync**, against the rollout's own `Qwen2MoTForCausalLM`. The trainer keeps q/k/v and gate/up apart and the rollout fuses them, and `load_weights` only warns about a name it cannot place, so a mismatch would cost nothing at sync time and would show up as a policy that never improves. Running the export the way `lora.merge=True` streams it -- PEFT merge, `normalize_peft_param_name`, `convert_weight_keys`, the adapter's `transformer.` rewrite -- a merged LoRA update lands in `qkv_proj.gen_exp`, `o_proj.gen_exp` and `mlp_moe_gen.gate_up_proj` with the halves in the right order, the frozen text expert arrives untouched, and every rollout parameter except `lm_head.weight` is written. The nine parameters the language model does not carry (`time_embedder`, `vae2llm`, `llm2vae`, `latent_pos_embed`) are the same nine BAGEL does not sync, and this recipe freezes all of them. The tests check destinations rather than warnings, so a projection renamed out of the remap -- which still matches the text pattern by substring -- is caught by the generation half being left unwritten. Note that the two tests covering this build a vLLM module and therefore need an inferable device; see the CPU tests paragraph.

**Prompt data.** The prep script is BAGEL's with Lance's tokenizer resolution. Over the 25432 train and 2048 test PickScore prompts the parquet schema matches the BAGEL script's exactly, `max_prompt_length` truncates 27 of them (0.11%), and every caption that is not truncated survives the round trip through the wrapping and back out of `_extract_prompt_text`, which is what the rollout applies before Lance renders its own chat template. The released tokenizer designates neither `eos_token` nor `pad_token`, and the agent loop pads its first batch, so the prep writes a repaired copy to `<output_dir>/tokenizer` and the recipe points `model.tokenizer_path` at it; `<|im_end|>` is already in the vocabulary, so no token is added.

**Steps.** Validation runs at 30 rather than the BAGEL recipe's 50: 50 is BAGEL's own pipeline default, and Lance's is 30 (`LANCE_DEFAULTS.num_timesteps`). Rollouts stay at 15.

**CPU tests** (`tests/pipelines/test_lance_flowgrpo_on_cpu.py`, 28 cases). `tests/pipelines/` goes from 427 passed on the base commit to 456 passed on this branch, with 29 tests added and none failing. On a machine with a device the Lance file is 28 passed on its own.

Two of those tests, `test_weight_sync_reaches_every_trainable_parameter` and `test_merged_lora_lands_in_the_rollout_packed_weights`, are the only ones in the file that instantiate a vLLM module, because checking where a merged LoRA lands needs the rollout's own `Qwen2MoTForCausalLM`; the rest either do not touch vLLM or use only its pure-torch helpers. So they need a device, and on a runner without one they fail at `VllmConfig()` with `Failed to infer device type` (with `VLLM_TARGET_DEVICE=cpu`, `_OpNamespace '_C' object has no attribute 'silu_and_mul'`, since this vLLM wheel is CUDA-only). That is an environment property, not a defect in the sync code, but those two tests do need a GPU runner rather than the CPU pool. `ruff check` and `ruff format --check` are clean on the touched files.

**One RL step and rollout/trainer agreement** (4 GPUs, batch 4 x 8, 15 denoise steps). A first run took one complete step: 32 images, PickScore mean 0.792 (min 0.667, max 0.989, `zero_std_ratio` 0), trainer log-prob recompute 4.1s, actor update 10.1s, weight sync back to the rollout 8.7s, `grad_norm` 1.48e-3, and `actor/ratio_mean` 1.0 with `ppo_kl` and `pg_clipfrac` at 0.

A second run added `actor_rollout_ref.rollout.calculate_log_probs=True` and widened the SDE window to cover all 15 steps from step 0, so every recorded transition is compared:

| | |
|---|---|
| `rollout_corr/logprob_abs_diff_mean` | 3.85e-03 |
| `rollout_corr/logprob_abs_diff_max` | 2.59e-01 |
| `rollout_corr/log_ppl_diff` (max / min) | 3.85e-03 (1.84e-02 / 1.35e-03) |
| `rollout_corr/ppl_ratio` | 1.00387 |
| `rollout_corr/kl` | 3.85e-03 |
| `actor/grad_norm` | 7.06e-03 |

The per-timestep split (`rollout_corr/logprob_abs_diff/ts_*`) decays monotonically over three orders of magnitude, from 4.05e-02 at sigma 0.2 to 1.99e-05 at sigma 1. There is no spike at any single timestep, which is what a wrong schedule point at the end of the grid would produce, and no uniform offset across timesteps, which is what a wrong rotary basis would produce. The shape is what bf16 accumulation looks like once the log-prob's `1/std^2` term amplifies it at low sigma, not a structural disagreement.

For a magnitude, the nearest reference in the tree is `examples/gspo_trainer/README.md`, whose healthy runs report `rollout_corr/log_ppl_diff` of about 0.001 on gsm8k and about 0.002 and 0.007 on MMK12 and AVQA-R1-6K. This run's 3.85e-03 sits inside that band. That comparison is still loose: those runs are a text and omni AR model, so the quantity is taken over token sequences rather than diffusion latent transitions. The tight control is a BAGEL run of this same recipe with the same 15-step window, which would separate noise common to this path from anything Lance specific.

`actor/ratio_mean` being 1.0 is not this comparison: `old_log_probs` is the trainer's own recompute, so it only shows the update path is deterministic. `rollout_corr/*` is the comparison, and it requires `calculate_log_probs=True`, which no shipped recipe sets.

## Notes

- `3B` counts one expert. The MoT carries a text and a generation copy of every layer, so the trainer holds roughly 5.9B parameters, about 12 GB in bf16, and the rollout holds its own copy. The example recipe uses 4 GPUs, like the BAGEL one.
- On this stack (torch 2.13, vLLM 0.28) the diffusion rollout default `FLASH_ATTN_3_HUB` cannot load: `kernels-community/flash-attn3` publishes nothing built for torch 2.13. Runs need `actor_rollout_ref.rollout.rollout_attn_backend=FLASH_ATTN`. Neither this recipe nor the BAGEL pickscore recipe sets it, so the default is worth revisiting upstream. The trainer is unaffected, since the MoT attends through `F.scaled_dot_product_attention`.
- `lance_transformer.py`'s module docstring still says the generation block "shares a single rope position, same as BAGEL, so the reused rotary is correct there", and a second comment says `rope_scaling` is forced to `None`. Both predate `prepare_vae_latent`'s per-token positions; the code is what this PR follows. Worth a separate docs fix upstream.
- Video, image and video editing are the later milestones of #222. The 2-D positions added here are the image case of the same construction, so the video case extends `build_position_ids` rather than replacing it.


ruiling-smartbear:
Opened by accident while the branch was still being adjusted. Closing for now; I will open it again once it is ready.

PR #639, opened 2026-09-20
[rollout, diffusion] feat: opt-in per-sample rollout progress

### LINKED ISSUE #638: [Feature] Report per-sample progress inside a diffusion training step
### What

Inside one training step nothing reports progress. The only counter is the
trainer's step-level `Training Progress` bar, which advances once per step, so
between two of its updates a healthy but slow step and a hung one look identical.

### Evidence

`tqdm` appears in the diffusion trainer only, as the step bar:

```
verl_omni/trainer/diffusion/ray_diffusion_trainer.py:1353  tqdm(total=self.total_training_steps, desc="Training Progress")
verl_omni/trainer/diffusion/ray_diffusion_trainer.py:1797  same, the other fit branch
```

Every other phase of a step is silent:

- `verl_omni/agent_loop/diffusion_agent_loop.py` has no `tqdm` at all, and
  `generate_sequences` dispatches one task per sample and then awaits
  `asyncio.gather` with nothing reported in between
- the vllm-omni pipeline logs nothing per request
- `ray_diffusion_trainer.py:_validate()` prints one `test_gen_batch meta info`
  line at the start and nothing until the pass ends

### Why it matters

A diffusion sample costs seconds to minutes, so a step is minutes to tens of
minutes, and the phases are strictly serial: only one of rollout, reward,
log-prob recompute and update runs at a time. Without a counter, telling them
apart needs attaching to a worker and reading its stack by hand. The step bar
cannot help, because it only moves when the whole step is over.

This was hit in practice. With `trainer.test_freq=30` and
`trainer.total_training_steps=2`, step 2 is `is_last_step`, so it runs a
validation pass. The recipe sets neither `data.val_max_samples` nor
`data.val_batch_size`, so that pass covers the whole val split (2048 samples) at
`val_kwargs.pipeline.num_inference_steps=30`, twice the training value. It took
about 26 minutes and looked exactly like a hung rollout. Nothing in the logs
distinguished the two, and the run had to be diagnosed by sampling GPU
utilisation and worker stacks from outside.

### Proposed

Report per-sample progress in `generate_sequences`, labelled for the pass it
covers (`Rollout` or `Validation`).

Implemented in https://github.com/ruiling-smartbear/verl-omni/pull/2 (a fork PR,
since the change was found while bringing up a recipe there).

Note on visibility: Ray proxies worker stdout, so a bar drawn inside an agent
loop worker reaches the driver in chunks. It is written continuously to that
worker's own file under `/tmp/ray/session_latest/logs/worker-*.err`, which can be
tailed for a live view.

ruiling-smartbear:
Closes #638.

## What

A per-sample progress bar for the rollout and validation passes. Off by default.

## Usage

```bash
actor_rollout_ref.rollout.enable_rollout_progress=True
```

```
Rollout:     46%|####4     | 354/768 [04:12<04:55,  1.40sample/s]
Validation:  12%|#2        | 246/2048 [05:31<40:24,  1.35s/sample]
```

The label names the pass. Each agent loop worker prints its own share of the
batch, so a batch of 48 x 16 across 4 workers gives four bars of 192 each.

To follow it live, tail the worker's own log rather than the driver's, because
Ray proxies worker stdout and delivers it in chunks:

```bash
tail -f /tmp/ray/session_latest/logs/worker-*.err
```

## Why

A diffusion step is mostly its rollout, and nothing inside the step reported
progress. The diffusion agent loop logged nothing per sample, the reward,
log-prob recompute and update phases log nothing either, and the trainer's
step-level bar only moves once the whole step is over. Between two of its updates
a healthy but slow step and a hung one are indistinguishable, and the only way to
tell them apart was to attach to a worker and read its stack.

That is not hypothetical. With `trainer.test_freq=30` and
`trainer.total_training_steps=2`, step 2 is `is_last_step`, so it runs a
validation pass. The recipe sets neither `data.val_max_samples` nor
`data.val_batch_size`, so that pass covers the whole val split: 2048 samples at
`val_kwargs.pipeline.num_inference_steps=30`, twice the training value. It took
about 26 minutes and looked exactly like a hung rollout.

## Default

Off. A diffusion sample costs seconds to minutes, so the output only pays off on
long runs, and the step-level bar is enough otherwise. With the flag off the
`asyncio.gather` path is unchanged.

## Verification

`ruff check` and `ruff format --check` pass on the changed files, and the config
generation was kept in sync (`diffusion_rollout.yaml` plus both generated trainer
configs). The bar's output was observed in a 4xH100 smoke run of the Lance
recipe: `exit=0`, two steps, 8 images per step.


NancyFyong:
Thx for your contribution, but I think we use 'ray summary' is the same usage, what do @zhtmike @knlnguyen1802 think?

ruiling-smartbear:
I'm working on model support in this repo and needed this while using it myself.

`ray summary` reports task and actor state counts; a rollout is one long call per agent loop worker whose samples are asyncio tasks, so it shows a single RUNNING task for the whole step, the same as a hung one.

It is off by default behind `actor_rollout_ref.rollout.enable_rollout_progress=True`, so it does not replace anything and costs nothing unless you turn it on.

NancyFyong:
> I'm working on model support in this repo and needed this while using it myself.
> 
> `ray summary` reports task and actor state counts; a rollout is one long call per agent loop worker whose samples are asyncio tasks, so it shows a single RUNNING task for the whole step, the same as a hung one.
> 
> It is off by default behind `actor_rollout_ref.rollout.enable_rollout_progress=True`, so it does not replace anything and costs nothing unless you turn it on.

fine, we can wait to listen to others suggestion.

verl_omni/agent_loop/diffusion_agent_loop.py
```diff
@@ -245,7 +246,35 @@ async def generate_sequences(self, batch: DataProto) -> DataProto:
...
+        if not self.rollout_config.enable_rollout_progress:
+            outputs = await asyncio.gather(*tasks)
+        else:
+            # One diffusion sample costs seconds to minutes, and a step is mostly
+            # this rollout, so the step-level "Training Progress" bar alone leaves
+            # a healthy run looking stalled between two of its updates.  Count the
+            # samples as they finish so a slow rollout reads as slow rather than
+            # invisible.  The label says which pass, because the same bar covers
+            # the validation pass that runs on the last step.  Each worker counts
+            # only its own share of the batch, so the totals add up across workers.
```
knlnguyen1802: Simplify the comment

knlnguyen1802:
> I'm working on model support in this repo and needed this while using it myself.
> 
> `ray summary` reports task and actor state counts; a rollout is one long call per agent loop worker whose samples are asyncio tasks, so it shows a single RUNNING task for the whole step, the same as a hung one.
> 
> It is off by default behind `actor_rollout_ref.rollout.enable_rollout_progress=True`, so it does not replace anything and costs nothing unless you turn it on.

Could you mention which model support are you working on, if you can create a issue and link the PR with the issue, it is better to see the big picture

ruiling-smartbear:
It is Lance (ByteDance) FlowGRPO support - the RFC is #222, which lists the modes and the validation plan: https://github.com/verl-project/verl-omni/issues/222

The progress flag came out of that work: a rollout is one long call per agent-loop worker with the samples running as asyncio tasks, so at the Ray level a stuck step and a running one look the same, and I needed per-sample visibility while bringing the modes up. The per-mode implementation PRs are still on my fork; they will come up as the #222 milestones land, and I will link them here.

zhtmike:
I dont think it is necessary; you can get the information from profiling tools.

ruiling-smartbear:
Thanks for taking a look, and sorry for the long explanation earlier. Understood - if you still feel it is not needed, I am happy to close this PR. Thanks for the suggestion!

ruiling-smartbear:
Thanks for the suggestion - that makes sense, and profiling is certainly the better tool for finding where the time goes. The only thing I could not get from it is the live per-sample view, since a profiler has to be enabled beforehand, whereas this reported "6 of 16 samples done" during an ordinary run. If you would still prefer not to carry it, I am happy to close the PR and add the py-spy / torch-profiler recipe to the docs instead. Thanks again!

PR #644, opened 2026-09-20
fix: choice_reward should not score an untagged response as a match

### LINKED ISSUE #643: choice_reward.compute_score scores empty/malformed responses as correct when ground_truth has no <answer> tag
## Summary

`verl_omni/utils/reward_score/choice_reward.py::compute_score` can silently reward a completely empty or malformed model response with `score: 1.0` when `ground_truth` is a plain label without `<answer>` tags.

## Where

`verl_omni/utils/reward_score/choice_reward.py:35-37`

```python
prediction = extract_answer(solution_str)
target = extract_answer(ground_truth)
accuracy = float(prediction == target)
```

`extract_answer` returns `""` when the input has no `<answer>...</answer>` tag (`verl_omni/utils/reward_score/choice_reward.py:22-25`). If the model's response never emits the tag (truncated generation, garbage output, or simply not following the format) *and* `ground_truth` is stored as a bare label such as `"B"` instead of `"<answer>B</answer>"`, both `extract_answer` calls return `""`, so `prediction == target` is `True` and the reward is a perfect `1.0`.

## Repro

```python
import importlib.util
from pathlib import Path

path = Path("verl_omni/utils/reward_score/choice_reward.py")
spec = importlib.util.spec_from_file_location("choice_reward", path)
m = importlib.util.module_from_spec(spec)
spec.loader.exec_module(m)

print(m.compute_score("I ran out of tokens before answering", "B"))
print(m.compute_score("", "B"))
print(m.compute_score("asdkjfh", "B"))
```

Output on current `main` (b708218):

```
{'score': 1.0, 'accuracy': 1.0}
{'score': 1.0, 'accuracy': 1.0}
{'score': 1.0, 'accuracy': 1.0}
```

## Impact

`choice_reward` is a general, reusable rule reward (referenced from `examples/gspo_trainer/README.md` and `examples/dapo_trainer/README.md`), and nothing in its own module enforces that `ground_truth` be tag-wrapped. Any config/dataset that stores plain-string ground truths (the natural thing to do — it's just the answer letter) hits this directly: RL training could learn to exploit it by never emitting `<answer>` at all, since that scores identically to a correct tagged answer instead of `0`.

Note: the one shipped dataset converter that uses this reward, `examples/gspo_trainer/data_process/avqa.py`, happens to validate that `solution` already matches the `<answer>X</answer>` pattern before writing the parquet, so this specific exploit isn't reachable through that example alone — but the reward function itself has no such guarantee and is documented as generally reusable.

## Suggested fix

An empty prediction (no tag found at all) should never count as a match, even if `ground_truth` also fails to parse:

```python
accuracy = float(bool(prediction) and prediction == target)
```

I have a fix + regression test ready and will open a PR.

OnePunchMonk:
## Summary

Fixes #643.

`compute_score` in `choice_reward.py` extracts the `<answer>...</answer>` payload from both the response and `ground_truth` via `extract_answer`, which returns `""` when no tag is present, then compares `prediction == target`. When the model emits no `<answer>` tag at all (empty/truncated/malformed response) and `ground_truth` is a bare label without tags, both sides extract to `""` and the comparison reports a perfect match — rewarding a response that never even attempted the required answer format.

## Fix

An empty prediction (i.e. no tag found in the response) can never count as a match, regardless of whether `ground_truth` itself parses:

```python
accuracy = float(bool(prediction) and prediction == target)
```

## Independent verification

Reproduced against both commits directly (not just the added unit test):

Pre-fix (parent commit `7af77f5`):
```
'I ran out of tokens before answering' -> {'score': 1.0, 'accuracy': 1.0}
''                                      -> {'score': 1.0, 'accuracy': 1.0}
'asdkjfh'                               -> {'score': 1.0, 'accuracy': 1.0}
```

Post-fix (this branch):
```
'I ran out of tokens before answering' -> {'score': 0.0, 'accuracy': 0.0}
''                                      -> {'score': 0.0, 'accuracy': 0.0}
'asdkjfh'                               -> {'score': 0.0, 'accuracy': 0.0}
```

## Test plan

- [x] Added `test_compute_score_untagged_response_never_matches_untagged_ground_truth` covering empty/malformed responses against an untagged ground truth.
- [x] `python -m pytest tests/utils/reward_score/test_choice_reward_on_cpu.py -q` — 8 passed (existing 5 tagged-match cases + 3 new regression cases).


verl_omni/utils/reward_score/choice_reward.py
```diff
@@ -34,7 +34,10 @@ def compute_score(
     del kwargs
     prediction = extract_answer(solution_str)
     target = extract_answer(ground_truth)
-    accuracy = float(prediction == target)
+    # An empty prediction means no ``<answer>`` tag was found at all; never treat
+    # that as a match, even when ``ground_truth`` also fails to parse (both would
+    # otherwise extract to "" and compare equal).
+    accuracy = float(bool(prediction) and prediction == target)
```
wtomin, line 40: If the prediction is an empty string, and the ground_truth is not empty, I think the accuracy is zero.

If both the prediction and ground_truth are empty, I think this data point should be dropped. Setting accuracy as zero sounds not correct to me.

wtomin:
I think it is a critical correctness guard. It's just, I think we should filter out these samples with empty targets (ground truth) before computing reward scores.

OnePunchMonk:
Agreed filtering empty-ground-truth samples out of the dataset is worth doing, but I'd keep it as a separate change: it's a data-pipeline concern (which samples get scored at all), while this fix is a scoring-function invariant (an untagged response can never be counted as a correct match, regardless of what ground_truth looks like). Even with upstream filtering in place, this guard is still needed for malformed/truncated responses at inference time. Happy to open a follow-up issue for the filtering if that's useful — let me know.

